PR #11, opened 2019-03-26
fix : fix config file not work for cz

wgjak47:
fix the config files like .cz didn't work when using `cz`


Woile:
Hey thanks! Gonna tests and merge if everything goes ok :+1:

commitizen/cli.py
```diff
@@ -78,7 +77,6 @@ def load_cfg():
     for cfg in configs:
         if not os.path.exists(config_file) and os.path.exists(cfg):
             config_file = cfg
-            break
```
Woile: This `break` is not needed. If we remove it, then the app will pick up the latest. 
The idea is to pick up the first one and continue.

Woile:
Thanks, minor comment, once addressed it can be merged :+1:

commitizen/cli.py
wgjak47: But there is another break after `settings.update(dict(config.items("commitizen")))`, and it ensures to pick up the first one and continue.

Woile:
Can u rebase the PR and fix the conflicts? 
Thanks!

wgjak47:
I will try to reopen a new PR... Looks like I am not a master of git rebase

PR #13, opened 2019-04-02
fix: fix config file not working

wgjak47:
The break is not needed and it lead to config files not work.


Woile:
Thanks.

Just for future cases, the commit message could be `fix: config file not working` without the redundancy of 2 `fix`

PR #29, opened 2019-06-04
fix: #28 allows poetry add on py36 envs

### LINKED ISSUE #28: `poetry add -D commitizen` fails on python 3.6 projects
`commitizen` is supposed to be working with py36 and py37, but fails when added as dependency with `poetry` on py36 compliant project:

```
$ poetry add -D commitizen
Using version ^1.5 for commitizen

Updating dependencies
Resolving dependencies... (0.1s)

[SolverProblemError]
The current project must support the following Python versions: ^3.6
Because no versions of commitizen match >1.5,<2.0
 and commitizen (1.5.0) requires Python >3.6, commitizen is forbidden.
So, because pyteleinfo depends on commitizen (^1.5), version solving failed.

add [-D|--dev] [--git GIT] [--path PATH] [-E|--extras EXTRAS] [--optional] [--python PYTHON] [--platform PLATFORM] [--allow-prereleases] [--dry-run] [--] <name> (<name>)...
```

This is because `commitizen`'s [`python` line in `pyproject.toml`](https://github.com/Woile/commitizen/blob/a432c64bcb6dfc7202c684f0f7cdba180cbffa9f/pyproject.toml#L47) contains `python = ">3.6"` instead of `python = "^3.6"` or `python = ">=3.6"` (with a preference for the former).

Changing this should not cause a pb since [tests are indeed passing for py36](https://travis-ci.com/esciara/commitizen/jobs/205236691).

esciara:
fixes #28 .

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = ">3.6"
+python = "^3.6"
 questionary = "^1.0"
 decli = "^0.5.0"
 colorama = "^0.4.1"


Woile:
Thanks a lot, gonna release today

Issue #37, opened 2019-08-10
highlight the current option


Singein:
Just like this: 
![图片](https://user-images.githubusercontent.com/18239246/62820413-83aaad00-bb96-11e9-83ce-ff6f55bba6e5.png)

Woile:
Will add it, thanks

Lee-W:
It seems we can add a style argument to enhance it.

https://github.com/Woile/commitizen/blob/master/commitizen/commands/commit.py#L39

e.g., 

```python
from prompt_toolkit.styles import Style

custom_style_fancy = Style([
    ('qmark', 'fg:#673ab7 bold'),       # token in front of the question
    ('question', 'bold'),               # question text
    ('answer', 'fg:#f44336 bold'),      # submitted answer text behind the question
    ('pointer', 'fg:#673ab7 bold'),     # pointer used in select and checkbox prompts
    ('highlighted', 'fg:#673ab7 bold'), # pointed-at choice in select and checkbox prompts if use_pointer=False
    ('selected', 'fg:#cc5454'),         # style for a selected item of a checkbox
    ('separator', 'fg:#cc5454'),        # separator in lists
    ('instruction', ''),                # user instructions for select, rawselect, checkbox
    ('text', ''),                       # plain text
    ('disabled', 'fg:#858585 italic')   # disabled choices for select and checkbox prompts
])

...

questionary.prompt(questions, style=custom_style_fancy)

...

```

Maybe we can add a default style in `BaseCommitizen`.

https://github.com/Woile/commitizen/blob/master/commitizen/cz/base.py#L5


However, we cannot enable both prompt and the pointed text with color at this time due to the implementation here.

https://github.com/tmbo/questionary/blob/master/questionary/prompts/common.py#L229

Does this make sense to you?
If so, I would like to help with this issue, or maybe we can discuss more on it 🙂

Singein:
Adding a style argument is a good idea!
Thxs~

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

Lee-W:
I've also sent a [PR](https://github.com/tmbo/questionary/pull/27) to [questionary](https://github.com/tmbo/questionary).

Woile:
`v1.7.0` has been released :tada: 
Thanks @Lee-W for the help

PR #42, opened 2019-11-08
make style configurable for commit command

Lee-W:
#37

### DIFF
--- a/commitizen/cz/base.py
+++ b/commitizen/cz/base.py
@@ -1,7 +1,7 @@
 from typing import Optional, List, Tuple
 from abc import ABCMeta, abstractmethod
 
-from prompt_toolkit.styles import Style
+from prompt_toolkit.styles import merge_styles, Style
 
 
 class BaseCommitizen(metaclass=ABCMeta):
@@ -35,7 +35,7 @@ class BaseCommitizen(metaclass=ABCMeta):
 
     @property
     def style(self):
-        return Style(self.config["style"])
+        return merge_styles(BaseCommitizen.DEFAULT_STYLE_CONFIG, Style(self.config["style"]))
 
     def example(self) -> str:
         """Example of the commit message."""
--- a/commitizen/cz/base.py
+++ b/commitizen/cz/base.py
@@ -1,13 +1,30 @@
-from typing import Optional
+from typing import Optional, List, Tuple
 from abc import ABCMeta, abstractmethod
+
+from prompt_toolkit.styles import Style
 
 
 class BaseCommitizen(metaclass=ABCMeta):
     bump_pattern: Optional[str] = None
     bump_map: Optional[dict] = None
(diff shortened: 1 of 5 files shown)


commitizen/cz/base.py
```diff
@@ -1,13 +1,29 @@
...
+from typing import Optional, List, Tuple
 from abc import ABCMeta, abstractmethod
 
+from prompt_toolkit.styles import merge_styles, Style
+
 
 class BaseCommitizen(metaclass=ABCMeta):
     bump_pattern: Optional[str] = None
     bump_map: Optional[dict] = None
+    DEFAULT_STYLE_CONFIG: List[Tuple[str, str]] = [
```
Woile: What do you think of making this lower case to be consistent with the other attributes?

Woile:
Looks good @Lee-W ! Thanks a lot!

commitizen/cz/base.py
```diff
@@ -17,6 +33,10 @@ def questions(self) -> list:
     def message(self, answers: dict) -> str:
         """Format your git message."""
 
+    @property
+    def style(self):
+        return merge_styles(BaseCommitizen.DEFAULT_STYLE_CONFIG, Style(self.config["style"]))
```
Woile: This line fails, I think it should be something like this instead:

```python
@property
def style(self):
    return merge_styles(
        [Style(BaseCommitizen.DEFAULT_STYLE_CONFIG), Style(self.config["style"])]
    )


```

commitizen/cz/base.py
```diff
@@ -1,13 +1,29 @@
...
+from prompt_toolkit.styles import merge_styles, Style
+
 
 class BaseCommitizen(metaclass=ABCMeta):
     bump_pattern: Optional[str] = None
     bump_map: Optional[dict] = None
+    DEFAULT_STYLE_CONFIG: List[Tuple[str, str]] = [
+        ("qmark", "fg:#673ab7 bold"),
+        ("question", "bold"),
+        ("answer", "fg:#f44336 bold"),
```
Woile: I would use `#FF9D00` or `#673ab7` for the answer. Whichever you choose, it should also be the same for: `qmark`, `pointer`, `highlighted`. What do you think?

commitizen/cz/base.py
Lee-W: I agree with you. I've updated it. Thanks!

commitizen/cz/base.py
Lee-W: Yes, I've fixed it. Thanks!

commitizen/cz/base.py
Lee-W: I have no preference on colors. I just change it to `#FF9D00`

Woile:
Great work @Lee-W ! gonna release soon :tada:

Issue #45, opened 2019-11-08
handle commiting when no file change


Lee-W:
In [cz-cli](https://github.com/commitizen/cz-cli), if we run `git cz` without any file change, we'll get the message "No files added to staging! Did you forget to run git add?".

However, in `cz c`, we can still run the whole process.
The git commit is not actually created.

![Screen Shot 2019-11-08 at 5 53 05 PM](https://user-images.githubusercontent.com/5144808/68467190-ba08ce80-0250-11ea-9d0e-f617d9e25eed.png)

Maybe we can detect file change like cz-cli? Or, at least we could let the user know that this commit is not actually created.

Woile:
I like the idea. I'm not sure whether we should do it before or after the
prompted questions. What would you suggest?

Lee-W:
I would suggest do it before because in such situation all the actions after that are meaningless.

Woile:
Agreed, I had that very same issues a few minutes ago haha. Feel free to create a PR to fix this.

PR #49, opened 2019-11-09
Abort if no change

Lee-W:
* Abort `cz commit` it there is nothing to the commit
    * The underhook git commit follows [cz-cli::src/commitizen/staging.js#L9](https://github.com/commitizen/cz-cli/blob/7c4703bb2afc8bbb8a910aa958a855cad2f6beb7/src/commitizen/staging.js#L9). Although in [cz-cli::src/cli/strategies/git-cz.js/#L51](https://github.com/commitizen/cz-cli/blob/26533fcff90544a4152009ad38f4f15f96e7fefd/src/cli/strategies/git-cz.js#L51), they provide a `--allow-empty`, it seems not necessary to me. What ddo you think?

#45

### DIFF
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -9,6 +9,7 @@ from commitizen import factory, git, out
 NO_ANSWERS = 5
 COMMIT_ERROR = 6
 NO_COMMIT_BACKUP = 7
+NOTHING_TO_COMMIT = 8
 
 
 class Commit:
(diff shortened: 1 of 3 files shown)


Woile:
Fine by me, if eventually someone needs `--alow-empty` we can consider it.

alecgerona:
@Woile I think the time has come. I think we need to respect the --allow-empty flag.

Woile:
Hahahah 😱
Please create a ticket describing the problem and proposed solution (sorry for the burocracy but otherwise we will forget 😅)

Issue #6, opened 2018-11-04
How to type multiple lines for body part?


WqyJh:
I was trying to type multiple lines for body part, but when I type the `Enter` key it moved to footer part, which means that I could only type one line for body part.
So I was wondering how to change lines when typing for the body part.

Woile:
hey there's no current support for this unfortunately. I will implement something to not escape the new line symbol. It'll be possible to do `My body \n is great`.

Eventually I'll try to migrate to the new python-prompt-toolkit 2.0.0 and we'll see if it possible to use something like `shift + Enter`.

Meanwhile you can modify the body with `git commit --amend`

WqyJh:
Well, thank you for reply. I think `git commit --amend` is a good way of doing this.

televi:
Have you considered switching from whaaat to PyInquirer?  In addition to supporting the latest versions of the prompt-toolkit, it support the editor type so that users can enter multiple lines of text via their favorite editor.  I can git that a shot if you'd like an submit a pull request if I get it working.  The developer for whaaat seems to have no time at present to maintain that project while PyInquirer is still active (https://github.com/CITGuru/PyInquirer).

Woile:
I'm moving to questionary. It had support for prompt toolkit 2 and is maintained.

https://github.com/tmbo/questionary

Wil check if it's possible to multi line

Lee-W:
As I tested, it seems commitizen doesn't support multiple line input yet.

After reading [python-prompt-toolkit::multiline-input](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#multiline-input), it does not seem to be a good solution to me.

How about implement it like cz-cli?
> Provide a LONGER description of the change (optional). Use "|" to break new line:

Woile:
Seems like a valid option. I'd give it a try, it may not be the best, but it's something to start with

Issue #54, opened 2019-11-15
Easier Customization


Lee-W:
#### Goal

Make [customization](https://woile.github.io/commitizen/customization/) through configuration possible

#### Description
There is a [related discussion](https://github.com/Woile/commitizen/issues/24#issuecomment-491382322) about customization through configuration.
In the js community, there is a [leonardoanalista/cz-customizable](https://github.com/leonardoanalista/cz-customizable) that can do a similar thing.

Issue #24, opened 2019-05-09
Restrict list of types / scopes


waghanza:
Hi @Woile,

Thanks you for this awesome project. 

It really let developers (at least me) focusing on code.

However, I have a little suggestion. I see (in https://woile.github.io/commitizen/config/) that we can specify the format to use.

It could be a great feature (at least for `private` **project**) to have an option to specify
+ `type`s allowed
+ `scope`

What do you think ?

PS : It's more to avoid errors (keeping a flexibility) than restricts

Woile:
Hey I'm glad you like it! Thanks for the kind words.

Regarding the `type` and `scope`. If I understand correct you'd like to use format like:

```
v$minor.$major.$path-$scope-$type
```

The problem with this approach is that you might have more than one commit between versions, so it doesn't make much sense to specify a tag with information of one of those commits, I think it would be quite confusing. What do you think?

waghanza:
Hi,

I think I wasn't clear enough. In fact I'm talking about commit messages.

As conventional commit specifications explain, there could be a `scope`, an optional one.

In some projects, there could be a list of pre-defined `scope`s (as long as a pre-defined list of `type`s).

What I had suggested is to have them into `.cz` file, as long as we can define **per repo** standard.

I hope to be clearer this time

Woile:
Oh sorry, I completely misunderstood it haha.

Ideally this should be done using a [custom rule](https://woile.github.io/commitizen/customization/)

you could even inherit from `ConventionalCommitsCz`

```python
from commitizen.cz.conventional_commits import ConventionalCommitsCz

class ConventionalCommitsScope(ConventionalCommitsCz):
    def questions(self) -> list:
        questions = super().questions()
        questions[2] = {
                "type": "list",
                "name": "scope",
                "message": "Choose your scope",
                "choices": [
                    {
                        "value": "mysuperscope",
                        "name": "this is a crazy scope"
                    }
                ]
        }
       return questions

discover_this = ConventionalCommitsScope
```

You can read more in the link provided on how to create the `setup.py` file.

I think that would solve your problem in the sort term.

Regarding providing a configuration I will try to come up with something. I don't think those values should be part of the global config, because you can write your own commit rules and `scope` and `type` would stop having meaning in the context of your custom rule (different to conventional commits). 

But maybe, there should be a way to customize your current rule through configuration.

```toml
[tool.commitizen.rule-conf.scope]
type = "list"
name = "scope"
message = "Choose your scope"
[...]

PR #74, opened 2019-11-19
Enhance customization

Lee-W:
* add utils
  * required_validator
  * multiple_line_breaker
* add jinja support for cz_customization

#6 , #24 , #54


Woile:
Hey Lee please rebase and remove the extra commit. 

```bash
git rebase -i origin/master
```
and add the option `drop` into the extra commit. Then continue normally the rebase.

Woile:
I prefer not introducing dependencies like `jinja`. For me it's more than enough with `from string import Template`. I'd be willing to add jinja as an extra dependency.
In order to do so, you can install it doing `poetry add jinja2 --optional`
And then in the code:

```
try:
    from jinja2 import Template
except ImportError
    from string import Template
```
I think the interface of both templates is slightly different so it should be adjusted.

Lee-W:
Yes, that makes sense to me. I've updated it.

Lee-W:
Should we close #6 , #24 , #54? Or, is there any other we should do on those issues?

PR #79, opened 2019-11-19
style: previous PR broke style guidelines

gabe-camp:
Corrected quoting in error message.


Woile:
Hey @gabe-camp thanks for the PR. 
There's a problem with your previous commits. The solution is a bit complicated so the simplest I can recommend is to introduce your changes in a new PR branched from master.
```
git fetch --all
git checkout -b style-message origin/master
```

Apply your **new** changes and submit another PR.

gabe-camp:
> Apply your **new** changes and submit another PR.

ok

PR #84, opened 2019-11-19
feat(Commands/check): enforce the project to always use conventional …

### LINKED ISSUE #59: New command: cz check
#### Goal
Prevent users from commiting if it doesn't match the current cz rules.

#### Description
Let's say a team is using as cz rule `conventional-commits`, and they want to enforce the project to always use conventional-commits by doing this in a `pre-commit` hook. 
Adding the command `cz check` would make it easy to validate this.

dany2691:
…commits

closes #59

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -160,13 +160,14 @@ Usage
     --version             get the version of the installed commitizen
 
     commands:
-    {ls,commit,c,example,info,schema,bump}
+    {ls,commit,c,example,info,schema,bump,check}
         ls                  show available commitizens
         commit (c)          create new commit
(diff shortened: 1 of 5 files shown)


Woile:
Awesome work dany! 
Can I ask you for another PR with docs? 
Add an new `docs/check.md` explaining how to create the hook manually. 
Thanks a lot!

dany2691:
Thanks :D

Sure! I can do that. I think, probably, I'll do it this Saturday. Your
project is awesome and I'm going to include it in my projects.

PR #93, opened 2019-11-27
Add support for different commitizens in `cz check`

### LINKED ISSUE #91: "cz check" for different cz config
**Goal**
Make "cz check" work for cz config other than conventional-commits.

**Description**
Currently, "cz check" checks only whether the commit complies to the conventional-commits. It would be better if it can check other config as well. Maybe we should move the check pattern to each cz class

**Related Issues**
https://github.com/Woile/commitizen/issues/59

dariemp:
Closes #91


Woile:
Thanks a lot!

dariemp:
> Thanks a lot!

My pleasure!

PR #94, opened 2019-11-30
docs(check): check documentation

dany2691:
add the documentation for the check command in the docs directory


Woile:
Awesome! Thanks a lot, could you add it to mkdocs.yaml as well?

Issue #97, opened 2019-12-09
name cannot be overwritten in config


Lee-W:
**Goal**
Fix the error that config "name" cannot be overwritten

**Description**
If "name" is not provided in command line but in config only, it will not use the "name" in config.

Woile:
`v1.10.1` released :tada: :tada:

PR #100, opened 2019-12-18
Refactor config

Lee-W:
* Refactor config
    * Create a `BaseConfig` class for the default configuration
    * Use `TomlConfig` and `IniConfig` to separate the logic of parsing different type of config file
* Use new config structure to fix the previously fixed Issue #97
* Rename the config inside a config object as settings
* Use config class as an argument for **cz** and **commands** instead of dict config


Woile:
Wow, great work Lee, I will review it this week

commitizen/config/ini_config.py
```diff
@@ -0,0 +1,69 @@
+import configparser
+import json
+import warnings
+
+from .base_config import BaseConfig
+
+
+class IniConfig(BaseConfig):
+    warnings.simplefilter("always", DeprecationWarning)
```
Woile: This is being executed always, even when a user is using `toml`, which is confusing. I'd move it to the `__init__`

commitizen/config/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+import warnings
+from pathlib import Path
+
+from commitizen import defaults
+from .base_config import BaseConfig
+from .toml_config import TomlConfig
+from .ini_config import IniConfig
+
+
+def load_global_conf() -> dict:
```
Woile: return `Optional[dict]`

commitizen/config/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    warnings.warn(message, category=DeprecationWarning)
+
+    with open(global_cfg, "r") as f:
+        data = f.read()
+
+    conf = IniConfig(data)
+    return conf
+
+
+def read_cfg() -> dict:
```
Woile: This is also not dict. But I'm not sure which one is instead, I guess it's something like `typing.Type[BaseConfing]`

commitizen/config/toml_config.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        with open(self.path, "r") as f:
+            parser = parse(f.read())
+
+        parser["tool"]["commitizen"][key] = value
+        with open(self.path, "w") as f:
+            f.write(parser.as_string())
+        return self
+
+    def _parse_setting(self, data: str) -> dict:
```
Woile: This should either return something or remove the `-> dict`

commitizen/config/base_config.py
```diff
@@ -0,0 +1,34 @@
...
+
+from commitizen.defaults import DEFAULT_SETTINGS
+
+
+class BaseConfig:
+    def __init__(self):
+        self._settings: dict = DEFAULT_SETTINGS.copy()
+        self._path: Optional[str] = None
+
+    @property
```
Woile: If possible add the typehints of both properties. Thanks

commitizen/config/ini_config.py
Lee-W: Yes, I added on another branch. I just cherry-pick it to this branch. Thanks!

commitizen/config/__init__.py
Lee-W: Done.

commitizen/config/toml_config.py
Lee-W: Done.

commitizen/config/base_config.py
Lee-W: Done.

Lee-W:
Maybe we can add mypy check in the future? But I think it might be better to do it after these PRs.

commitizen/config/__init__.py
Lee-W: Done.

Issue #57, opened 2019-11-15
New command: cz init


Woile:
#### Goal
Make it easier for users to use commitizen

#### Description
Let's suppose we have an already existing repo without commitizen.
The user types `cz init` and we try to be smart and detect some information:

1. Find last tag and ask if that is the latest tag. If not prompt which is the latest tag.
2. With that data we can fill the `version` in `[tool.commitizen]`.
3. Detect if the tag begins with `v` or not and define the `tag_format` in `[tool.commitizen]`
4. Find files with a reference to the latest version and define the `files` in `[tool.commitizen]`, we could let the user choose which files from the found should be added (with a multi-choice from questionary).
5. Ask to the user if they want the changelog to be generated from scratch if not present.

This is my initial draft, I'm not sure what else could be detected. Potentially we could ask for every configuration, but it's important to keep it short. It's not about customizing, but initializing quickly. If the user wants to customize more -> read the docs.

Lee-W:
I'm working on it.
[init_command](https://github.com/Lee-W/commitizen/tree/init_command)

It seems to me adding "name" in init step is also a reasonable option.
Currently, I'm not sure how "files" should be implement. Do text search no the whole project to find version is defined in which files does not seem to be a good way.

Woile:
Yes, the name should be added, and should use as options the available rules (even the plugins). It shouldn't be a problem

We could search in the current directory and inside a folder called `src` or like the project name (for this project would be `commitizen`), and make it not recursively.

```
.
src/
commitizen/
```

Lee-W:
It seems to me we might refactor part of the config so that this enhancement can be finished. One of the problems I encounter now is that I'm not able to distinguish the default config and config load from files.
My Initial thought is to have a `BaseConfig` abstract class, `DefaultConfig`, `TomlConfig`. The difference in each child class will be how to parse config so that we might be able to support other common data formats. I think `json` would be a good choice. It's commonly used and not as hard to handle as `ini` does.

As for the `version_files`, do you mean code like `glob.glob('*version*')` and `glob.glob(*/*version*)` ?

Woile:
Fine by me, let's give it a try, I think the config part needs some love.

Regarding the `version_files` I was thinking more like:
1. `ls src/` `ls .` `ls commitizen` (maybe using `os.walk` or `os.listdir`)
2. Get the *files* only (ignore directories).
[...]

PR #102, opened 2019-12-26
Init command

Lee-W:

#57 
* Add init command for setup "name", "tag", "tag_format" when no config file exists.

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -92,12 +92,12 @@ class Init:
 
         if not is_correct_format:
             tag_format = questionary.text(
-                "Please enter the correct version format: ", style=self.cz.style
+                'Please enter the correct version format: (default: "$version")',
+                style=self.cz.style,
             ).ask()
 
             if not tag_format:
-                out.error("Tag format is required!")
-                raise SystemExit()
+                tag_format = "$version"
         return tag_format
 
     def _update_config_file(self, values):
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -1,3 +1,5 @@
+from packaging.version import Version
+
 import questionary
 
 from commitizen import factory, out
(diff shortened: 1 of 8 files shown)


commitizen/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+            if "toml" in config_path:
+                self.config = TomlConfig(data="", path=config_path)
+            else:
+                self.config = IniConfig(data="", path=config_path)
+
+            self.config.init_empty_config_file()
+
+            values_to_add["name"] = self._ask_name()
+            tag = self._ask_tag()
+            values_to_add["version"] = tag
```
Woile: The `tag` should be `tag.strip()` to remove the new lines gotten from git. And actually the version should be parsed to retrieve the version only part of the tag. 
I've tried it and the output of version in my case was `v1.10.1` and it should be `1.10.1`.

You can fix it by doing:

```py
from packaging.version import Version
v = Version(tag.strip())
v.public
```

Woile:
Looks great Lee!

commitizen/commands/init.py
Lee-W: Thanks! I didn't notice that. I move the `strip()` into `git.py` so that the function can get the correct tag, and handle the `Version(...)` part in the `init.py` so that we can get the correct version from a tag.

commitizen/commands/init.py
```diff
@@ -0,0 +1,109 @@
...
+            is_correct_format = questionary.confirm(
+                f'Is "{tag_format}" the correct tag format?', style=self.cz.style
+            ).ask()
+
+        if not is_correct_format:
+            tag_format = questionary.text(
+                "Please enter the correct version format: ", style=self.cz.style
+            ).ask()
+
+            if not tag_format:
```
Woile: Should we make tag format required? 
For a new user there's no information in the prompt when seeing this message:
`Please enter the correct version format`, they probably don't know how to set the format, there should be an example, or it should be optional. Nothing means that the raw version will be used (which is fine). I think if a user really wants to format the tag, they'll go to the docs

Lee-W: I think adding a default is a reasonable option to me.

Issue #73, opened 2019-11-19
show the version of the current project


Lee-W:
#### Goal
show the version of the current project

Lee-W:
This PR has been merged. Should we close this issue?

PR #104, opened 2019-12-27
Show proj version

Lee-W:

#73 

* Add `--project` and `--commitizen` argument to `cz version` for showing either current project or installed commitizen


commitizen/commands/version.py
```diff
@@ -4,10 +4,19 @@
...
-    """Get the version of the installed commitizen."""
+    """Get the version of the installed commitizen or the current project."""
 
     def __init__(self, config: BaseConfig, *args):
         self.config: BaseConfig = config
+        self.parameter = args[0]
 
     def __call__(self):
-        out.write(__version__)
+        if self.parameter.get("project"):
```
Woile: What do you think of outputting just the version without "text". I think it would be useful to pipe it into other commands, like doing `cz version -p > VERSION`, etc. Otherwise the user will end up having to parse the version. I already regret `cz bump --dry-run` not showing only the version.

Instead, for more information, a `--verbose` flag could be added.

Lee-W: Yes, that's would be a good idea! We can use `-p`, `-c` for only showing the versions, and '--verbose' showing the message like the following.

```text
Project Version: ......
Installed Commitizen Version: ......
```

Issue #106, opened 2019-12-27
Github Action: Run test did not failed when unit test failed


Lee-W:
**Goal**
Make pipeline failed once a step failed

**Description**
There are three steps in `./script/test`. If the first two steps failed, but the third one passes, the pipeline will continue. It's because the third step will overwrite the status code and GitHub will not know the first two fail.

**Solution**
Split "Run test" into 3 separate steps

PR #107, opened 2019-12-27
fix(github_action/pythonpackage): separate run test into 3 steps

Lee-W:
There is an error hard to fix in the original design but is fixed in the "refactor-config" new design.

without separating, the pipeline will continue when the third step succeeded even if the first two steps failed

#106

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -24,7 +24,13 @@ jobs:
         poetry install
     - name: Run tests
       run: |
-        ./scripts/test
+        poetry run pytest --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen tests/
+    - name: Run black style check
+      run: |
+        poetry run black commitizen tests --check
+    - name: Run flake8 style check
+      run: |
+        poetry run flake8 --max-line-length=88 commitizen/ tests/
     - name: Upload coverage to Codecov
       if: runner.os == 'Linux'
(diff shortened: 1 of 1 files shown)


Lee-W:
@Woile I've tested the failed test `test_bump_command` on my local machine, but it succeeded. Do you have any clue on why this failed in Github Action?

Lee-W:
I've already figured out the problem is lacking git user when running test and fixed it.

Woile:
I would like to leave the `./scripts/test`. I want to keep the same interface for local development and ci.
If you check `.github/workflows/bumpversion.yml`, there's a `Configure repo` step, we can try adding that. But I've seen it failed when any of the steps fail. Could you share which step failed? Keep in mind that github actions are quite new and they are not working 100% fine.

Lee-W:
The configure is not the main problem. The problem is that the script will go on even if the previous steps fail. For example, if I send a pull request that cannot pass the unit tests but without any style error. The pipeline will succeed. I would recommend using `invoke` or `make` to solve this problem. I just tested if any of the steps return status code other than 0, both tool will break immediately.

Woile:
We only need to add `#!/bin/sh -e` at the beginning of the script and it will be solved.

> -e | If non-interactive, exit immediately if a command fails.

Lee-W:
That sounds great. I just send another PR https://github.com/Woile/commitizen/pull/109 for it. I'll close this one.

PR #109, opened 2019-12-31
fix(scripts/test): ensure the script fails once the first failure happens

Lee-W:
#100


Woile:
Thanks a lot Lee for the great work! :+1:

PR #110, opened 2020-01-02
fix(cli): fix name cannot be overwritten through config in newly refactored config design

Lee-W:
#97

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -186,7 +186,7 @@ def main():
         out.error("Command is required")
         raise SystemExit()
 
-    if args.name:
+    if args.name and not conf.path:
         conf.update({"name": args.name})
 
     if args.version:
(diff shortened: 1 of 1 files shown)


commitizen/cli.py
```diff
@@ -20,7 +20,6 @@
         {"name": "--debug", "action": "store_true", "help": "use debug mode"},
         {
             "name": ["-n", "--name"],
-            "default": "cz_conventional_commits",
```
Woile: by removing it from here we don't get the documentation when doing `--help` :/

Lee-W: I actually hardcode it in the help message.

Lee-W: Since `argparse` won't tell us whether it's default or not, it's hard for us to know whether we should use the name in the config because user does not enter the name or user wants to use "cz_conventinoal_commits"

Lee-W: @Woile Is there any concern on this solution? Or, do you have idea on how I should to fix this issue? 🙂

Issue #119, opened 2020-01-13
mkdoc fails in CI


Lee-W:
**Goal**
fix making documment issue
e.g., [Upload Python Package - on: push (2a2c1cc)](https://github.com/Woile/commitizen/commit/2a2c1ccb53508f70ac75dc102bb8dbf2e7c20173/checks?check_suite_id=395301498)

**Description**
It seems `GITHUB_ACTOR` and `GITHUB_TOKEN` are not given. If exposing of these data  is the concern, maybe we can try [actions-gh-page](https://github.com/peaceiris/actions-gh-pages)
?

Woile:
Nice, I like the github actions page.
I think that the problem is [here](https://github.com/Woile/commitizen/blob/master/scripts/publish#L9), since my last update from `actions/checkout@v1` to `actions/checkout@v2` this has somehow broken. I've been updating the docs manually.

Woile:
@Lee-W can we close this one?

Lee-W:
Yes 🙂

PR #121, opened 2020-01-14
use peaceiris/actions-gh-pages@v2 to publish docs

Lee-W:
#119

### DIFF
--- a/.github/workflows/pythonpublish.yaml
+++ b/.github/workflows/pythonpublish.yaml
@@ -29,3 +29,10 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         ./scripts/publish
+        poetry run mkdocs build
+    - name: Push doc to Github Page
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonpublish.yaml
```diff
@@ -29,3 +29,10 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         ./scripts/publish
+        poetry run mkdocs build
+    - name: Push doc to Github Page
+      uses: peaceiris/actions-gh-pages@v2
+      env:
+        PERSONAL_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+        PUBLISH_BRANCH: gh-pages
+        PUBLISH_DIR: ./site
```
Woile: Just checking, but are you sure mkdocs build output goes to `./site`?

Lee-W: Yep, the pages are generated in `./site` locally.

Issue #53, opened 2019-11-15
Generate Changelog from commits


Lee-W:
#### Goal
Generate a changelog using [Keep a changelog](https://keepachangelog.com/en/1.0.0/)
#### Description
* [WIP commit](https://github.com/Woile/commitizen/commit/f5943536b94fc088ac27396f487db4e733d30413)
  * Step 1: Parse changelog (Done)
  * Step 2: Parse git log

#### Other Reference
* [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages)
    * [conventional-changelog/packages/git-raw-commits/index.js](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/git-raw-commits/index.js)
    * [conventional-changelog-cli/](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli)
        * [cli.js](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/cli.js)
    * [conventional-changelog-core](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core)
        * [lib/merge-config.js](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core)
* [vaab/gitchangelog (python)](https://github.com/vaab/gitchangelog)
    * [Good ways to manage a changelog using git?](https://stackoverflow.com/questions/3523534/good-ways-to-manage-a-changelog-using-git)
* [git-extras](https://github.com/tj/git-extras)
    * [git-changelog](https://github.com/tj/git-extras/blob/master/Commands.md#git-changelog)
* [github-changelog-generator (ruby)](https://github.com/github-changelog-generator/github-changelog-generator#why-should-i-care)
* Original Git
    * [Generate your changelogs with Git log](https://coderwall.com/p/5cv5lg/generate-your-changelogs-with-git-log)
    * [Git: How to get latest tag version from bash script](https://cleanprogrammer.net/git-how-to-get-latest-tag-version-from-bash-script/)

pawamoy:
I don't what's your progress on this, but you could take a look at the one I wrote as well: [git-changelog](https://github.com/pawamoy/git-changelog)

It's not as flexible as I would like it to be, but I'm slowly improving it. Maybe you'll find interesting snippets :)

Lee-W:
It looks great! Will take a deeper look. Thanks 😃

[...]

PR #126, opened 2020-01-24
 Generate Changelog from commits

Lee-W:
#53 
Generate changelog from commits but it will overwrite the existing file. It now supports `--dry-run`, `--start-rev` and `--file-name` arguments. I also merge the [WIP commit](https://github.com/Woile/commitizen/commit/f5943536b94fc088ac27396f487db4e733d30413) you implemented here, but it's not used. They may be useful when we're merging the new changelog with the existing one.

## TODO
* [ ] Find out a way to merge changelog with existing file
* [ ] Get the latest version from config and use it as the default of `changelog` command

## Example

The following changelog is an example generated comparing to the master branch.

```markdown
# CHANGELOG

## Unreleased
### feat
- make changelog_file an option in config
- exit when there is no commit exists
- add --start-rev argument to `cz changelog`
- generate changelog based on git log
- generate changelog_tree from all past commits
- add changelog_map, changelog_pattern and implement process_commit
- add default process_commit for processing commit message
- changelog tree generation from markdown

### fix
- fix list format
- set changelog_map, changelog_pattern to none as default
- remove --skip-merge argument
- add changelog arguments


```


Woile:
You are the best @Lee-W !! Would you mind sending this PR to the branch `command-changelog`, I'd also like to contribute to this changes, and once we have it a bit more polish we release it

Lee-W:
Sure! This workflow makes more sense to me. The changelog seems to be a huge update to me. We should make sure it works well enough before release. 😀

PR #135, opened 2020-02-01
Fix pre commit hook

Lee-W:
In the previous pre-commit PR, I add the wrong `stages` argument into the `.pre-commit-hooks.yaml`. It should be defined on the user side. I'll also update the document.

BTW, I notice it didn't bump the version and release on the previous PR.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,6 +4,5 @@
   entry: cz check --commit-msg-file
   language: python
   language_version: python3
-  stages: [commit-msg]
   require_serial: true
   minimum_pre_commit_version: "0.15.4"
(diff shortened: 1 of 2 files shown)


Kurt-von-Laven:
@Lee-W, I believe it was correct before. Hooks can [confine themselves to particular stages](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages) when they know they would only be useful in those stages. Maybe one could argue that there is value in adding `prepare-commit-msg` to the list of stages in which one could run the hook, but unless someone comes forward with a good reason to do that, I would wager even that case is more likely to be a misconfiguration. Confining the hooks precludes what I suspect to be a far more likely misconfiguration, where the user accidentally runs the hook pre-commit, for example. I have opened #514 proposing to revert this revert, so please let me know if I am missing something.

PR #136, opened 2020-02-20
Remove redundant readme

Lee-W:
* refactor(docs): remove README.rst and use docs/README.md

By removing `README.rst`, we no longer need to maintain two documents with almost the same content.
Github can read docs/README.md as README for the project.


Woile:
Thanks Lee

PR #155, opened 2020-04-09
feat(find_increment): allow to use regex keys for bump_map

vBLFTePebWNi6c:
Default bump_map doesnt consider breaking change identificator "!". Regex keys should take care of it.

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -6,6 +6,6 @@ if [ -d 'venv' ] ; then
 fi
 
 ${PREFIX}pytest --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen tests/
-${PREFIX}black commitizen tests
-${PREFIX}isort --recursive commitizen tests
+${PREFIX}black commitizen tests --check
+${PREFIX}isort --recursive --check-only commitizen tests
 ${PREFIX}flake8 --max-line-length=88 commitizen/ tests/
--- a/scripts/test
+++ b/scripts/test
@@ -6,6 +6,6 @@ if [ -d 'venv' ] ; then
 fi
 
 ${PREFIX}pytest --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen tests/
-${PREFIX}black commitizen tests --check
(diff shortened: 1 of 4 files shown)


scripts/test
```diff
@@ -6,6 +6,6 @@ if [ -d 'venv' ] ; then
 fi
 
 ${PREFIX}pytest --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen tests/
-${PREFIX}black commitizen tests --check
```
Woile, line 9: These shouldn't be removed, these are for "testing" purposes, in order to apply them there's a `lint` script

vBLFTePebWNi6c: fixed

tests/test_bump_find_increment.py
```diff
@@ -47,17 +66,18 @@
...
+        (MAJOR_INCREMENTS_CC, "MAJOR", defaults.bump_pattern, defaults.bump_map),
+        (MAJOR_INCREMENTS_REGEX_CC, "MAJOR", BUMP_PATTERN_REGEX_CC, BUMP_MAP_REGEX_CC),
+        (NONE_INCREMENT_CC, None, defaults.bump_pattern, defaults.bump_map),
     ),
 )
-def test_find_increment(messages, expected_type):
+def test_find_increment(messages, expected_type, regex, increment_map):
     commits = [GitCommit(rev="test", title=message) for message in messages]
-    increment_type = bump.find_increment(commits)
+    increment_type = bump.find_increment(commits, regex, increment_map)
```
Woile, line 80: Seems to me that the regex from conventional commits has not been updated, as far as I see, only the tests are using the new regex. Before, by not passing `increment_map` we were using the default one.
Yon can also change `BUMP_PATTERN_REGEX_CC` to point to `defaults.bump_pattern`

`https://github.com/commitizen-tools/commitizen/blob/master/commitizen/defaults.py#L19`

vBLFTePebWNi6c: Seems fair. Changed default `bump_pattern` and `bump_map`

Woile:
Thanks a lot :)

PR #156, opened 2020-04-09
Fix mypy check

Lee-W:
* style
   * fix all the mypy warning
    * config: move .flake8 to setup.cfg and add mypy, pytest config
* ci
    * add mypy check
    * add stricter config but comment out those we not yet follow
    * remove redundant arugment (it's been configed in setup.cfg)
* test(cz/customize): add test case initialize_cz_customize_failed


Lee-W:
@Woile I made some change on config files (i.e. [setup.cfg](https://github.com/commitizen-tools/commitizen/blob/24fb4ad23854aa78d7f4595463e92dc2a7c39db4/setup.cfg)) of this project. Could you check whether they make sense to you? Other than that is just fixing mypy warnings.

setup.cfg
```diff
@@ -0,0 +1,38 @@
+[tool:pytest]
+addopts = --strict-markers
+norecursedirs = .* build dist CVS _darcs {arch} *.egg venv env virtualenv
+
+
+[mypy]
+files = commitizen, tests
```
Woile, line 7: Do we need to enforce typehints in tests?

setup.cfg
```diff
@@ -0,0 +1,38 @@
...
+ignore_missing_imports = true
+# disallow_untyped_calls = True
+# disallow_untyped_defs = True
```
Woile, line 10: I'd be fine with `ignore_missing_imports` and `disallow_untyped_defs`, what'd be the benefit for the others?

Woile:
Thanks for taking the time to do this!

setup.cfg
```diff
@@ -0,0 +1,38 @@
+[tool:pytest]
+addopts = --strict-markers
```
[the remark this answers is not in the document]
Lee-W: `--strict-markers` prevents developers from using markers not explicitly listed in `setup.cfg`. By doing so, we can avoid using markers with the typo and let `pytest` pass silently. It's one of the best practices I learn from some pytest book.

setup.cfg
Lee-W: It can enhance the correctness of our test cases. It does not require us to annotate type in tests, but ask us to use the correct type.

setup.cfg
Lee-W: The basic idea of all others is the same as `disallow_untyped_defs` which is to enforce a stricter type annotation.

Actually, they are what I just learned from watching the excellent talk by Vita Smid in EuroPython 2019 haha
[Static typing: beyond the basics of def foo(x: int) -str:](https://lee-w.github.io/pycon-note/posts/europython-2019/2020/03/static-typing-beyond-the-basics-of-def-foo-x-int-str/)

Lee-W:
@Woile If the configuration looks good to you, I'll merge this pull request late today 🙂

Woile:
Yes it's fine :+1:  if any of the new settings starts annoying we can discuss about changing it.

PR #158, opened 2020-04-11
Added conventional commits with JIRA scope.

Crystalix007:
This adds a new commit standard which basically enforces the scope of the conventional commit to be in the form `project-issueNumber`, in order to comply with JIRA issue tagging.


Woile:
I don't know if this needs to be part of the library, looks more like a plugin, they can easily be created with [commitizen_cz_template](https://github.com/commitizen-tools/commitizen_cz_template).
We can create here a new page with a list of alternative commit rules.

What do you think @Lee-W ?

Lee-W:
@Woile I agree with you. It seems to me we should include only the most used templates in commitizen itself to make it lightweight. Maybe we can create a awesome-commitizen-template list for it?

Woile:
I'd link directly in the commitizen docs, like "Third party commitizen" or "Packages", or `Catalog`.

Lee-W:
"Third party commitizen" seems to be a better one for me.

Lee-W:
@Crystalix007 Thanks for your contribution! As discussed with @Woile, we're not planning on making it a part of commitizen. Thus, I'm closing this PR. But we're happy to add this template as a third-party commitizen in our documentation. Please let us know if you want us to add it.

Crystalix007:
Sorry for the delay.

Is [this](https://pypi.org/project/conventional-JIRA/) okay?

The source is hosted [here](https://github.com/Crystalix007/conventional_jira).

Woile:
Yes, that's fine, you can create a PR for the docs, create a new section: `Third party commitizen`.

Issue #161, opened 2020-04-18
Use pypi api token instead of account/password


Lee-W:
@Woile We're currently using your pypi account/password for publishing commitizen to PyPI. [1] Do you think we should use API token instead to avoid exposing your credentials?

[1] https://github.com/commitizen-tools/commitizen/blob/master/scripts/publish#L2
[2] https://pypi.org/help/#apitoken
[3] https://github.com/pycontw/pycontw-postevent-report-generator/blob/master/.github/workflows/python-publish.yaml#L35

Woile:
I'll take care soon, thanks Lee

Lee-W:
I just update the documentation. 🙂 I'm considering [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) as a reference as well. It's not necessary for us since we're using poetry. But it could be a great option for those using setup.py However, I'm thinking of whether it will make the document too complicate to read. What do you think?

Woile:
We can add a recommendation for the github action. 
But poetry and twine require one line to publish the package. And because username is the string "__token__" and the password (token) is set as a secret, there's not much complexity, is used as a normal upload.

What I actually noticed is that we should specify that the example is specifically for python.

Lee-W:
The title of that section is "Publishing a python package". It seems to be clear for me.

Woile:
Perfect, brain fart haha

Woile:
we are now using `access_token` :tada:

PR #162, opened 2020-04-20
docs(tutorial/github_actions): suggest using pypi api token

Lee-W:
#161

### DIFF
--- a/docs/tutorials/github_actions.md
+++ b/docs/tutorials/github_actions.md
@@ -56,10 +56,13 @@ Push to master and that's it.
 
 Once the new tag is created, triggering an automatic publish command would be desired.
 
-In order to do so, the first two secrets need to be added with the information
-of our pypi account.
+In order to do so, the crendetial needs to be added with the information of our PyPI account.
 
-Go to `Settings > Secrets > Add new secret` and add the secrets: `PYPI_USERNAME` and `PYPI_PASSWORD`.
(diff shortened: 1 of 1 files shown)


docs/tutorials/github_actions.md
```diff
@@ -56,10 +56,13 @@ Push to master and that's it.
...
 Once the new tag is created, triggering an automatic publish command would be desired.
 
-In order to do so, the first two secrets need to be added with the information
-of our pypi account.
+In order to do so, the crendetial needs to be added with the information of our PyPI account.
 
-Go to `Settings > Secrets > Add new secret` and add the secrets: `PYPI_USERNAME` and `PYPI_PASSWORD`.
+Instead of using username and password, we suggest using [api token](https://pypi.org/help/#apitoken) generated from PyPI.
+
+After generate api token, use the token as the PyPI password and `__token__` as the username.
```
Woile, line 63: If I read the [docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#create-an-account) and the [twine github action](https://github.com/pypa/gh-action-pypi-publish/blob/master/twine-upload.sh#L7) correctly, the password should have a `pypi-` prefix.

Lee-W: Yes, but it's included as you generate the api token.

Woile: Cool

PR #163, opened 2020-04-20
Command changelog wip

Woile:
TODO:

- [x] Create partial changelog from latest version. This allows people to modify the changelog manually.
- [x] Include `--changelog` flag in bump command


commitizen/changelog.py
```diff
@@ -131,3 +150,82 @@ def generate_block_tree(block: List[str]) -> Dict:
 def generate_full_tree(blocks: Iterable) -> Iterable[Dict]:
     for block in blocks:
         yield generate_block_tree(block)
+
+
+def get_commit_tag(commit: GitProtocol, tags: List[GitProtocol]) -> Optional[GitTag]:
```
Woile, line 155: @Lee-W maybe you can help me here. I wanted to create a `Protocol` here, but mypy is failing to detect it. 
By Protocol I understood that I could say: "What's being received as GitProtocol should have the attribute `rev`". But for some reason, mypy sees it as a normal type.
Any idea how can I annotate this?

Lee-W: We could use `def get_commit_tag(commit: GitProtocol, tags: List[GitProtocol]) -> Optional[GitProtocol]:`. I'm still learning this concept haha 😆 The protocol seems to be a less strict type, so if you annotate it as a stricter type (i.e., `GitTag`), it cannot guarantee that's correct.

BTW, could you rebase master back to the `command-changelog`  branch. I notice that I forget to do so for a while and you just push some commits to the branch.

Woile: I removed my commits from `command-changelog`, because the idea was to make a PR to that branch haha.  But I'l rebase it and this one then

Woile: But the thing regarding Protocol, is that is not working as I intend. Maybe I have to configure mypy to be less strict... I haven't check that.

Even the example from the [PEP](https://www.python.org/dev/peps/pep-0544/#type-and-class-objects-vs-protocols) is linting errors with mypy:

```
from typing import Any, Protocol

class ProtoA(Protocol):
    def meth(self, x: int) -> int: ...
class ProtoB(Protocol):
    def meth(self, obj: Any, x: int) -> int: ...

class C:
    def meth(self, x: int) -> int: ...

a: ProtoA = C  # Type check error, signatures don't match!
b: ProtoB = C  # This should be OK but still lints an error

```

I'm also familirizing with it, will try to find a solution

Lee-W: I already rebase the master branch back to the `command-changelog`.

The warning does seem to be related to a stricter mypy setting. It still failed when I use a less strict config.

It seems to be an error on the page. By changing the last 2 lines, it will pass the linter. It makes sense because a and b are an instance of `C` instead of the type `C`

```python
a: ProtoA = C()
b: ProtoB = C()
```

Back to the commitizen part, I still don't think mypy can infer `def get_commit_tag(commit: GitProtocol, tags: List[GitProtocol]) -> Optional[GitTag]:`. Because it could be `GitTag`, `GitCommit`, `GitObject`  or even other kinds of object that has the same signature. mypy could not be sure which type it should be.

Woile:
@Lee-W this is almost ready, I need to add more tests and fix the mypy errors, maybe you wanna run it.

Lee-W:
@Woile I just test it. It looks great! It seems you miss the `install_requires` `devtools` (or maybe it's not needed when we actually publish it?). Also, the latest version it generates is **v1.16.4 (2020-03-03)** instead of 1.18.2 or 1.18.3.

Woile:
Sorry, I'm gonna remove the `debug`

> Also, the latest version it generates is v1.16.4 (2020-03-03) instead of 1.18.2 or 1.18.3.

What are the steps to reproduce this?

Lee-W:
> What are the steps to reproduce this?

Checkout this branch and run `cz changelog --incremental`

Woile:
@Lee-W now is quite complete, I think it's ready to be merged into `changelog-command` and from there make a new release

Lee-W:
Is the default behavior now generate incremental changelog? I think this choice makes more sense. Just want to make sure everything works as you expect.

Woile:
Yes, for now I think it should be the default behavior, specially for existing projects with an old changelog. If issues araise we can always support some arguments in the bump's changelog somehow.
I will merge later today then :D :D

PR #166, opened 2020-04-22
fix(git): fix returned value for GitCommit.message when body is empty

vBLFTePebWNi6c:
GitCommit.message should return only title without excess newlines if commit body is empty.

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -21,10 +21,7 @@ class GitCommit(GitObject):
 
     @property
     def message(self):
-        message = self.title
-        if self.body:
-            message = message + f"\n\n{self.body}"
-        return message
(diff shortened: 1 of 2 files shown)


commitizen/git.py
```diff
@@ -21,7 +21,10 @@ def __init__(self, rev, title, body=""):
 
     @property
     def message(self):
-        return f"{self.title}\n\n{self.body}"
+        message = self.title
+        if self.body:
+            message = message + f"\n\n{self.body}"
+        return message
```
Lee-W, line 27: How about using `f"{self.title}\n\n{self.body}".strip()` instead?

vBLFTePebWNi6c: Yep
That will be better

Woile:
Awesome!

Issue #142, opened 2020-03-08
Feature request: `commitizen` should support revert commits


lucatrv:
I know that a similar request was discussed [before](https://github.com/Woile/commitizen/issues/36), however [`angular` does contemplate revert commits](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#revert).

As a minimum, `cz check` should accept commit messages beginning with `revert:`, otherwise when it is configured as a `pre-commit` hook it does not allow revert commits.

Moreover, IMHO a `cz revert` command should be added, accepting the `<commit>` to revert. It should run `git revert <commit>`, including in the message body: `This reverts commit <hash>.` as per `angular` instructions. See information on [`git revert`](https://www.atlassian.com/git/tutorials/undoing-changes/git-revert).

For examples of revert commit messages, see for instance [`standard-version`](https://github.com/conventional-changelog/standard-version/commits/master) and [`conventional-changelog`](https://github.com/conventional-changelog/conventional-changelog/commits/master) commit logs.

Notice that both [`commitlint`](https://github.com/conventional-changelog/commitlint#what-is-commitlint) and [`gitlint`](https://jorisroovers.com/gitlint/contrib_rules/#ct1-contrib-title-conventional-commits) support revert commits.

lucatrv:
I add that IMHO, other than `revert`, also the `chore` type should be added. In fact the [Conventional Commits website](https://www.conventionalcommits.org/) clarifies that their specifications are **inspired** by the Angular Commit Guidelines (so not identical), and reports as an example the [@commitlint/config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional) recommendations, which include also `chore` and `revert` types.

lucatrv:
I now noticed the following definition of the `schema_pattern` function in `conventional_commits.py`:

```python
def schema_pattern(self) -> str:
    PATTERN = (
        r"(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)"
        r"(\(\S+\))?:\s.*"
    )
    return PATTERN
```

So it includes both `revert` and `chore`.

Lee-W:
To have `revert` and `chore` type, we'll have to add it [here](https://github.com/Woile/commitizen/blob/v1.17.0/commitizen/cz/conventional_commits/conventional_commits.py#L38) as well.

Lee-W:
[...]

PR #168, opened 2020-04-22
fix(parsing): add option for improvement type

fabiopedrosa:
Simple PR to add the greatly missed `improvement` type


Lee-W:
Hi, thanks for contributing! As discussed in #142, we're not planning on adding new types to cz_conventional_commit. We're using the latest [types from Angular](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)

fabiopedrosa:
@Lee-W ok, I see.

PR #169, opened 2020-04-22
Adding a shell return code to cmd output

alozovskoy:
Adding a shell return code to cmd output for simplify command results checking.

Feel the difference:

```
# before
>>> from commitizen import cmd
>>> cmd.run('true')
Command(out='', err='', stdout=b'', stderr=b'')
>>> cmd.run('false')
Command(out='', err='', stdout=b'', stderr=b'')

# after
>>> from commitizen import cmd
>>> cmd.run('true')
Command(out='', err='', stdout=b'', stderr=b'', rc=0)
>>> cmd.run('false')
Command(out='', err='', stdout=b'', stderr=b'', rc=1)
```


Woile:
I like the error code in the cmd :+1: , but only if it were used around the code, at the moment the app works without it. I would avoid including stuff that's not being used.

I'd be willing to merge this if the current code using `if c.err` as a check were replaced by this new `rc` and the test pass, that'd be a good refactor for the app.

And the commits need to fit conventional commits (in order to auto-bump the version), or we will squash the PR with a relevant commit for the CI.

Lee-W:
Nice, then we can go with it. But, I would recommend using `return_code` instead of `rc`.

Lee-W:
This PR is inactive. I'll close it for now.

PR #170, opened 2020-04-24
docs(third-party): add third-party template documentation

Crystalix007:
Add a new docs section for third-party contributed commit templates.
This way new templates do not need to be merged into the main codebase,
and instead exist as satellite repositories.

I wasn't exactly sure if you wanted all the third-party commit templates in the `README.md` itself, so I placed them inside a separate `docs` file instead.

See #158 for motivation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -102,6 +102,10 @@ commands:
     init                init commitizen configuration
 ```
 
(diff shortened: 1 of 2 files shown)


Woile:
Thanks

Issue #164, opened 2020-04-21
`cz bump` fails when using pre-commit to check commit message


Lee-W:
## commitizen version
1.18.1

## How to reproduce

```sh
mkdir bump_test
cd bump_test
git init
cz init

touch .pre-commit-config.yaml
echo "repos:\n \
  - repo: https://github.com/Woile/commitizen\n\
    rev: v1.18.1\n\
    hooks:\n\
      - id: commitizen\n\
        stages: [commit-msg]" > .pre-commit-config.yaml

pre-commit install -t commit-msg
git add .
cz commit
cz bump
```

## Error Message

```text
git.commit errror: "commitizen check.........................................................Passed"
```

Lee-W:
Currently, I cannot find out the reason for this error. The output message from pre-commit is not output to stderr. But we still get the message "commitizen check.........................................................Passed" from stderr.

We can use `git commit --no-verify` to bypass this error for the "bump" command since we don't have to check when bumping. @Woile What do you think?

Issue #157, opened 2020-04-10
CI is skipping commits with the word "bump"


Woile:
**Goal**
Allow execution of auto bump when writing the word "bump" in the commit message.

**Description**
Currently we are skipping a bump if the message contains the word `bump`, but it should be skipping if it begins with `bump:`. [see line](https://github.com/commitizen-tools/commitizen/blob/master/.github/workflows/bumpversion.yml#L10)

**Suggestions**
Change from 
`if: "!contains(github.event.head_commit.message, 'bump')"`
to
`if: "!contains(github.event.head_commit.message, 'bump:')"`

This will only skip commits auto-generated because they have a `:`.

Another option is to is to use the [begins with](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#startswith) from github actions or something similar

PR #183, opened 2020-05-04
feat(bump): add optional --no-verify argument for bump command

kudlatyamroth:
fix: #164

When we have in repo some pre-commit hooks, like black, mypy or anything like that
cz bump fails, even if pre-commit hooks passed.

it is strange as pre-commit write messages to sys.stdout.buffer not stderr...

example pre-commit config:
```
repos:
  - repo: local
    hooks:
      - id: flake8
        name: flake8
        stages: [commit]
        language: system
        entry: poetry run flake8
        types: [python]
        exclude: setup.py

      - id: mypy
        name: mypy
        stages: [commit]
        language: system
        entry: poetry run mypy
        types: [python]
        pass_filenames: false
```

This PR add optional --no-verify argument to bump command

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -93,6 +93,12 @@ data = {
                         "help": "generate the changelog for the newest version",
                     },
                     {
+                        "name": ["--no-verify"],
+                        "action": "store_true",
+                        "default": False,
+                        "help": "this option bypasses the pre-commit and commit-msg hooks",
+                    },
+                    {
                         "name": "--yes",
                         "action": "store_true",
                         "help": "accept automatically questions done",
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -31,6 +31,7 @@ class Bump:
         }
         self.cz = factory.commiter_factory(self.config)
         self.changelog = arguments["changelog"]
+        self.no_verify = arguments["no_verify"]
 
     def is_initial_tag(self, current_tag_version: str, is_yes: bool = False) -> bool:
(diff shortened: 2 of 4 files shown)


kudlatyamroth:
@Woile what do you think about that?
personally i think its hacky way to fix this, but can't find out why stdout from pre-commit is taken as stderr...

Lee-W:
I also encounter the same problem. I'm thinking of adding a `--no-verify` argument to `cz bump` command to make it optional for users.

kudlatyamroth:
@Lee-W i changed this pr to make --no-verify as optional argument :)

commitizen/cli.py
```diff
@@ -92,6 +92,12 @@
                         "default": False,
                         "help": "generate the changelog for the newest version",
                     },
+                    {
+                        "name": ["--no-verify"],
+                        "action": "store_true",
+                        "default": False,
+                        "help": "disable git hooks during bumping",
```
Lee-W, line 99: I just take a look of the documentation of git. The help message is 

```text
--no-verify
    This option bypasses the pre-commit and commit-msg hooks.
```

I think it might be better if we can align with them

kudlatyamroth:
@Lee-W @Woile 
description changed, and test  added :)
also updated with master :)

kudlatyamroth:
@Lee-W thanks for merge :)
but there is no release. propably this line:
https://github.com/commitizen-tools/commitizen/actions/runs/96287380/workflow#L10

should check for `bump:` or `bump: version`

Lee-W:
@kudlatyamroth It's probably due to #157 I'll take a look tomorrow

PR #185, opened 2020-05-05
ci(github/bumpversion): fix workflow to not skip commits with bump word

kudlatyamroth:
fix: #157

skip only commits that startsWith: `bump:`


kudlatyamroth:
wasn't sure if push it as new commit or amend last one... :thinking:

Woile:
It's fine, thanks a lot for the help :D

PR #187, opened 2020-05-12
feature: maps and hooks

Woile:
## Detail

Add support for map and hooks
## Checklist:

- [x] Run "./script/lint" and "./script/test" locally to ensure this change passes linter check and test
- [x] Add test cases to all the changes

### DIFF
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -73,7 +73,7 @@ def generate_tree_from_commits(
     changelog_pattern: str = defaults.bump_pattern,
     unreleased_version: Optional[str] = None,
     change_type_map: Optional[Dict[str, str]] = None,
-    message_hook: Optional[Callable] = None,
+    changelog_message_builder_hook: Optional[Callable] = None,
 ) -> Iterable[Dict]:
     pat = re.compile(changelog_pattern)
     map_pat = re.compile(commit_parser)
@@ -120,8 +120,8 @@ def generate_tree_from_commits(
 
             if change_type_map:
                 change_type = change_type_map.get(change_type, change_type)
-            if message_hook:
-                parsed_message = message_hook(parsed_message, commit)
+            if changelog_message_builder_hook:
+                parsed_message = changelog_message_builder_hook(parsed_message, commit)
             changes[change_type].append(parsed_message)
         if message_body:
             parsed_message_body: Dict = message_body.groupdict()
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -29,7 +29,7 @@ import os
 import re
 from collections import defaultdict
 from datetime import date
-from typing import Dict, Iterable, List, Optional
+from typing import Callable, Dict, Iterable, List, Optional
 
 import pkg_resources
 from jinja2 import Template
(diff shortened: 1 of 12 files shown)


commitizen/cz/base.py
```diff
@@ -27,7 +27,14 @@ class BaseCommitizen(metaclass=ABCMeta):
     # It can be modified per rule
     commit_parser: Optional[str] = r"(?P<message>.*)"
     changelog_pattern: Optional[str] = r".*"
-    changelog_map: Optional[dict] = None  # TODO: Use it
+    change_type_map: Optional[dict] = None
+
+    # Executed per message parsed by the commitizen
+    message_hook: Optional[Callable] = None  # (dict, GitCommit) -> dict
```
Lee-W, line 33: Maybe we could use annotation like `Optional[Callable[[dict, GitCommit], dict]`

commitizen/cz/base.py
```diff
@@ -27,7 +27,14 @@ class BaseCommitizen(metaclass=ABCMeta):
     # It can be modified per rule
     commit_parser: Optional[str] = r"(?P<message>.*)"
     changelog_pattern: Optional[str] = r".*"
-    changelog_map: Optional[dict] = None  # TODO: Use it
+    change_type_map: Optional[dict] = None
```
Lee-W, line 30: I would prefer using more explicit annotaion `Dict[str, str]`

tests/commands/test_changelog_command.py
```diff
@@ -234,3 +236,26 @@ def test_changlog_incremental_keep_a_changelog_sample(mocker, capsys):
...
+
+@pytest.mark.usefixtures("tmp_commitizen_project")
+def test_changlog_hook(mocker, config):
+    hook_mock = mocker.Mock()
+
+    def changelog_hook(full_changelog: str, partial_changelog: Optional[str]) -> None:
+        if partial_changelog:
+            hook_mock()
+        if full_changelog:
+            hook_mock()
```
Lee-W, line 249: How about sending `full_changelog` and `partial_changelog` into `hook_mock` and use `assert_called_with`?

commitizen/changelog.py
```diff
@@ -72,6 +72,8 @@ def generate_tree_from_commits(
     commit_parser: str,
     changelog_pattern: str = defaults.bump_pattern,
     unreleased_version: Optional[str] = None,
+    change_type_map: Optional[Dict[str, str]] = None,
+    message_hook: Optional[Callable] = None,
```
Lee-W, line 76: I'm not sure whether we should name it as `message_hook`. It looks like `changelog_hook` and makes me think it as a webhook intuitivly but it actually has a different functionality.

commitizen/cz/base.py
Woile: Cool, much better actually haha. My intention was there with the comment :rofl:

commitizen/changelog.py
Woile: mmmm :thinking: some alternatives I can think of:

- `per_commit_hook`
- `commit_hook`
- `changelog_line_hook`

any other ideas?

Woile:
Addressed most of the suggestions,  only one is missing

commitizen/changelog.py
Lee-W: Hmmm... My first though was `message_builder`. Or maybe `commit_preprocess_hook`?

Woile: `commit_preprocess_hook`: gives me the impression of a precommit hook, or something unrelated to changelog. 

What about `changelog_message_builder_hook`? it may be long but goes to the point

Lee-W: `changelog_message_builder_hook`  looks good to me.

Woile:
If no more comments I'll merge it :D

Issue #128, opened 2020-01-29
[test] check the error code when raising System Exit


Lee-W:
**Goal**
make unit tests more accurate

**Description**
In [tests/commands/test_bump_command.py#L86](https://github.com/Woile/commitizen/blob/v1.16.0/tests/commands/test_bump_command.py#L86), we do not check the error code. It's possible to raise the same exception but not our expected case. Maybe we should refactor them as custom exceptions?

Woile:
Yes, custom exceptions makes sense. But this would be a breaking change, we can schedule it for 2.0

Lee-W:
Maybe we can fix it in two steps. The first step is to check only the error code and merge it into `1.*`. The second step is to refactor it into custom exceptions and release it in `2.0`.

Woile:
To be honest, I don't see much value in adding the checks for `1.x`. The error code thing is an idea I had to give people flexibility to decide what to do on each error when writing configuration scripts, but I've never seen anyone do conditions on error in bash, even less with commitizen.

So I say we don't waste time on it (be my guest if you want to do it, but I don't think it really adds value). And refactor into more expressive custom exceptions for 2.0

Lee-W:
Yes, that makes sense. Do we have any plan on when or how we release 2.0? Maybe a branch `2.0` so we can merge `2.0-only` commits into the branch?

Woile:
I've created a branch `next`, we can start sending PR there

Lee-W:
I think the error code is quite useful when I use some of the new features recently. I'm thinking of refactoring these errors into classes but still have the same error code which makes it a non-breaking change. @Woile What do you think?

Woile:
Great idea! Go for it!

Woile:
@Lee-W can we close this ticket?

Lee-W:
Ah yes, I forgot it haha. Let's close it!

PR #201, opened 2020-06-05
refactor: use custom exception for error handling

Lee-W:
the error code used is embedded into these exceptions

(Thanks for sending a pull request! Please fill in the following content to let us know better about this change.)

## Types of changes

- [x] **Refactoring**
- [x] **Documentation Update**

## Description
* use custom exception for error handling
* do not show traceback if the raised exception is `CommitizenException`
* introduce `DryRunExit`, `ExpectedExit`, `NoCommandFoundError`, `InvalidCommandArgumentError` (Remove all the `SystemExit`)
* remove unused `NO_COMMIT_MSG` error code

### TODO
* [x] Handle print message
* [ ] Add documentation about how to use error code
* [ ] Group `CommitizenException` (e.g. git group, command line group and etc.)

## Checklist:
- [x] Run `./script/lint` and `./script/test` locally to ensure this change passes linter check and test

Try to execute the commands that will raise exceptions. e.g., `cz bump --dry-run`

## Expected behavior
No behavior should change after this change

## Related Issue
#128 

## Additional context
Add any other context or screenshots about the pull request here.


Lee-W:
@Woile I've finished the major part of this refactoring. Could you please take a look at it? 🙂 I'm still working on the error message part. I'll also add documentation for these error codes.

commitizen/cli.py
```diff
@@ -237,6 +238,18 @@
     },
 }
 
+original_excepthook = sys.excepthook
+
+
+def commitizen_excepthook(type, value, tracekback):
```
Woile, line 244: neat! cool!
What do you think if the exception also handles the error message?

Instead of:
```
out.error(f"'{self.config.settings['name']}' rule does not support bump")
raise NoPatternMapError()
```

```
raise NoPatternMapError(f"'{self.config.settings['name']}' rule does not support bump")
```

and this hook takes care of `out.error` if message exists.

Lee-W: Yes, that's what I'm currently working on! I'm thinking of adding a default message while supporting a customized message. Also, some of these exceptions only show as a warning.

BTW, I notice `--debug` is deprecated after 2.0. May I know the reason of it? I'm thinking of keeping the `--debug` flag and handle it through expcetionhook so that users can get extra information (like traceback) when encountering an error

Woile: There wasn't much for debugging purposes haha but your idea seems great, I like it :+1:

Lee-W:
@Woile This pr is finally ready for reviewing 🎉

Issue #209, opened 2020-06-26
[Documentation] Add instructions on how to use commitizen with pre-commit in README.md


altusgerona:
## Type
* [ ] Conent inaccurate
* [x]  Content missing
* [ ] Typo

## Description
We should add a small snippet for `pre-commit` integration with Commitizen. I realize that this is in the docs, would be nice to include it in the `README.md` as well for an easy reference to first-time viewers.

Would be happy to submit a PR if needed. Would love for `commitizen` to be more widely used.

Woile:
Feel free to submit a PR :smiley: 

Thanks! We too!

altusgerona:
I just realized this is sort of an extension of #196. The resolution for that was the full documentation link to the README.md. I'll make this issue specifically about visibility in the README.md. Will submit a PR in a few.

Lee-W:
Since this issue has been raised more than once, maybe there is a better way to structure/link the document. We can discuss more on this issue or your PR 🙂

altusgerona:
We can close this now due to #210 being merged.

Issue #196, opened 2020-06-02
Need doc for pre-commit integration


guihao-liang:
## Type
* [x] Content missing

## URL

My understanding is that this tool can only work in `commit-msg`  and `prepare-commit-msg` stage based on this [doc](https://pre-commit.com/#pre-commit-for-commit-messages).

But I haven't found any doc showing me how to use this nice tool with `pre-commit` tool.

## Description

I think it's better to provide a doc to explicitly mention this in README.md.

```
 27 - repo: https://github.com/commitizen-tools/commitizen
 28     rev: v1.22.2
 29     hooks:
 30       - id: commitizen
 31         stages: ["commit-msg"]
```

When used, it should use `pre-commit install -t commit-msg`.

guihao-liang:
I can add this to README if you think it's appropriate to add this to there.

Lee-W:
It can be found [here](https://commitizen-tools.github.io/commitizen/check/).

guihao-liang:
> Thank you for your suggestion! But it can be found [here](https://commitizen-tools.github.io/commitizen/check/). Or do you think it's not that easy to find? We can discuss more on how to make this better 🙂

Wow! I didn't find that. Yeah, it's good to provide a link to this in the README.md.

Lee-W:
I just add a link to our GitHub page. I'll close this issue for now. We can discuss more if you think we can do better. Also, feel free to send us a pr 🙂

PR #210, opened 2020-06-27
Precommit docs

altusgerona:
## Types of changes

- [ ] **Bugfix**
- [ ] **New feature**
- [ ] **Refactoring**
- [ ] **Breaking change** (any change that would cause existing functionality to not work as expected)
- [x] **Documentation Update**
- [ ] **Other (please describe)**

## Description
Add instructions for pre-commit integration in the README.md itself. 

## Checklist:
- [x] Run `./script/lint` and `./script/test` locally to ensure this change passes linter check and test

## Related Issue
#196  #209

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -81,7 +81,7 @@ You can integrate this in your [pre-commit](https://pre-commit.com/) config with
 ---
 repos:
   - repo: https://github.com/commitizen-tools/commitizen
-    rev: v1.17.0
+    rev: master
     hooks:
       - id: commitizen
         stages: [commit-msg]
--- a/docs/README.md
+++ b/docs/README.md
@@ -74,6 +74,20 @@ or the shortcut
 cz c
 ```
 
+### Integrating with Pre-commit
+Commitizen can lint your commit message for you with `cz check`.
+You can integrate this in your [pre-commit](https://pre-commit.com/) config with:
+```yaml
+---
+repos:
+  - repo: https://github.com/commitizen-tools/commitizen
+    rev: v1.17.0
+    hooks:
+      - id: commitizen
+        stages: [commit-msg]
+```
+Read more about the `check` command [here](https://commitizen-tools.github.io/commitizen/check/).
+
 ### Help
 
 ```bash
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,7 @@ venv.bak/
 # mypy
 .mypy_cache/
 
+.idea
 .vscode/
 *.bak
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -8,13 +8,13 @@ If you're a first-time contributor, you can check the issues with [good first is
 
 ## Before making a pull request
 
-1. Fork [the repository](https://github.com/commitizen-tools/commitizen)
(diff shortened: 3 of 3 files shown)


docs/README.md
```diff
@@ -74,6 +74,20 @@ or the shortcut
...
 ```
 
+### Integrating with Pre-commit
+Commitizen can lint your commit message for you with `cz check`.
+You can integrate this in your [pre-commit](https://pre-commit.com/) config with:
+```yaml
+---
+repos:
+  - repo: https://github.com/commitizen-tools/commitizen
+    rev: v1.17.0
```
Lee-W, line 84: I would suggest updating the version to the latest commitizen

docs/README.md
```diff
@@ -74,6 +74,20 @@ or the shortcut
 cz c
 ```
 
+### Integrating with Pre-commit
+Commitizen can lint your commit message for you with `cz check`.
+You can integrate this in your [pre-commit](https://pre-commit.com/) config with:
+```yaml
```
Lee-W, line 80: Adding this config file is not enough for `pre-commit`, I'd suggest add the `pre-commit install --hook-type commit-msg` command

docs/README.md
altusgerona: Wouldn't it be better if we just leave it at `master`. That's actually what I use for our company projects.

docs/README.md
altusgerona: Hmm. This already assumes that the dev has run `pre-commit` before tho. When I was trying this out for myself I just added that snippet to my config and reran pre-commit. I think the `stages: [commit-msg]` is enough.

docs/README.md
Lee-W: My reason for not doing so is that pre-commit caches repos. You might not notice that you're still using the older version.

altusgerona: True, but it's the same with using a fixed version. Also, this is the README. We'd have to update the version every time we bump.

altusgerona: I think we'll be documenting `pre-commit` for them with that. It's already documented by pre-commit themselves. But if you guys want I'll put it in. Just let me know.

Woile:
Can't we add to pyproject.toml

```
version_files = [
  "pyproject.toml:version",
  "commitizen/__version__.py",
  "docs/README.md:\trev:"
]
```

or something like that?

altusgerona:
> Can't we add to pyproject.toml
> 
> ```
> version_files = [
>   "pyproject.toml:version",
>   "commitizen/__version__.py",
>   "docs/README.md:\trev:"
> ]
> ```
> 
> or something like that?
Sure. But I think it'll just make it more verbose than it has to be. Other python libs that use pre-commit—pre-commit included just use `master` or `stable`.

Woile:
`master` is fine by me as well, I have no issues with it

docs/README.md
altusgerona: Hey @Lee-W, just to follow up, you're right. It needs the `pre-commit install --hook-type commit-msg` command.

Lee-W: Ah, I miss this part when I merged this PR.  No worries. I'll add it in another PR.

PR #215, opened 2020-07-01
Runtime dependency on setuptools

blaggacao:
Imported here: https://github.com/commitizen-tools/commitizen/blob/b8fc1eeaf5a915d8d0f7179dc303ec2f3835b79f/commitizen/changelog.py#L33

## Types of changes

- [x] **Bugfix**

Without it in  a pristine python environment: 

```shell
cz --help
  File "/nix/store/8wvn1d6hfdcribcinn21rzlkvjvzab2p-python3.8-commitizen-1.23.0/bin/.cz-wrapped", line 6, in <module>
    from commitizen.cli import main
  File "/nix/store/k4h5dy9yqdm8gisdmbhzg75m06p1fb9z-python3-3.8.3-env/lib/python3.8/site-packages/commitizen/cli.py", line 9, in <module>
    from commitizen import commands, config
  File "/nix/store/k4h5dy9yqdm8gisdmbhzg75m06p1fb9z-python3-3.8.3-env/lib/python3.8/site-packages/commitizen/commands/__init__.py", line 1, in <module>
    from .bump import Bump
  File "/nix/store/k4h5dy9yqdm8gisdmbhzg75m06p1fb9z-python3-3.8.3-env/lib/python3.8/site-packages/commitizen/commands/bump.py", line 7, in <module>
    from commitizen.commands.changelog import Changelog
  File "/nix/store/k4h5dy9yqdm8gisdmbhzg75m06p1fb9z-python3-3.8.3-env/lib/python3.8/site-packages/commitizen/commands/changelog.py", line 6, in <module>
    from commitizen import changelog, factory, git, out
  File "/nix/store/k4h5dy9yqdm8gisdmbhzg75m06p1fb9z-python3-3.8.3-env/lib/python3.8/site-packages/commitizen/changelog.py", line 33, in <module>
    import pkg_resources
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ termcolor = "^1.1"
 packaging = ">=19,<21"
 tomlkit = "^0.5.3"
 jinja2 = "^2.10.3"
+setuptools = ">=47.3.1"
 
 [tool.poetry.dev-dependencies]
 ipython = "^7.2"


blaggacao:
Not sure about the correct version constraint: please amend.

pyproject.toml
```diff
@@ -52,6 +52,7 @@ termcolor = "^1.1"
 packaging = ">=19,<21"
 tomlkit = "^0.5.3"
 jinja2 = "^2.10.3"
+setuptools = ">=47.3.1"
```
blaggacao, line 55: What would be a more appropriate version constraint? (this is latest on pypi as of today)

Woile:
Could you provide the full traceback? Thanks!

blaggacao:
> Could you provide the full traceback? Thanks!

Actually that's the full traceback when running `cz --help` - What where you expecting?

Woile:
I was expecting an exception at the end. It's not clear why it fails.
Which OS are you on?

blaggacao:
Oh sorry! Might be that I clipped that unintentionally.

The exception was that `pkg_resources` is not importable.
This modules is made available by `setuptools`.

In a python environment without `setuptools` installed this results in an error as per the dependencies declared in the `pyproject.toml`.

In such a environment, the truth is: `setuptools` is a runtime dependency because of this import.

The python environment is the one crafted by the following nix expression:

```
   4   │   mach-nix = import (
   5   │     builtins.fetchGit {
   6   │       url = "https://github.com/DavHau/mach-nix/";
   7   │       ref = "2.0.1";
   8   │     }
   9   │   );
  10   │
  11   │   commitizenPython = mach-nix.mkPython {
  12   │     python = pkgs.python38;
  13   │     requirements = ''
  14   │       commitizen
  15   │       setuptools
  16   │     '';
  17   │   };
```

What this does:
1. takes a bare `pkgs.python38`
2. adds `commitizen` (from pypi)
3. additionally declares (unexposed) dependency `setuptools` - this is the motivation for this PR

With this setup it works. If I left `setuptools` undeclared, it would result in the exception.

In most common standard python environments setuptools might be shipeed or available.
But that doesn't change the fact that it still is a runtime dependency.

blaggacao:
Note: https://github.com/DavHau/mach-nix/ has it's build in completely resolved dependency graph of all pypi packages (250MB database). This database is published every 12 hours or so and used for static dependency resolution. Since `setuptools` is not declared, the tool cannot generate the complete runtime closure for `commitizien` and I have to complement it manually.

blaggacao:
@Woile Does my latter explanation make sense? I still wonder what an appropriate version of `setuptools` would be, though.

Woile:
Hey, yes, thanks, I gotta research now, I don't want to run in the issue that it doesn't work in a machine which already has `setuptools`.

blaggacao:
Yeah, I understand. I think tagging the lowest possible setuptools version, which also implements `pkg_resouces`, upwards could be pretty safe.

Woile:
We have removed the dependency :tada: You can try the latest version :muscle:

Issue #211, opened 2020-06-27
[Feature Request] Add pre-commit integration


altusgerona:
## Description
Currently, the contributing guide states that we run the lints and tests manually before resolving a commit. Why not force this already by using [pre-commit](https://pre-commit.com/)?

## Possible Solution
Add pre-commit config file and dev dependency.

Lee-W:
do you mean generate it when running `cz init`?

altusgerona:
Nope. Basically adding `pre-commit-config.yaml` to the repository and having contributors run `pre-commit install` before running a commit to make sure that their staged changes and commit message are linted and follow convention.

altusgerona:
We could also add `pre-commit install` to poetry scripts so it enforces the standard.

Lee-W:
> Nope. Basically adding `pre-commit-config.yaml` to the repository and having contributors run `pre-commit install` before running a commit to make sure that their staged changes and commit message are linted and follow convention.

Yep, it sounds good to me 🚀


> We could also add `pre-commit install` to poetry scripts so it enforces the standard.

Do you mean to set it up for contributors when they run `poetry install`? How could we achieve it?

altusgerona:
I was thinking maybe using [`poetry` scripts](https://python-poetry.org/docs/pyproject/#scripts) but that looks like it doesn't behave the same way as package.json's `postinstall`. It's okay though. We should be enforcing the convention at CI level too. So before tests start, `pre-commit run -a` should not fail.

Lee-W:
We already have those scripts under `scripts`. I think I'll have the time to set up the pre-commit config late today.

altusgerona:
The `scripts` would be redundant with the pre-commit config. I can also submit a PR if you'd like.

Lee-W:
#216 merged.

PR #216, opened 2020-07-01
Add pre commit integration

Lee-W:
(Thanks for sending a pull request! Please fill in the following content to let us know better about this change.)

## Types of changes

- [ ] **Other (please describe)**: CI

## Description
Add pre-commit config and fix styles through it

## Checklist:
- [x] Run `./script/reformat` and `./script/test` locally to ensure this change passes linter check and test

```sh
pre-commit install
# make some change
cz commit
git push ...
```

## Expected behavior
* pre-commit should fix trailing space and eof when commiting
* pre-commit should run `./script/reformat` and `./script/test` when pushing

## Related Issue
#211

## Additional context

image


Woile:
I very much prefer `format` (like [fastapi](https://github.com/tiangolo/fastapi/blob/master/scripts/format.sh)) instead of `reformat` which feels redundant.

Woile:
I'm not sure about adding pre-commit Lee. It will be an entry barrier, and it's not adding much. It will be a pain-point during the sprints.

From my perspective:
- Our CI catches lint and test errors just fine.
- We don't make mistakes often when creating PRs to justify pre-commit.
- We can squash commits from contributors if they don't match conventional commits.

What do you think about it?

Lee-W:
> I very much prefer `format` (like [fastapi](https://github.com/tiangolo/fastapi/blob/master/scripts/format.sh)) instead of `reformat` which feels redundant.

I totally agree with it 😄


> I'm not sure about adding pre-commit Lee. It will be an entry barrier, and it's not adding much. It will be a pain-point during the sprints.
> 
> From my perspective:
> 
>     * Our CI catches lint and test errors just fine.
> 
>     * We don't make mistakes often when creating PRs to justify pre-commit.
> 
>     * We can squash commits from contributors if they don't match conventional commits.
> 
> 
> What do you think about it?

I agree with your points of view but also like pre-commit 😆. I don't have a strong option on whether to use pre-commit after reading your comments. Here's my thought on why using pre-commit might be good and might not be a huge entry barrier. 🙂

1. pre-commit is optional. We're not able to force contributors to install it when they're to contribute. It's a kinda helper for them to meet our requirements. But even if they don't, we could still handle it as you said. 
2. As a contributor, I like to learn best practices from contributing to open source projects, especially in a sprint. Even if the contributors run into a problem, I don't think it will take to much of our time to instruct them during the sprint.
3. As a maintainer, it can save some of our time to do those trivial kinds of stuff.

Woile:
You are right, but everything we add adds more maintenance, plus what's the point of the extra noise if not used?
I remember my first contributions/projects and there were things there were super confussing and I was lost in a sea of questions haha (specially with tox).
But okay, If you think it will really add value to us, I'm in for it, we can try it (worst case if we feel like it's not adding value, we can remove it as it's not part of the src code).

Lee-W:
Yes, I think it will add value to us. 🙂 But I also understand your concern. We can test it in euro python 2020 sprint and pycon tw 2020 sprint (I will host one in Taiwan 🇹🇼) to see whether we will encounter the problem you mention. I'll merge it for now.

Issue #175, opened 2020-04-30
Show all ill-formatted commits when running batch cz check


Lee-W:
`cz check --rev-range master..HEAD` currently only shows the latest commits that do not follow the pattern. It would be more convenient if these ill-formatted are showed at once

josix:
Hi, is this issue currently processing ? May I try to solve this issue?

Woile:
Fine by me, I don't know if @Lee-W is working on this

Lee-W:
@josix Sure! I've not yet worked on this one.

josix:
Thank you for your quick reply! I'll make a draft PR first to let you be able to track my progress.

Lee-W:
We're looking forward to it 🎉

Lee-W:
Fix by #244

PR #244, opened 2020-08-03
feat(cz_check): Update to show all ill-formatted commits

josix:
## Description

Update `cz check --rev-range` to show all ill-formatted commits when running batch checking

## Checklist

(Not sure if I need to add test case and update the docs)

## Expected behavior

All ill-formatted commits should be displayed when running `cz check --rev-range`

1. Go to any repo containing ill-formatted commits
2. Run `cz check --rev-range HEAD~10..HEAD` to check if all ill-formatted commits in the range is displayed (`HEAD~10..HEAD` could be replaced with any valid argument)

## Additional context

Solve Issue #175

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -50,14 +50,17 @@ class Check:
             raise NoCommitsFoundError(f"No commit found with range: '{self.rev_range}'")
 
         pattern = self.cz.schema_pattern()
+        ill_formated_commits = []
         for commit_msg in commit_msgs:
             if not Check.validate_commit_message(commit_msg, pattern):
-                raise InvalidCommitMessageError(
-                    "commit validation: failed!\n"
-                    "please enter a commit message in the commitizen format.\n"
-                    f"commit: {commit_msg}\n"
-                    f"pattern: {pattern}"
-                )
+                ill_formated_commits.append(f"commit: {commit_msg}\n")
+        if ill_formated_commits != []:
+            raise InvalidCommitMessageError(
+                "commit validation: failed!\n"
+                "please enter a commit message in the commitizen format.\n"
+                f"{''.join(ill_formated_commits)}\n"
+                f"pattern: {pattern}"
+            )
         out.success("Commit validation: successful!")
 
     def _get_commit_messages(self):
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -76,13 +76,7 @@ class Check:
             with open(self.commit_msg_file, "r") as commit_file:
                 commit_title = commit_file.readline()
                 commit_body = commit_file.read()
-            return [
-                git.GitCommit(
-                    rev="",
-                    title=commit_title,
-                    body=commit_body,
-                )
-            ]
+            return [git.GitCommit(rev="", title=commit_title, body=commit_body)]
 
         # Get commit messages from git log (--rev-range)
         return git.get_commits(end=self.rev_range)
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -57,7 +57,7 @@ class Check:
         ]
         displayed_msgs_content = "\n".join(
             [
-                f"commit {commit.rev}: {commit.message}"
+                f'commit "{commit.rev}": "{commit.message}"'
                 for commit in ill_formated_commits
             ]
         )
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -53,11 +53,11 @@ class Check:
         ill_formated_commits = [
             commit
             for commit in commits
-            if not Check.validate_commit_message(commit["msg"], pattern)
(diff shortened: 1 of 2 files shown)


commitizen/commands/check.py
```diff
@@ -50,14 +50,17 @@ def __call__(self):
             raise NoCommitsFoundError(f"No commit found with range: '{self.rev_range}'")
 
         pattern = self.cz.schema_pattern()
+        ill_formated_commits = []
```
Lee-W, line 53: Maybe we could try list comprehension here. I'd suggest not adding `f"commit: {...}\n"` here to keep the variable simple and handle the representation below.

```python
ill_formated_commits = [
    commit_msg
    for commit_msg in commit_msgs
    if not Check.validate_commit_message(commit_msg, pattern)
]
```

commitizen/commands/check.py
```diff
@@ -50,14 +50,17 @@ def __call__(self):
...
         for commit_msg in commit_msgs:
             if not Check.validate_commit_message(commit_msg, pattern):
-                raise InvalidCommitMessageError(
-                    "commit validation: failed!\n"
-                    "please enter a commit message in the commitizen format.\n"
-                    f"commit: {commit_msg}\n"
-                    f"pattern: {pattern}"
-                )
+                ill_formated_commits.append(f"commit: {commit_msg}\n")
+        if ill_formated_commits != []:
```
Lee-W, line 57: I think we don't need `!= []` part. `if ill_formated_commits:` alone would work.

commitizen/commands/check.py
```diff
@@ -50,14 +50,17 @@ def __call__(self):
...
+            raise InvalidCommitMessageError(
+                "commit validation: failed!\n"
+                "please enter a commit message in the commitizen format.\n"
+                f"{''.join(ill_formated_commits)}\n"
```
Lee-W, line 61: I'm thinking of adding the commit rev (e.g., SHA) here which would make the user find the commit easier. @josix What do you think?

josix: Sure! It would be great if we add the commit SHA to each ill-formatted commit. I'll add it in next commit.

commitizen/commands/check.py
josix: Thank you for your review. I've fixed it.

commitizen/commands/check.py
josix: Thanks for your review. I've fixed it.

josix:
Hello @Lee-W, thank you for your quickly review. I'll add one unittest to this update these day. Thank you for your help.

commitizen/commands/check.py
```diff
@@ -45,30 +45,43 @@ def __call__(self):
...
-                    "please enter a commit message in the commitizen format.\n"
-                    f"commit: {commit_msg}\n"
-                    f"pattern: {pattern}"
-                )
+        ill_formated_commits = [
+            commit
+            for commit in commits
+            if not Check.validate_commit_message(commit["msg"], pattern)
+        ]
+        displayed_msgs_content = "".join(
```
Lee-W, line 58: Instead of putting "\n" at the end of the string inside, I'd try `displayed_msgs_content = ""\n.join(`. Also, it might be better if we quote `rev` and `msg` in displaying messgage.

commitizen/commands/check.py
```diff
@@ -45,30 +45,43 @@ def __call__(self):
...
         if self.commit_msg_file:
             with open(self.commit_msg_file, "r") as commit_file:
                 commit_msg = commit_file.read()
-            return [commit_msg]
+            return [{"rev": None, "msg": commit_msg}]
 
         # Get commit messages from git log (--rev-range)
-        return [commit.message for commit in git.get_commits(end=self.rev_range)]
+        return [
+            {"rev": commit.rev[:8], "msg": commit.message}
```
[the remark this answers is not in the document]
josix: Oh... Sorry for that. I didn't notice `GitCommit` Class since I'm not quite familiar with all components. I've updated it, please have a look. Thank you very much.

commitizen/commands/check.py
```diff
@@ -45,30 +45,43 @@ def __call__(self):
...
         out.success("Commit validation: successful!")
 
-    def _get_commit_messages(self):
+    def _get_commits(self):
         # Get commit message from file (--commit-msg-file)
         if self.commit_msg_file:
             with open(self.commit_msg_file, "r") as commit_file:
                 commit_msg = commit_file.read()
-            return [commit_msg]
+            return [{"rev": None, "msg": commit_msg}]
```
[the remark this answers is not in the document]
josix: I've updated it, thank you for your advice.

commitizen/commands/check.py
josix: Thank you for your advice, I've corrected it. Please have a look.

josix:
@Lee-W  Thanks for your review! I learn a lot from it! 🙏

Issue #200, opened 2020-06-03
interactively pipe stream to cz check


guihao-liang:
## Description
similar to [commitlint](https://commitlint.js.org/#/), I want to interactively try commitizen with pipe.

```bash
echo "fix(parser): appending extra parse arms" | cz check
```
## Possible Solution
1. read from a string input (from stdin)
2. read from a stream (such as /dev/fd/1, stdin too, but you need to know if you are piped). [How do I know the python script is called in a pipe](https://stackoverflow.com/questions/35247817/is-there-a-way-for-a-python-script-to-know-if-it-is-being-piped-to)?

In this mode, no requirement for `git`. When I invoke cz in non-git repo, it will complain

```
$ cz check file-name
fatal: not a git repository (or any of the parent directories): .git
```
## Additional context

current spec

```
usage: cz check [-h]
                [--commit-msg-file COMMIT_MSG_FILE | --rev-range REV_RANGE]

optional arguments:
  -h, --help            show this help message and exit
  --commit-msg-file COMMIT_MSG_FILE
                        ask for the name of the temporal file that contains
                        the commit message. Using it in a git hook script:
                        MSG_FILE=$1
  --rev-range REV_RANGE
                        a range of git rev to check. e.g, master..HEAD
```

Besides the issue I mentioned, I think this spec is not clear about the **positional** argument.

Just let me know if you guys have the bandwidth or not. I can try to help during weekends

Lee-W:
This seems great! Thanks for the suggestion!
So basically, we have the following three things to do.

1. new feature: pipe stream to `cz check`
2. bug fix: check git project only on git related commands
3. doc(?): unclear positional argument (I don't quite get it. Could you please describe a bit more?)

I'm still working on other issues and might not have time to work on it soon. 😞  But I'd be happy to discuss it with you if you need any help when making this change. 😄 We can also enhance our [Contributing Guide](https://commitizen-tools.github.io/commitizen/contributing/) during the process. 🚀

guihao-liang:
> 3\. **doc(?)**: unclear positional argument (I don't quite get it. Could you please describe a bit more?)

Sure! I mean, when you invoke, `cz check args`, what do you expect `args` to be? Files? And what long option is used by default?

Lee-W:
Got it! Thanks! I'll add it after I finish what I'm working on.

Lee-W:
[...]

PR #255, opened 2020-08-31
feat(cz_check): cz check can read commit message from pipe

jordanSu:
## Description
As mentioned in #200, this PR makes `cz check` can read message from pipe. In this mode, "--message" option is used.

## Checklist
- [x] Run `./script/format` and `./script/test` locally to ensure this change passes linter check and test (**but failed**)

## Expected behavior
String input from pipe can be read by `cz check`

1. `echo COMMIT_MESSAGE | cz check`
2. `cat COMMIT_FILE | cz check`

## Additional context

#200 

This PR can't pass `./script/test`, because the test "test_check_conventional_commit_succeeds" in test_check_command.py failed. 
However, I don't understand why this test failed, because a similar test "test_check_jira_multiple_commands" succeeded.
I put a screenshot of the error message below
Screen Shot 2020-08-31 at 9 54 46 PM

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -1,5 +1,6 @@
 import os
 import re
+import sys
 from typing import Dict, Optional
 
 from commitizen import factory, git, out
@@ -32,9 +33,12 @@ class Check:
         self.cz = factory.commiter_factory(self.config)
 
     def _valid_command_argument(self):
-        if not (
-            bool(self.commit_msg_file) ^ bool(self.commit_msg) ^ bool(self.rev_range)
-        ):
+        number_args_provided = (
+            bool(self.commit_msg_file) + bool(self.commit_msg) + bool(self.rev_range)
+        )
+        if number_args_provided == 0 and not os.isatty(0):
+            self.commit_msg: Optional[str] = sys.stdin.read()
+        elif number_args_provided != 1:
             raise InvalidCommandArgumentError(
                 (
                     "One and only one argument is required for check command! "
--- a/docs/check.md
+++ b/docs/check.md
@@ -1,71 +1,47 @@
+# Check
+
 ## About
-
 This feature checks whether the commit message follows the given committing rules.
+If you want to setup an automatic check before every git commit, please refer to 
+[Automatically check message before commit](auto_check.md).
 
-## Checking before the commit
-To automatically check a commit message prior to committing, you can use a [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks). There are two common methods for installing the hook:
+## Usage
+There are three arguments that you can use one of them to check commit message.
 
-### Method 1: Add git hook through [pre-commit](https://pre-commit.com/)
-
-* Step 1: Install [pre-commit](https://pre-commit.com/)
-
-```sh
-python -m pip install pre-commit
-```
-
-* Step 2: Create `.pre-commit-config.yaml` at your root directory with the following content
-
-```yaml
----
-repos:
-  - repo: https://github.com/commitizen-tools/commitizen
-    rev: v1.17.0
-    hooks:
-      - id: commitizen
-        stages: [commit-msg]
-```
-
-* Step 3: Install the configuration into git hook through `pre-commit`
+### Git Rev Range
+If you'd like to check a commit's message after it has already been created, then you can specify the range of commits to check with `--rev-range REV_RANGE`.
 
 ```bash
-pre-commit install --hook-type commit-msg
+$ cz check --rev-range REV_RANGE
 ```
-
-### Method 2: Manually add git hook
-The command might be included inside of a Git hook (inside of `.git/hooks/` at the root of the project).
-
-The selected hook might be the file called commit-msg.
-
-This example shows how to use the check command inside of commit-msg.
-
-At the root of the project:
-
-```bash
-cd .git/hooks
-touch commit-msg
-chmod +x commit-msg
-```
-
-Open the file and edit it:
-
-```sh
-#!/bin/bash
-MSG_FILE=$1
-cz check --commit-msg-file $MSG_FILE
-```
-
-Where `$1` is the name of the temporary file that contains the current commit message. To be more explicit, the previous variable is stored in another variable called `$MSG_FILE`, for didactic purposes.
-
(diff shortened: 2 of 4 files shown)


Lee-W:
@jordanSu Hi, it seems you do not pass all the checks. If this's still a draft, you can make it a draft PR by click the button under `Reviews`
image

Lee-W:
@jordanSu hmmm... I'm not able to reproduce this error on mac. I'll see whether I can do so in a docker container

commitizen/cli.py
```diff
@@ -267,7 +268,11 @@ def main():
 
     # This is for the command required constraint in 2.0
     try:
-        args = parser.parse_args()
+        if sys.argv[1] == "check" and not os.isatty(0):
```
Lee-W, line 271: @jordanSu Could you please explain a bit on what is`os.isatty(0)` and why it's used? My best guess is that this line causes the error. Also, `sys.argv[1] == "check"` does not guarantee running `cz check` or not. e.g, (`cz -n cz_jira check --commit-msg-file some_file`) . This is why `test_check_jira_multiple_commands` passes while `test_check_conventional_commit_succeeds` does not

Lee-W (changes requested):

commitizen/cli.py
jordanSu: Sure. So in #200 guihao-liang provided a [possible solution](https://stackoverflow.com/questions/35247817/is-there-a-way-for-a-python-script-to-know-if-it-is-being-piped-to). In this link it mentions the usage of `os.isatty(0)`, which is used to detect if the program is being piped. My implementation is use it to detect if the program is being piped first, then trying to read from stdin.
But I was thinking that reading from stdin causes the error, from error it seems that pytest doesn't let me read from stdin in this case.
For `sys.argv[1] == "check"`, I got it. I found that we actually don't need this one, instead, it can be moved to check.py

Lee-W: Yes, I think we should move it to check.py. I'm not sure but we might need to mock `os.isatty` in that case

jordanSu: Hello @Lee-W , I've submitted a new PR #267 , which has passed the test. Please help review it.

Lee-W: Thanks! I'm reviewing it. Just curious, is there any reason why you create a new PR?

jordanSu: Yes. I force pushed to my branch few days ago, but I didn't see new commit appear in this PR, so I create a new PR. (But it seems new commits appear again in this PR now)

Lee-W: Hmmm... GitHub sometimes got weird these days. Then, I'll close this one and work on the new one

Lee-W:
Hmmm... it's werid. looks like the new commit is not updated to the new one but this one

commitizen/commands/check.py
```diff
@@ -32,9 +33,13 @@ def __init__(self, config: BaseConfig, arguments: Dict[str, str], cwd=os.getcwd(
         self.cz = factory.commiter_factory(self.config)
 
     def _valid_command_argument(self):
-        if not (
-            bool(self.commit_msg_file) ^ bool(self.commit_msg) ^ bool(self.rev_range)
-        ):
+        # The number of args that user provided
```
Lee-W, line 36: I think we don't need this comment since the naming is great 🙂

docs/check.md
```diff
@@ -1,71 +1,38 @@
...
+```bash
+$ cz check --commit-msg-file COMMIT_MSG_FILE
 ```
+In this option, COMMIT_MSG_FILE is the path of the temporal file that contains the commit message
 
-* Step 3: Install the configuration into git hook through `pre-commit`
-
+### Git Rev Range
+If you'd like to check a commit's message after it has already been created, then you can specify the range of commits to check with `--rev-range REV_RANGE`
 ```bash
```
Lee-W, line 19: I would suggest add one more blank line around ``` block. Sometime the markdown parser won't parse correctly if there's no blank line

docs/check.md
```diff
@@ -1,71 +1,38 @@
...
-  - repo: https://github.com/commitizen-tools/commitizen
-    rev: v1.17.0
-    hooks:
-      - id: commitizen
-        stages: [commit-msg]
+### Commit Message File
+```bash
+$ cz check --commit-msg-file COMMIT_MSG_FILE
 ```
+In this option, COMMIT_MSG_FILE is the path of the temporal file that contains the commit message
```
Lee-W, line 15: I'd suggest move this session to the lastest sesssion. Also, we can metion that it's useful when cooperate with git hook. Otherwise, I doubt many would check commit message in a file.

docs/check.md
```diff
@@ -1,71 +1,38 @@
...
-
-This example shows how to use the check command inside of commit-msg.
-
-At the root of the project:
+For more info on how git commit ranges work, you can check the [git documentation](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_commit_ranges).
 
+### Commit Message
+There are two ways you can provide your plain message and check it
+#### Method 1: use -m or --message
 ```bash
```
Lee-W, line 29: Add one blank line

docs/check.md
```diff
@@ -1,71 +1,38 @@
...
+In this option, MESSAGE is the commit message to be checked
 
-Open the file and edit it:
-
-```sh
-#!/bin/bash
-MSG_FILE=$1
-cz check --commit-msg-file $MSG_FILE
+#### Method 2: use pipe to pipe it to `cz check`
+```bash
```
Lee-W, line 35: Add one blank line

Lee-W (changes requested):
Overall, it looks great 🎉 We just need to fix some minor content issues. Also, I notice that you miss the period in some sentences. It'd be even better if you could take care of it

Lee-W (approved):
LGTM. Let's merge it 👍

Lee-W:
@jordanSu I just found out the reason why the branches are not synced. You have a typo haha (`cz-pipe-messsage`, `cz-pipe-message`)

jordanSu:
> @jordanSu I just found out the reason why the branches are not synced. You have a typo haha (`cz-pipe-messsage`, `cz-pipe-message`)
@Lee-W 
Ah I see. I didn't notice that lol

PR #267, opened 2020-09-15
feat(cz_check): cz check can read commit message from pipe

jordanSu:
## Description
As mentioned in #200, this PR makes `cz check` can read message from pipe. In this mode, "--message" option is used.

## Checklist

## Expected behavior
String input from pipe can be read by `cz check`

1. `echo COMMIT_MESSAGE | cz check`
2. `cat COMMIT_FILE | cz check`

## Additional context

#200

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -1,5 +1,6 @@
 import os
 import re
+import sys
 from typing import Dict, Optional
 
 from commitizen import factory, git, out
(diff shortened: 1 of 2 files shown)


commitizen/commands/check.py
```diff
@@ -32,9 +33,12 @@ def __init__(self, config: BaseConfig, arguments: Dict[str, str], cwd=os.getcwd(
         self.cz = factory.commiter_factory(self.config)
 
     def _valid_command_argument(self):
-        if not (
-            bool(self.commit_msg_file) ^ bool(self.commit_msg) ^ bool(self.rev_range)
-        ):
+        sources = (
```
Lee-W, line 36: The naming `sources` does not look straightforward to me. Currently, I do not have a better name for it. But we could at least add some comments to it.

Lee-W (commented):
Overall, it looks great 🙂 I've also tested it on my local machine. But this feature is not obvious enough for users to find it. Could you please add some example to the documentation? Thanks!

commitizen/commands/check.py
Woile: it's the amount/number of params provided, so something like `params_amount` or `number_params_provided` could be

Woile:
Nice work! Looks great, I'm not much an user of `check` so Lee can decide when to merge this. But I have to say, this PR would be super cool and I'd start using something like `echo "fix: bfoo bar" | cz check`.
Docs would be highly appreciated.

Lee-W:
As #255 is merged, we can close this one

Issue #271, opened 2020-09-22
lightweight tags create no "create event" in github actions


diefans:
## Description
Commitizen creates only lightweight tags, which are not creating a "create event" in github actions.

## Steps to reproduce
https://github.com/commitizen-tools/commitizen/blob/master/commitizen/git.py#L49

## Current behavior
A lightweight tag created by commitizen does not lead to a usefull event in github actions.

## Desired behavior
Create an annotated tag, e.g. `git tag -a <version> -m <version>`. This should be at least a configurable option IMHO.

## Environment
- commitizen version: master

Woile:
thanks for the finding! some questions, how did you find this problem? could you point the GA documentation where they mention lightweight tag don't create events? I couldn't find it.

I'll run some tests in to verify it as well :muscle:

diefans:
I have no documentation at hand. But as far as I know, lightweight tags are not "real" objects in git and have also no creation date. You can see this via `git for-each-ref` where the objecttype of a lightweight tag is `commit` and for an annotated tag it is `tag` (also find this in my unit test).
So basically I found this by trying github actions to do something after tag creation (see https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create), but there was no event triggered after commitizen created the "tag". So I just created a dummy annotated tag and my workflow was running - so I came to that conclusion....

Lee-W:
We use the following syntax to detect release which might be the reason why we did not encounter such problem.

```yaml
on:
  push:
    tags:
      - 'v*'
```

But I think it's a feature we could have. Thanks for great work 🙂 I'll also run some test to verify it.

diefans:
@Lee-W I do not really understand how this is working for you:
I have a job running basically the example in your docs:
```
  bump-version:
    if: github.ref == 'refs/heads/master'
    runs-on: ubuntu-latest
    name: "Bump version and create changelog with commitizen"
    needs:
      - build
    steps:
      - name: Check out
        uses: actions/checkout@v2
        with:
          token: '${{ secrets.GITHUB_TOKEN }}'
          fetch-depth: 0
      - name: Create bump and changelog
        uses: commitizen-tools/commitizen-action@master
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
```
The job appears to work fine: ```
...
Configuring git user and email...
Runnung cz...
bump: version 0.2.1 → 0.2.2
tag to create: 0.2.2
increment detected: PATCH

Done!
Pushing to branch...
[...]

Issue #261, opened 2020-09-03
[Question] Shouldn't releases be automated? 


alecgerona:
Just curious why releases don't happen as soon as changes are merged to master. The version is already bumped anyway.

Woile:
Yes, they should. There's a bug, the new CI is using `secrets.GITHUB_TOKEN`, but we should go back to `secrets.PERSONAL_ACCESS_TOKEN`, because the first one doesn't trigger new builds :/

Woile:
mmm not really. The vision we have is a bit more decoupled.
- a **new commit to master** triggers a bump version, which in turn will push a tag.
- a **new tag** with the format `v*` triggers a publish. This means that if I want to make a release pushing a local tag we can (for example a release candidate).

See it like "which git event triggers which action".

The update I mention is when we started using our own github action, you can see the change [here](https://github.com/commitizen-tools/commitizen/commit/7ec5c1fbbb2d7e1b2cc86778cf111d56f3b1294b)

It's fixed now, a new release will happen soon

alecgerona:
We can have release candidates or alphas or betas as branches, too. Just for context my flow draws from how [semantic-release](https://github.com/semantic-release/semantic-release) does it.

alecgerona:
Hmm I think there's no release yet.

Woile:
I've released the latest versions, but indeed it's not automatically releasing, I'll make an investigation

alecgerona:
Thanks! Let's close this once it's resolved. If you need help lemme know.

Lee-W:
Is it the action that releases 2.3.1?
https://github.com/commitizen-tools/commitizen/actions/runs/242737189

alecgerona:
Seems like it.

Lee-W:
Is there anything we need to do for this issue?

Woile:
I manually pushed the tags in order to trigger the action.
Since we introduced the github action, it doesn't trigger other github actions. We need to find out why.
I'm still not sure, because I've updated the secret to use the `PERSONAL_ACCESS_TOKEN`.

AFAIK if you use `GITHUB_TOKEN` it doesn't trigger other jobs, kind of like using `[skip ci]`, so you have to use your own token to trigger the jobs. But again, I updated the token, and it didn't make a release. Maybe some leftovers, and the next release will work.

Lee-W:
Got it. Thanks for the explanation 😄

Woile:
Releases are working fine again

PR #272, opened 2020-09-22
fix(#271): add annotated_tag option to bump

diefans:
Enable creation of annotated tags when bumping.

## Description
Create annotated tags, when bumping.

## Checklist

## Expected behavior
With `annotated_tag` enabled commitizen will create annotated tag instead of lightweight ones.

Use `cz bump --annotated_tag` or `annotated_tag = 1` in config file

## Additional context

maybe related to https://github.com/commitizen-tools/commitizen/issues/261 ???

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -131,6 +131,11 @@ data = {
                         ),
                         "action": "store_true",
                     },
+                    {
+                        "name": ["--annotated-tag", "-at"],
+                        "help": "create annotated tag instead of lightweight one",
+                        "action": "store_true",
+                    },
                 ],
             },
             {
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -31,7 +31,13 @@ class Bump:
             **config.settings,
             **{
                 key: arguments[key]
-                for key in ["tag_format", "prerelease", "increment", "bump_message"]
+                for key in [
+                    "tag_format",
+                    "prerelease",
+                    "increment",
+                    "bump_message",
+                    "annotated_tag",
+                ]
                 if arguments[key] is not None
             },
         }
@@ -157,7 +163,11 @@ class Bump:
         c = git.commit(message, args=self._get_commit_args())
         if c.return_code != 0:
             raise BumpCommitFailedError(f'git.commit error: "{c.err.strip()}"')
-        c = git.tag(new_tag_version)
+        c = git.tag(
+            new_tag_version,
+            annotated=self.bump_settings.get("annotated_tag", False)
+            or bool(self.config.settings.get("annotated_tag", False)),
+        )
         if c.return_code != 0:
             raise BumpTagFailedError(c.err)
         out.success("Done!")
--- a/docs/bump.md
+++ b/docs/bump.md
@@ -54,32 +54,30 @@ Some examples:
 
 ```bash
 $ cz bump --help
-usage: cz bump [-h] [--dry-run] [--files-only] [--changelog] [--no-verify]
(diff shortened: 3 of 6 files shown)


docs/bump.md
```diff
@@ -216,6 +214,15 @@ Some examples
 bump_message = "release $current_version → $new_version [skip-ci]"
 ```
 
+### `annotated_tag`
+
+Whether to create annotated tags or lightweight ones.
+
+```toml
+[tool.commitizen]
+annotated_tag = 1
```
Woile, line 223: Can this be a boolean instead? `true`

Woile:
Even if this were unrelated to the original issue (not triggering a workflow). I think it would be useful for traceability. An annotated tag adds:
- date
- creator
- message -> it could even contain the changes from the release which would appear in the changelog

I'll test how it could works, for `v3` could be the default.

What do you think @Lee-W ?

docs/bump.md
diefans: done

docs/bump.md
```diff
@@ -216,6 +214,15 @@ Some examples
 bump_message = "release $current_version → $new_version [skip-ci]"
 ```
 
+### `annotated_tag`
+
+Whether to create annotated tags or lightweight ones.
+
```
Lee-W, line 220: If my understand is correct, we can use `cz bump -at` and set it in config to achive it. If so, I'd suggest adding both usages here.

commitizen/cli.py
```diff
@@ -131,6 +131,11 @@
                         ),
                         "action": "store_true",
                     },
+                    {
+                        "name": ["--annotated-tag", "-at"],
+                        "help": "create annotated tag instead of lightweight one",
```
Lee-W, line 136: Maybe we can change the help message to `create an annotated tag (by default, commitizen creates lightweight tag)`

commitizen/commands/bump.py
```diff
@@ -157,7 +163,11 @@ def __call__(self):  # noqa: C901
         c = git.commit(message, args=self._get_commit_args())
         if c.return_code != 0:
             raise BumpCommitFailedError(f'git.commit error: "{c.err.strip()}"')
-        c = git.tag(new_tag_version)
+        c = git.tag(
+            new_tag_version,
+            annotated=self.bump_settings.get("annotated_tag", False)
```
Lee-W, line 168: to enhance readbility, I would suggest make `self.bump_settings.get("annotated_tag", False) or or bool(self.config.settings.get("annotated_tag", False))` a variable. I read this line for a few times to understand

Lee-W:
> Even if this were unrelated to the original issue (not triggering a workflow). I think it would be useful for traceability. An annotated tag adds:
> 
>     * date
> 
>     * creator
> 
>     * message -> it could even contain the changes from the release which would appear in the changelog
> 
> 
> I'll test how it could works, for `v3` could be the default.
> 
> What do you think @Lee-W ?

Although I love this feature, I'm a bit against making it a default option. I'm not sure how common annotated tag is used in practice.

Woile:
Doing `man git-tag` gives this:

> Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels. 

[Here's an answer from SO](https://stackoverflow.com/a/25996877/2047185)

I have the same concerns as you, not sure how used is in practice, spite of what the documentation says.
But let's see, maybe once we start using it, it does make more sense.

Lee-W:
Got it. It makes more sense after reading it. I'm now neutral to make it as a default.

Woile:
Please remove the changes on `CHANGELOG.md` and rebase. Looks good.

Woile:
This was already added

Issue #256, opened 2020-09-01
Make commitizen homebrew installable


Lee-W:
## Description
Make commitizen homebrew installable

## Possible Solution
Publish to homebrew

harens:
Hi there. I tried having a quick go at this now, but I didn't realise that commitizen uses Poetry, which doesn't play nicely with Homebrew. As a result, you might find it difficult to write a formula due to a known cyclic dependency issue found here https://github.com/python-poetry/poetry/issues/1975.

Hopefully this saves you some time for something more productive. I've left my code below in case it helps.

<details><summary>Homebrew Formula</summary>
<p>

```ruby
class Commitizen < Formula
  include Language::Python::Virtualenv

  desc "Create committing rules for projects, auto bump versions and auto changelog generation"
  homepage "https://commitizen-tools.github.io/commitizen/"
  url "https://github.com/commitizen-tools/commitizen/archive/v2.2.0.tar.gz"
  sha256 "ef983be8eafbd72220cc0802c86b907c4966701d9628431ed70ba69534bf86e8"
  license "MIT"

  depends_on "python@3.8"

  resource "colorama" do
    url "https://files.pythonhosted.org/packages/82/75/f2a4c0c94c85e2693c229142eb448840fba0f9230111faa889d1f541d12d/colorama-0.4.3.tar.gz"
    sha256 "e96da0d330793e2cb9485e9ddfd918d456036c7149416295932478192f4436a1"
  end

  resource "decli" do
    url "https://files.pythonhosted.org/packages/9f/30/064f53ca7b75c33a892dcc4230f78a1e01bee4b5b9b49c0be1a61601c9bd/decli-0.5.2.tar.gz"
    sha256 "f2cde55034a75c819c630c7655a844c612f2598c42c21299160465df6ad463ad"
  end

  resource "Jinja2" do
    url "https://files.pythonhosted.org/packages/64/a7/45e11eebf2f15bf987c3bc11d37dcc838d9dc81250e67e4c5968f6008b6c/Jinja2-2.11.2.tar.gz"
    sha256 "89aab215427ef59c34ad58735269eb58b1a5808103067f7bb9d5836c651b3bb0"
  end

  resource "MarkupSafe" do
    url "https://files.pythonhosted.org/packages/b9/2e/64db92e53b86efccfaea71321f597fa2e1b2bd3853d8ce658568f7a13094/MarkupSafe-1.1.1.tar.gz"
[...]

PR #278, opened 2020-10-06
Homebrew Update

harens:
## Description
Following the addition of commitizen to homebrew, I've updated the docs and added a Github Action.

The workflow depends on a [Personal Access Token](https://github.com/settings/tokens/new?scopes=public_repo) being made available for it.

## Checklist

## Expected behavior
Whenever a release is published, a PR should be sent to update the Homebrew formula.

N/A

## Additional context
Closes https://github.com/commitizen-tools/commitizen/issues/256

Github Action based on https://github.com/commitizen-tools/commitizen/issues/256#issuecomment-704182520 and [nef's workflow](https://github.com/bow-swift/nef/blob/master/.github/workflows/bump-formula.yml).

An example of nef's workflow being run successfully can be found [here](https://github.com/bow-swift/nef/actions/runs/181470176)

### DIFF
--- a/.github/workflows/homebrewpublish.yaml
+++ b/.github/workflows/homebrewpublish.yaml
@@ -1,0 +1,17 @@
+name: Publish to Homebrew
+
+on:
+  push:
+    tags:
+      - 'v*'
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Update Homebrew formula
+      uses: dawidd6/action-homebrew-bump-formula@v3
+      with:
+        token: ${{secrets.PERSONAL_ACCESS_TOKEN}}
+        formula: commitizen
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,6 +5,7 @@ Commits](https://img.shields.io/badge/Conventional%20Commits-1.0.0-yellow.svg?st
 release](https://img.shields.io/pypi/v/commitizen.svg?style=flat-square)](https://pypi.org/project/commitizen/)
 [![Supported
 versions](https://img.shields.io/pypi/pyversions/commitizen.svg?style=flat-square)](https://pypi.org/project/commitizen/)
+[![homebrew](https://img.shields.io/homebrew/v/commitizen?color=teal&style=flat-square)](https://formulae.brew.sh/formula/commitizen)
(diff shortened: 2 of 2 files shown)


harens:
Error seems unrelated:

```
commitizen/commands/init.py:50: error: Argument "choices" to "select" has incompatible type "List[str]"; expected "List[Union[str, Choice, Dict[str, Any]]]"
```

Lee-W:
I'm still trying to understand this issue. It seems related to https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types

Woile:
I think it's related to questionary, but not sure.
we could skip linting there by adding a comment: `# type: ignore`
When I have more time I'll check again

Woile:
I've opened a ticket on mypy https://github.com/python/mypy/issues/9555

Lee-W:
I think that is not actually a bug but rather how mypy is designed.
https://github.com/python/mypy/issues/3351


The error message of this issue is as followed.

```
commitizen/commands/init.py:50: error: Argument "choices" to "select" has incompatible type "List[str]"; expected "List[Union[str, Choice, Dict[str, Any]]]"
commitizen/commands/init.py:50: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
commitizen/commands/init.py:50: note: Consider using "Sequence" instead, which is covariant
commitizen/commands/init.py:82: error: Argument "choices" to "select" has incompatible type "Optional[List[str]]"; expected "List[Union[str, Choice, Dict[str, Any]]]"
```

In [mypy doc](https://mypy.readthedocs.io/en/latest/common_issues.html#variance), it explains how we might be able to solve this problem. However, the type annotation is located in [questionary](https://github.com/tmbo/questionary/blob/master/questionary/prompts/select.py#L18)

If questionary uses `Sequence[Union[Text, Choice, Dict[Text, Any]]]` instead, we might not encounter this problem.

As for line 82, it's probably an error I created. I dicsussed it with the author or #279 offline, we'll fix that one.

Woile:
Can u rebase the PR? Thanks

harens:
Hopefully that should do it, although let me know if it didn't. I'm happy to squash my commits if necessary, although I left them as two separate ones since one's for the docs and the other is for the workflow.

(I like the summary above ☝by the way)

Lee-W:
Left them as separate commits is great! Thanks for your great work! Let's merge it 😃

PR #284, opened 2020-10-25
fix(commands/bump): Add NoneIncrementExit to avoid git fatal error

### LINKED ISSUE #280: Commitizen tries to create a tag eventhough no increment is detected and it fails with tag already exists.
## Description
Whenever something is merged in our main branch commitizen checks if a version bump is required. Usually it's fine but whenever no increment is detected it will still try to create a tag while it shouldn't. Because the version is still the same and matches a previously created tag the command and the pipeline fails. Am I just not handling this correctly? I can't imagine nobody has run into this yet.

## Steps to reproduce
Make sure a git tag exists that matches the version commitizen is tracking. Create a new commit that shouln't trigger a version bump then run `cz bump` and watch it fail.

```
bump: version 0.10.2 → 0.10.2
tag to create: v0.10.2
increment detected: None
fatal: tag 'v0.10.2' already exists
```

## Desired behavior
I think it shouldn't try to create a git tag if no change is detected.

christian-hawk:
… creating existing tag

raises NoneIncrementExit if `increment is None and new_tag_version == current_tag_version`

closes #280 

## Checklist

## Expected behavior

Not raising git fatal error (please see #280 )

1. commit a test, ci, build or doc w/ `cz c`
2. run `cz bump`
3. **no** fatal error will occur

### DIFF
--- a/tests/commands/test_bump_command.py
+++ b/tests/commands/test_bump_command.py
@@ -1,7 +1,10 @@
+import inspect
 import sys
+from unittest.mock import MagicMock
 
 import pytest
 
+import commitizen.commands.bump as bump
 from commitizen import cli, cmd, git
 from commitizen.exceptions import (
     BumpTagFailedError,
@@ -9,6 +12,7 @@ from commitizen.exceptions import (
     DryRunExit,
     ExpectedExit,
     NoCommitsFoundError,
+    NoneIncrementExit,
     NoPatternMapError,
     NotAGitProjectError,
     NoVersionSpecifiedError,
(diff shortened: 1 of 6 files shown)


christian-hawk:
Opened issue related to github action test errors: #285

christian-hawk:
Now also closes #285

tests/commands/test_bump_command.py
```diff
@@ -262,3 +266,38 @@ def test_bump_in_non_git_project(tmpdir, config, mocker):
...
+def test_none_increment_exit_should_exist_in_bump():
+    assert hasattr(bump, "NoneIncrementExit")
+
+
+def test_none_increment_exit_is_exception():
+    assert bump.NoneIncrementExit == NoneIncrementExit
+
+
+@pytest.mark.usefixtures("tmp_commitizen_project")
+def test_none_increment_should_not_call_git_tag(mocker, tmp_commitizen_project):
```
Woile, line 288: Can u explain me what's going on in this test?
we are creating a tag, we bump and because there's no increment, it will rise an exception?
What is `assert_not_called` doing?

christian-hawk: Helo @Woile , thanks for the review.

As I usually use TDD, I write tests that break and write code to pass the test and refactor.

`assert_not_called` asserts that `git.tag` was not called, I started here: if there is no version increment, git.tag should not be called. So if by any case git.tag is called when there is no version increment, test will break.

But if git.tag does not respond properly, I cannot perform test that breaks. So I created a dummy return value here:
    dummy_value = git.tag("0.0.2")
    git.tag = MagicMock(return_value=dummy_value)

So when when I run a bump with None increment (i.e.: with a `test: ...` commit), when I run `cz bump --yes` git.tag should not be called and raise `NoneIncrementExit`)

Let me know your thoughts.

Issue #298, opened 2020-11-10
Bump: Optional Commit and Tag


marcossantiago:
## Description
It would be very great if the `bump` command accepted the Commit and Tag to be optional instead. 

## Possible Solution
Add the option in the configuration:
```
commit = true
tag = true
```

## Additional context
My workflow is usually require squashing and merging Pull Requests, making the tag in a branch not so useful. So I haveI have to re-tag the commit in the master branch.

Lee-W:
In my workflow, I usually do `cz bump` after the branch is merged into master branch.

Currently, commitizen use git tag to find which commit is the previous release. If git tag is not created, commitizen will not be able to bump version correctly

Woile:
As Lee pointed out, the idea is to run `cz bump` on your main branch. If you run it on your "work" or "feature" branch, you may have issues, like 2 people creating the same version, because both were creating features.

Anyway, you may find useful this command which doesn't create a tag nor a commit, just increments the version:

```cz bump --files-only```

marcossantiago:
Thanks for the input. The `cz bump --files-only` seems to be the right approach here, but it seems it does not update the ` pyproject.toml`. Is that correct ?

marcossantiago:
It seems to be the case...when using `files-only` the command exits before the toml config is written as per this [line](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/bump.py#L162). 

What do you think of changing the order to write the toml config before it exits [here](https://github.com/commitizen-tools/commitizen/blob/ae03624d512557804d86bcfbe8497035bb27ed74/commitizen/commands/bump.py#L148) ?

marcossantiago:
@Woile @Lee-W I have created a PR with a simple change.

PR #299, opened 2020-11-11
feat: allow files-only to set config version and create changelog

marcossantiago:
## Description
This PR allows the command `bump` with the flag `--files-only` to update the version in the configuration file and to create the changelog if the flag was passed. 

The current behavior exits the program before those 2 actions happen. 

## Checklist

- NA Add test cases to all the changes you introduce

## Expected behavior
The version in the configuration file will be updated and the changelog will be created (if flag passed) when using `bump --files-only`, changing the current behavior. 

1. run `cz bump --files-only --changelog`
2. check if the version was updated file `pyproject.toml` or `.cz.toml`. 
3. check if the changelog was actually created.

## Additional context
Related issue: #298

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -148,8 +148,6 @@ class Bump:
             version_files,
             check_consistency=self.check_consistency,
         )
-        if is_files_only:
-            raise ExpectedExit()
 
         if self.changelog:
             changelog_cmd = Changelog(
@@ -164,6 +162,10 @@ class Bump:
             c = cmd.run(f"git add {changelog_cmd.file_name}")
 
         self.config.set_key("version", str(new_version))
+
+        if is_files_only:
+            raise ExpectedExit()
+
         c = git.commit(message, args=self._get_commit_args())
         if c.return_code != 0:
             raise BumpCommitFailedError(f'git.commit error: "{c.err.strip()}"')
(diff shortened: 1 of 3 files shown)


Woile:
First, thanks for contributing, highly appreciated! I think this is a breaking change. I'll take a look tomorrow.

Lee-W (commented):
Thanks for your contribution! It looks good to me. Could you please modify [this test case](https://github.com/commitizen-tools/commitizen/blob/master/tests/commands/test_bump_command.py#L217) to test this functionality?

marcossantiago:
@Lee-W done!

Lee-W (approved):
Thanks for your contribution! It looks good to me. @Woile Could you please check again?

Woile:
Hey sorry for taking too long. I'm trying to find the original use case for `files-only`. 
I recall it was to be able to generate the changelog with a different tool before doing the actual bump, there was like a chicken and egg situation, so this feature was introduced.

I hope my lack of memory doesn't break the intention of this feature, but I think this PR can be merged.

Just so my mistake doesn't happen again, I'll ask you to add documentation about this flag: use case and examples :sweat_smile:

marcossantiago:
@Woile Added a simple explanation for the `--files-only`. The only use case it makes sense it's to avoid creating the commit and tag, which is how I use it. Let me know if you think we could add more details. 

In time, thanks for maintaining this project!

Lee-W:
Let's merge it 🎉

Issue #120, opened 2020-01-13
JSON support for configuration


Lee-W:
**Goal**
Support `.cz.json`, `cz.json` configuration

**Description**
JSON seems to be a more common configuration format for non-python project.

Woile:
That's relative, rust uses `toml`, netlify also uses `toml`, and it's becoming quite a standard for configuration (specially because it support comments).
Spite of that, because of the flexibility we now have, a json parser could be implemented.

winniepopu:
I would like to take this issue @Lee-W

Lee-W:
Sure 😃 Feel free to ask us if you encounter any problem

Lee-W:
It seems this feature has been added

PR #313, opened 2020-11-27
Add json support for config

danvergara:
## Description

As mentioned in #120 , this PR add json support for configuration

## Checklist

## Expected behavior

Replace a TOML config file with a JSON config file with no side effects, (of course, passing the same arguments).

 1. Run the tests
 2. Activate the virtual environment provided by the project/Poetry
 3. Since the json files are supported for configuration, try to add a new file called .cz.json to customize Commitizen.

## Additional context

Related to #120

### DIFF
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -149,4 +149,4 @@ class TestJsonConfig:
         json_config.init_empty_config_content()
 
         with open(path, "r") as json_file:
-            assert json.load(json_file) == {"commitizen": ""}
+            assert json.load(json_file) == {"commitizen": {}}
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -1,3 +1,4 @@
+import json
 import os
 from pathlib import Path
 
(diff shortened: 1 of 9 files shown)


tests/test_conf.py
```diff
@@ -23,6 +24,14 @@
 target-version = ['py36', 'py37', 'py38']
 """
 
+JSON_CONFIG = {
```
Lee-W, line 27: For this test case, I'd suggest using raw string instead of `dict` which might be a better simulation to our real use case.

e.g.,

```python
"""
    "commitizen": {
        "name": "cz_jira",
        "version": "1.0.0",
        "version_files": ["commitizen/__version__.py", "pyproject.toml"],
        "style": [["pointer", "reverse"], ["question", "underline"]],
    }
"""
```

Lee-W (commented):
Thanks for your contribution! I just browse through it and it looks great! I'll clone it and test it locally to confirm again and then I think we can merge it 🎉

Lee-W:
@danvergara You'll also need to handle the json format [here](https://github.com/danvergara/commitizen/blob/add-json-support-for-config/commitizen/commands/init.py#L27) so that `cz init` can run correctly

danvergara:
@Lee-W thank you, man! I miss include the json support there!

danvergara:
@Lee-W Done!

Lee-W:
@danvergara I just confirmed it. Let's merge it!

PR #316, opened 2020-12-16
fix(commands/commit): create tmp file divided by user info

SHAQ522:
For multi user siuation, it may get fail

```
Permission denied: '/tmp/cz.commit.backup'
``` 

The reason is another user has already created a temp file with the same name.
So if one user create temp_file should attatch with `$USER`  info to resolve permission issue

## Description

attach user info to temp-file name when $USER is possible

## Checklist

## Expected behavior

create `/tmp/cz.commit.backup.tom`

### DIFF
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -28,7 +28,10 @@ class Commit:
         self.config: BaseConfig = config
         self.cz = factory.commiter_factory(self.config)
         self.arguments = arguments
-        self.temp_file: str = os.path.join(tempfile.gettempdir(), "cz.commit.backup")
+        name = "cz.commit.backup"
+        if "USER" in os.environ:
+            name = name + "." + os.environ['USER']
+        self.temp_file: str = os.path.join(tempfile.gettempdir(), name)
 
     def read_backup_message(self) -> str:
         # Check the commit backup file exists


Woile:
Nice, thanks! You'll have to run `./scripts/format` and the tests will pass

commitizen/commands/commit.py
```diff
@@ -28,7 +28,10 @@ def __init__(self, config: BaseConfig, arguments: dict):
         self.config: BaseConfig = config
         self.cz = factory.commiter_factory(self.config)
         self.arguments = arguments
-        self.temp_file: str = os.path.join(tempfile.gettempdir(), "cz.commit.backup")
+        name = "cz.commit.backup"
+        if "USER" in os.environ:
+            name = name + "." + os.environ['USER']
```
Woile, line 33: I'd prefer something like this:
```
user = os.environ.get('USER', "")
name = f"cz.commit{user}.backup
```

commitizen/commands/commit.py
```diff
@@ -28,7 +28,10 @@ def __init__(self, config: BaseConfig, arguments: dict):
...
+        self.temp_file: str = os.path.join(tempfile.gettempdir(), name)
 
     def read_backup_message(self) -> str:
```
Woile, line 36: shouldn't the file be read here?

commitizen/commands/commit.py
SHAQ522: Thank you for the apply. we have 100+ users project using this package which have this issue, but I am new commiter for this python package. I will  rebuild this issue only after work.

Woile: No rush, take your time

PR #317, opened 2020-12-16
fix(commit): attach user info to backup for permission denied issue

SHAQ522:
For multi user situation, it may get fail

```
Permission denied: '/tmp/cz.commit.backup'
``` 

The reason is another user has already created a temp file with the same name.
So if one user create temp_file should attatch with `$USER`  info to resolve permission issue

## Description

attach user info to temp-file name when $USER is possible

## Checklist

## Expected behavior

create `/tmp/cz.committom.backup`


Woile:
Awesome, thanks! 🎉

PR #339, opened 2021-02-02
ci: attempt to fix homebrew action

Woile:
Reading the docs from the action it says that it may have issues if running on ubuntu. 
I thought it would be nice to run on macos and see if it works

https://github.com/dawidd6/action-homebrew-bump-formula


Woile:
mental note: if this doesn't work I'll ask to the developer of the action to help us out

Lee-W (approved):

PR #348, opened 2021-02-19
feat(changelog): add support for multiline BREAKING paragraph

### LINKED ISSUE #346: Breaking change should support multiline
## Description
Breaking change only extract the first line. is it possible to extract the whole paragraph?

Woile:
## Checklist

## Expected behavior

Support for multiline body, suppose a commit message like this:

```
fix: user issue

BREAKING CHANGE: this should be all picked up by commitizen itself, when
I do a new line like this, it will be included in the end result changelog

not this
```

Commitizen will detect a `fix` and a `BREAKING CHANGE`
both will be added to the changelog

1. Create a commit like the example above
2. Run `cz changelog`
3. Observe `cat CHANGELOG.md`

## Additional context

Closes #346

### DIFF
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -76,6 +76,8 @@ def generate_tree_from_commits(
 ) -> Iterable[Dict]:
     pat = re.compile(changelog_pattern)
     map_pat = re.compile(commit_parser, re.MULTILINE)
+    body_map_pat = re.compile(commit_parser, re.MULTILINE | re.DOTALL)
+
     # Check if the latest commit is not tagged
     latest_commit = commits[0]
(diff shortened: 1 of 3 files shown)


Woile:
I'm aware I'm introducing a side-effect which I think is cool.

Given this commit message:
```
fix: user module now work

fix: another fix

BREAKING CHANGE: You have to fix `here` and `there`

footer
```

Would output this CHANGELOG:

```
### Fix

- user module now work
- another fix

### BREAKING CHANGE

- You have to fix `here` and `there`
```

Lee-W (approved):
Just a minor suggestion. Other than that everything looks good to me 🙂

PR #359, opened 2021-03-05
Feature/changelog stdout

### LINKED ISSUE #225: [Feature Request] Add command to generate only recent changelog changes
## Description
So a normal workflow would be to automatically release to GitHub after a new version tag is created. That part works, but there is no good way to include the recent changelog additions as the body paragraph of the GitHub release.

To release with github actions, you do something like
```yaml
name: Release

on:
  push:
    tags:
      - "v*"

jobs:
  github_release:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: Release to GitHub
      uses: actions/create-release@v1
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      with:
        tag_name: ${{ github.ref }}
        release_name: Release ${{ github.ref }}
        body: |
        "my new changes here"
        draft: false
        prerelease: false
```

Currently, and please correct me if I'm wrong, there is no way to automatically fill "body" with the recent additions to the changelog

## Possible Solution
Introduce an option to `cz changelog` that would allow to to generate the most recent version so you could say something like this:
```yaml
    # so we can use it as the github release body
    - name: Store recent changelog changes
      run: |
        echo ::set-env name=CHANGELOG::$(cz changelog --dry-run --recent)

    - name: Release to GitHub
      uses: actions/create-release@v1
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      with:
        tag_name: ${{ github.ref }}
        release_name: Release ${{ github.ref }}
        body: ${{ env.CHANGELOG }}
        draft: false
        prerelease: false
```

Woile:
## Description

Include a new flag on `bump` to send the changelog to stdout.

This allows the incremented changelog data to be captured and sent to a file.

I decided to reuse the changelog with `--dry-run` when this flag is set, instead of adding a new flag on changelog.
I think this is not expensive because the changelog will only be generated for just a single release/tag.

## Checklist

## Expected behavior

A normal bump should happen + changelog to the stdout

1. Create a new commit.
2. Run `cz bump --changelog --changelog-to-stdout > body.md`
3. Observe `body.md` which should contain only the incremented changes

## Additional context

Closes  #225


Lee-W (approved):
Sorry for the late review. I was a bit busy recently. This one looks good to me 😃 Let's merge it

PR #361, opened 2021-03-20
feat: Support versions on random positions

jaysonsantos:
First thanks for this quite good project!

## Description
This is a rather confusing logic but I decided to send this PR as a way to start the conversation.
I am trying to use commitizen to bump rust projects and its default way works fine when bumping the version inside a `Cargo.toml` as usually, your own package's version is on the top. But if you have a lock file, rust also bumps the version there. This PR was done in a way to bump the version only after something is matched, in the example, I used the project name so the versions_file could have something like `Cargo.lock:version:project-name` where it only bumps version if the project name was found before it.

## Checklist

## Expected behavior
Bumping versions on `Cargo.lock` or any other file, should bump only the intended line.

Try to bump any Cargo.lock file or any file that can have your own version, randomly placed in it.

### DIFF
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -150,26 +150,25 @@ def update_version_in_files(
     """
     # TODO: separate check step and write step
     for location in files:
-        filepath, *regexes = location.split(":", maxsplit=1)
+        filepath, *regexes = location.split(":")
         regex = regexes[0] if regexes else None
-
-        # Read in the file
-        file_content = []
         current_version_found = False
-        with open(filepath, "r") as version_file:
-            for line in version_file:
-                if regex:
-                    match = re.search(regex, line)
-                    if not match:
-                        file_content.append(line)
-                        continue
 
-                # Replace the target string
-                if current_version in line:
-                    current_version_found = True
-                    file_content.append(line.replace(current_version, new_version))
-                else:
-                    file_content.append(line)
+        version_file = open(filepath, "r").read()
+        match = regex and re.search(regex, version_file, re.MULTILINE)
+        if match:
+            left = version_file[: match.end()]
+            right = version_file[match.end() :]
+            line_break = _get_line_break_position(right)
+            middle = right[:line_break]
+            current_version_found = current_version in middle
+            right = right[line_break:]
+            version_file = left + middle.replace(current_version, new_version) + right
+
+        if not regex:
+            current_version_regex = _version_to_regex(current_version)
+            current_version_found = current_version_regex.search(version_file) and True
+            version_file = current_version_regex.sub(new_version, version_file)
 
         if check_consistency and not current_version_found:
             raise CurrentVersionNotFoundError(
@@ -180,7 +179,17 @@ def update_version_in_files(
 
         # Write the file out again
         with open(filepath, "w") as file:
-            file.write("".join(file_content))
+            file.write("".join(version_file))
+
+
+def _get_line_break_position(text: str) -> int:
+    position = text.find("\n")
+    return max(position, 0)
+
+
+def _version_to_regex(version: str):
+    clean_regex = version.replace(".", r"\.").replace("+", r"\+")
+    return re.compile(f"\\b{clean_regex}\\b")
 
 
 def create_tag(version: Union[Version, str], tag_format: Optional[str] = None):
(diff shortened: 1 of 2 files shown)


Lee-W:
Hi @jaysonsantos, thanks for this PR. This one looks like a rust-specific solution to me. I'm thinking of using regex like `name = "there-i-fixed-it"\nversion = "\d\.\d\.\d\."`'. But currently, we're paring it by line. We might need to re-implement this method

jaysonsantos:
Hey there @Lee-W that was my first try and then I found out it was line based.
What if I change it to regex replace on the whole file with limit of one?

Woile:
Hey thanks for the contribution, this looks like a nice feature to have.

I think a re-implementation of the line based regex would be really cool , but we should not forget the no regex case.
Looking forward to see it 💪🏻

Lee-W:
I don't have a clear idea of how this should be implemented. Maybe we could discuss more on this thread before we implement the new version 🙂

jaysonsantos:
One of the ways I see this working is:
- Read the whole file contents into a variable
- Search with a regex so you get the start-end of the match
- Seek to the end of the match, get everything from there until the next \n or eof
- Replace only between the end of the match until \n or eof

jaysonsantos:
Hey there folks, check the latest commit I sent to see the implementation https://github.com/commitizen-tools/commitizen/pull/361/commits/4ee04dcf6496ba061067bd58cfe651b342326068

jaysonsantos:
Sorry, I forgot that bumpversion uses a bunch of python versions and I used the walrus operator

jaysonsantos:
Ok, no walrus operator now

Lee-W:
@jaysonsantos Thanks for your immediate update 🎉 The behavior is slightly changed (update all v.s. update the last one). I'm not sure whether we should consider it as a breaking change. We might need to describe it in our documentation clearly. Also, I'm thinking of displaying a warning if there're multiple regex matches.

jaysonsantos:
> The behavior is slightly changed (update all v.s. update the last one).

I didn't get I think it is the same logic, no?
The old one would loop through all lines and replace exact matches of the old version to the new one. This https://github.com/commitizen-tools/commitizen/pull/361/files#diff-ed9bead31489109a8441fbfb00d2ddec00198489b0702bab164825c2d3bd9b4fR191 will be even more strict as it will only update if the exact version is surrounded by boundaries, so v1.2.3 wouldn't be changed.
I **THINK** it does not break compatibility, if you can think of a test, can you send me and I add a case for it?

commitizen/bump.py
```diff
@@ -150,26 +150,25 @@ def update_version_in_files(
...
-                        continue
-
-                # Replace the target string
-                if current_version in line:
-                    current_version_found = True
-                    file_content.append(line.replace(current_version, new_version))
-                else:
-                    file_content.append(line)
+
+        version_file = open(filepath, "r").read()
```
Lee-W, line 157: Should we use `with open` here?

commitizen/bump.py
```diff
@@ -150,26 +150,25 @@ def update_version_in_files(
...
+            right = version_file[match.end() :]
+            line_break = _get_line_break_position(right)
+            middle = right[:line_break]
+            current_version_found = current_version in middle
+            right = right[line_break:]
+            version_file = left + middle.replace(current_version, new_version) + right
+
+        if not regex:
+            current_version_regex = _version_to_regex(current_version)
+            current_version_found = current_version_regex.search(version_file) and True
```
Lee-W, line 170: What is this `and True` for?

Lee-W:
> > The behavior is slightly changed (update all v.s. update the last one).
> 
> I didn't get I think it is the same logic, no?
> The old one would loop through all lines and replace exact matches of the old version to the new one. This https://github.com/commitizen-tools/commitizen/pull/361/files#diff-ed9bead31489109a8441fbfb00d2ddec00198489b0702bab164825c2d3bd9b4fR191 will be even more strict as it will only update if the exact version is surrounded by boundaries, so v1.2.3 wouldn't be changed.
> I **THINK** it does not break compatibility, if you can think of a test, can you send me and I add a case for it?

I'm actually ok with it. Just raise this issue as I was not yet familiar enough with the changes.

jaysonsantos:
@Lee-W you can take a look again

Lee-W (approved):
Looks good to me 😃

PR #370, opened 2021-04-09
ci(version): fix poetry project version and version_files config

Lee-W:
## Description
We didn't clearly specify which version should be updated in `pyproject.toml`. After the fix of #361, we need to clearly specify it.

## Checklist

## Expected behavior
Run `cz bump` and bump the version correctly

1. Run `cz bump`
2. Check the version in `pyproject.toml` in `poetry` and `commitizen` section

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,8 +2,9 @@
 version = "2.17.1"
 tag_format = "v$version"
 version_files = [
-  "pyproject.toml:version",
-  "commitizen/__version__.py"
+    "pyproject.toml:[tool.commitizen]\nversion",
+    "pyproject.toml:[tool.poetry]\nname = \"commitizen\"\nversion",
+    "commitizen/__version__.py"
 ]
 
 [tool.black]
@@ -29,7 +30,7 @@ exclude = '''
 
 [tool.poetry]
 name = "commitizen"
-version = "2.17.0"
+version = "2.17.1"
 description = "Python commitizen client tool"
 authors = ["Santiago Fraire <santiwilly@gmail.com>"]
 license = "MIT"


Woile:
Because we are using `ci` it won't trigger a new pypi deployment. Should we deploy manually or use `fix` instead?

Woile:
We should probably update the https://github.com/commitizen-tools/commitizen/blob/master/docs/config.md
Unfortunately I think this is a breaking change, we could fix it if we update the config before updating the version numbers I think.

https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/bump.py#L179
https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/bump.py#L211

What do you think?

Lee-W:
> Because we are using `ci` it won't trigger a new pypi deployment. Should we deploy manually or use `fix` instead?

Yes, I think it would be better that we manually push this one. Forget to mention it 🤦🏻

Lee-W:
> We should probably update the https://github.com/commitizen-tools/commitizen/blob/master/docs/config.md

I agree we should update the documentation as well.

> Unfortunately I think this is a breaking change, we could fix it if we update the config before updating the version numbers I think.
> 
> https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/bump.py#L179
> https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/bump.py#L211
> 

I'm afraid I did not understand what you mean. Could you describe it a bit more? 🙂

Woile:
I think a solution would be to check for all occurrences when running the regex, that would solve the reported issue and the issue with not modifying the poetry version.

Before we were updating all changes in a file, right? now it seems like we update only one

Lee-W:
I think what you propose makes sense 👍  I'm thinking of first manually push 2.7.1, then merge this one. After all these step, we could create an issue to fix it. What do you think of the steps?

Woile:
I would only update the version in the `tool.poetry`. If we fix the `version_files`, the one we currently have configured should work, right? 
I will publish manually to pypi

Lee-W:
Sounds good 👍 Then, I'll close this PR

PR #372, opened 2021-04-09
fix(wip): add test for current breaking change

Woile:
## Description
This is a WIP to fix the current BREAKING CHANGE which includes the test that should pass. 
Because we don't want to bump, we should provide a fix of the breaking change.
In the meanwhile you can use commitizen<2.17.0

## Checklist

### DIFF
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -157,15 +157,17 @@ def update_version_in_files(
         with open(filepath, "r") as f:
             version_file = f.read()
 
-        match = regex and re.search(regex, version_file, re.MULTILINE)
-        if match:
-            left = version_file[: match.end()]
-            right = version_file[match.end() :]
-            line_break = _get_line_break_position(right)
-            middle = right[:line_break]
-            current_version_found = current_version in middle
-            right = right[line_break:]
-            version_file = left + middle.replace(current_version, new_version) + right
+        if regex:
+            for match in re.finditer(regex, version_file, re.MULTILINE):
+                left = version_file[: match.end()]
+                right = version_file[match.end() :]
+                line_break = _get_line_break_position(right)
+                middle = right[:line_break]
+                current_version_found = current_version in middle
+                right = right[line_break:]
+                version_file = (
+                    left + middle.replace(current_version, new_version) + right
+                )
 
         if not regex:
             current_version_regex = _version_to_regex(current_version)
--- a/tests/test_bump_update_version_in_files.py
+++ b/tests/test_bump_update_version_in_files.py
@@ -126,6 +126,18 @@ def test_random_location(random_location_version_file):
         assert len(re.findall(new_version, data)) == 1
 
 
+def test_duplicates_are_change_with_no_regex(random_location_version_file):
+    old_version = "1.2.3"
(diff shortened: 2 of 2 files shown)


Woile:
Oh nice @Lee-W ! 💪🏻

commitizen/bump.py
```diff
@@ -157,15 +157,17 @@ def update_version_in_files(
...
-        if match:
-            left = version_file[: match.end()]
-            right = version_file[match.end() :]
-            line_break = _get_line_break_position(right)
-            middle = right[:line_break]
-            current_version_found = current_version in middle
-            right = right[line_break:]
-            version_file = left + middle.replace(current_version, new_version) + right
+        if regex:
+            for match in re.finditer(regex, version_file, re.MULTILINE):
```
jaysonsantos, line 161: Hey there, sorry to hear that my change broke the tool.
I think this approach could have a problem if the version would have a decimal change like from 9 to 10 as IIRC finditer still holds a reference to an old string

jaysonsantos: To reproduce the error you could use this:
```patch
diff --git a/tests/test_bump_update_version_in_files.py b/tests/test_bump_update_version_in_files.py
index aebd879..3d289e2 100644
--- a/tests/test_bump_update_version_in_files.py
+++ b/tests/test_bump_update_version_in_files.py
@@ -39,15 +39,39 @@ REPEATED_VERSION_NUMBER = """
 CARGO_LOCK = """
 [[package]]
 name = "textwrap"
-version = "1.2.3"
+version = "1.2.9"
 
 [[package]]
 name = "there-i-fixed-it"
-version = "1.2.3"
+version = "1.2.9"
 
 [[package]]
 name = "other-project"
-version = "1.2.3"
+version = "1.2.9"
+
+[[package]]
+name = "other-project"
+version = "1.2.9"
+
+[[package]]
+name = "other-project"
+version = "1.2.9"
+
+[[package]]
+name = "other-project"
+version = "1.2.9"
+
+[[package]]
+name = "other-project"
+version = "1.2.9"
+
+[[package]]
+name = "other-project"
+version = "1.2.9"
+
+[[package]]
+name = "other-project"
+version = "1.2.9"
 """
 
 
@@ -115,8 +139,8 @@ def test_partial_update_of_file(version_repeated_file):
 
 
 def test_random_location(random_location_version_file):
-    old_version = "1.2.3"
-    new_version = "2.0.0"
+    old_version = "1.2.9"
+    new_version = "1.2.10"
     location = f"{random_location_version_file}:there-i-fixed-it.+\nversion"
 
     bump.update_version_in_files(old_version, new_version, [location])
@@ -127,8 +151,8 @@ def test_random_location(random_location_version_file):
 
 
 def test_duplicates_are_change_with_no_regex(random_location_version_file):
-    old_version = "1.2.3"
-    new_version = "2.0.0"
+    old_version = "1.2.9"
+    new_version = "1.2.10"
     location = f"{random_location_version_file}:version"
 
     bump.update_version_in_files(old_version, new_version, [location])

```

i think that a way (hacky though) to fix this would keep and offset variable that gets the string difference from the old version to the new (it can be positive and negative) so you take accountable this kind of problem

PR #374, opened 2021-04-17
fix: fix multiple versions bumps when version changes the string size

jaysonsantos:
## Description

As I've commented [here](https://github.com/commitizen-tools/commitizen/pull/372#discussion_r615275718) the bug fix that fixes my bug also introduces a new one.
If you have a file that has many `version = 1.0.9` and you bump to `1.1.10` every time you bump the version, the file increases 1 byte, and `re.findall` keeps a reference to the old string making every iteration point to a character back like (version, versio, versi, vers, and etc) to the point where it would move to previous likes and the check that makes sure that the current version is found on the current line fails.

## Checklist

## Expected behavior

All versions should be bumped no matter the size of the file and occurrences of a regex

1 - Have a file with at least 8 `version = 1.0.9` lines
2 - Bump version to `1.1.10` using `file:version` as the regex

## Additional context

#372 
#361

### DIFF
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -152,24 +152,15 @@ def update_version_in_files(
     for location in files:
         filepath, *regexes = location.split(":")
         regex = regexes[0] if regexes else None
-        current_version_found = False
 
         with open(filepath, "r") as f:
             version_file = f.read()
 
         if regex:
-            for match in re.finditer(regex, version_file, re.MULTILINE):
-                left = version_file[: match.end()]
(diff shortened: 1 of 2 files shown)


Woile:
Looks good to me, feel free to merge @Lee-W

Lee-W (approved):

Lee-W:
This one is interesting. 🤔 Didn't notice this might happen. Thanks, @jaysonsantos 👍👍👍

Issue #173, opened 2020-04-28
[feature] add CalVer support


Lee-W:

Woile:
This is a tough one, I've read the document https://calver.org/ and it's still not clear how to do it.

It also point out:
> it's important to recognize the practicality of each and design the scheme to fit the project

which makes it hard to extract a convention, if it's gonna be tailored to each team.

But it would be a fun experiment to see if commitizen can be flexible enough to support different versioning schemas.

Lee-W:
This is actually one of the feature requests from Remote Python Pizza attendee 😆 This could be the next one we can play with after changelog.

gadse:
I think that was my feature request :wave: . I was just checking whether I actually created it here or not, and it seems @Lee-W was faster than me.

I hope I can get to know the codebase fast enough to help implement the feature. :blush:

KyleKing:
I was thinking about this and there might be a straightforward way to handle all of the edge cases for how someone might want to format CalVer using only `strftime`:

https://github.com/commitizen-tools/commitizen/blob/ba94e3f90e2df2a80c9946968e1ff15696584b79/commitizen/bump.py#L235-L238

Could be:

```py
t = Template(tag_format)
t_ver = t.safe_substitute(
    version=version, major=major, minor=minor, patch=patch, prerelease=prerelease
)
return datetime.utcnow().strftime(t_ver)
```

As an example: `tag_format='%y.%M.%d.$major.$minor-$prerelease'` > `21.26.24.1.2a0` (Note: would likely be used in tandem with custom bump rules so that all changes would increment `minor` rather than `patch`)

---

Other schemes from projects on https://calver.org/

| Project | `tag_format` | Notes |
| -- | -- | -- |
| [pip](https://pypi.org/project/pip/) | `%y.%major.%minor` | `21.1.2` |
| [PyCharm](https://www.jetbrains.com/pycharm/download/other.html) | `%Y.%m.%d` | `2020.3.5` (but they also have a regular SemVer build, I think) |
| [certifi](https://pypi.org/project/certifi/) | `%Y.%M.%D` | `2020.12.5` |

If that sounds OK, I can submit a PR? Or feel free to handle it because it would be such a small change

Woile:
Sounds interesting! More than welcome to submit a PR! Thanks!

KyleKing:
[...]

Issue #302, opened 2020-11-12
Support bumping local version


marcossantiago:
## Description

I am working on some projects where the version is a combination of the a public api version + our local version. 

Example: `5.3.1+1.0.0`. 

So when we bump we actually wants to bump only the local version like `5.3.1+1.0.0` -> `5.3.1+1.1.0`, but `cz` completely ignores the local version and only bumps the first release like: 

```
➜ cz bump --files-only --yes
bump: version 5.3.1+0.9.0 → 5.4.0
```

This does seems like would be a breaking change though...

## Possible Solution
A possible solution would be to modify the [semver_generator()](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/bump.py#L83) to check if the local version complies with semver and bump the local version instead. The method would return the entire version like `5.3.1+1.1.0`. 

```
>>> from packaging.version import Version
>>> a = Version("5.3.1+1.0.0")
>>> a.base_version
'5.3.1'
>>> a.local
'1.0.0'
>>> b.base_version
'1.0.0'
>>> b.release
(1, 0, 0)
```

If you accept, I will be happy to create a PR with the changes. 

## Additional context

References: [Python 440 Local version](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers)

Woile:
Seems related to #150 
It doesn't necessarily have to be a breaking change, what if a flag is used `--local-version`? would it be breaking?
If you want you can share a PR and we can discuss it there.

marcossantiago:
That would be great. I will create a PR passing the flag `--local-version`, once you're okay I will also commit the tests.

marcossantiago:
@Woile PR created :)

Woile:
Released on `2.7.0` :tada:

Issue #150, opened 2020-04-07
prerelease can't comply with the semver.org specification


noelmace:
Commitizen only implements the normalized [PEP-440](https://www.python.org/dev/peps/pep-0440/#pre-release-separators) prerelease format, excluding separators between :
- the release segment and the pre-release segment
- the pre-release signifier and the numeral

Therefore, when using the `bump` command, we can't define a version format which complies with the one defined by the [semver.org specification](https://semver.org/#spec-item-9) :

> "pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version."
>
> "Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92."

**consequences**

`cz bump` can't be used for prereleases alongside tools that follow this specification.

Especially, this leads to an error when using the Node.js [semver package](https://www.npmjs.com/package/semver).

Yet, this package is used by a bunch of other tools, like [Github Automatic Release](https://github.com/marvinpinto/actions/tree/master/packages/automatic-releases) (see error message below) and [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog).

```
The parameter "automatic_release_tag" was not set and the current tag "v1.0.0rc1" does not appear to conform to semantic versioning.
(node:3557) UnhandledPromiseRejectionWarning: Error: The parameter "automatic_release_tag" was not set and the current tag "v1.0.0rc1" does not appear to conform to semantic versioning.
  at /home/runner/work/_actions/marvinpinto/action-automatic-releases/latest/dist/index.js:1:224465
  at Generator.next (<anonymous>)
  at /home/runner/work/_actions/marvinpinto/action-automatic-releases/latest/dist/index.js:1:221749
  at new Promise (<anonymous>)
  at n (/home/runner/work/_actions/marvinpinto/action-automatic-releases/latest/dist/index.js:1:221494)
  at /home/runner/work/_actions/marvinpinto/action-automatic-releases/latest/dist/index.js:1:224406
  at /home/runner/work/_actions/marvinpinto/action-automatic-releases/latest/dist/index.js:1:224971
  at Generator.next (<anonymous>)
  at /home/runner/work/_actions/marvinpinto/action-automatic-releases/latest/dist/index.js:1:221749
  at new Promise (<anonymous>)
```

**details**

[...]

PR #385, opened 2021-05-25
feat(#173): support CalVer tag formatting

KyleKing:
Closes: #173 

## Description

Adds flexible support for CalVer using `strftime`

## Checklist

## Expected behavior / Steps to Test This Pull Request

Include `strftime` code in `tag_format` (i.e. `cz bump --tag-format="%Y.%M.$version"`)

---

FYI: there are type tests failing unrelated to this PR. All of `pytest`, formatting, and other tests pass locally except for `mypy`

```
commitizen/config/toml_config.py:52: error: Argument 1 to "parse" has incompatible type "Union[bytes, str]"; expected "str"
commitizen/config/toml_config.py:54: error: Value of type "Union[Item, Container]" is not indexable
commitizen/config/toml_config.py:54: error: Argument 1 to "update" of "dict" has incompatible type "Union[Any, Item, Container]"; expected "Mapping[str, Any]"
Found 3 errors in 1 file (checked 59 source files)
```

### DIFF
--- a/docs/bump.md
+++ b/docs/bump.md
@@ -183,26 +183,28 @@ cz bump --changelog --changelog-to-stdout > body.md
 
 ### `tag_format`
 
-It is used to read the format from the git tags, and also to generate the tags.
+The `tag_format` is used to parse the format from the git tags and to generate new tags.
 
-Commitizen supports 2 types of formats, a simple and a more complex.
+Commitizen defaults to the standard Semver format (`$version`) or can be customized using a predefined set of variables.
 
 ```bash
 cz bump --tag-format="v$version"
 ```
 
 ```bash
-cz bump --tag-format="v$minor.$major.$patch$prerelease"
+cz bump --tag-format="v$major.$minor.$patch$prerelease"
 ```
 
 In your `pyproject.toml` or `.cz.toml`
 
 ```toml
 [tool.commitizen]
-tag_format = "v$minor.$major.$patch$prerelease"
+tag_format = "v$major.$minor.$patch$prerelease"
 ```
 
-The variables must be preceded by a `$` sign.
+#### Standard SemVer Variables
+
+The SemVer variables must be preceded by a `$` sign.
 
 Supported variables:
 
@@ -213,6 +215,34 @@ Supported variables:
 | `$minor`      | MINOR increment                            |
 | `$patch`      | PATCH increment                            |
 | `$prerelease` | Prerelase (alpha, beta, release candidate) |
+
+#### CalVer Variables
+
+Commitizen also supports [CalVer](https://calver.org/) in any combination with or without SemVer; however, using standard CalVer formats is recommended for interoperability with other tools. Common CalVer `tag_format`'s include `%Y.%m` or `%y.%m.%d`.
+
+If a unique build number is desired, this could either be implemented by incrementing the string in the configuration file with whatever build tool is used (i.e. `%Y.%m.123` to `%Y.%m.124`) or by mapping all commit types to one SemVer component through the `bump_map` configuration and a `tag_format` of `"%Y.%m.$major`.
+
+```bash
+cz bump --tag-format="%y.%-m.%-d$prerelease"
+cz bump --tag-format="%Y.%m.$major.$prerelease"
+cz bump --tag-format="%Y.%m.$version"
+```
+
+```toml
+[tool.commitizen]
+tag_format = "%Y.%m.%d$prerelease"
+```
+
+The full `strftime` formatting options are supported. Below are the most common for CalVer. For a full list, see: [https://strftime.org/](https://strftime.org/)
+
+| Code  | Description                                                | Example       |
+| ----- | ---------------------------------------------------------- | ------------- |
+| `%y`  | Year without century as a zero-padded decimal number.      | 08 (for 2008) |
+| `%Y`  | Year with century as a decimal number.                     | 2008          |
+| `$m`  | Month as a zero-padded decimal number.                     | 01 (for Jan)  |
+| `$-m` | Month as a decimal number. (Platform specific)             | 1 (for Jan)   |
+| `$d`  | Day of the month as a zero-padded decimal number.          | 05            |
+| `$-d` | Day of the month as a decimal number. (Platform specific)  | 5             |
 
 ---
 
--- a/tests/test_bump_create_tag.py
+++ b/tests/test_bump_create_tag.py
@@ -1,4 +1,5 @@
 import pytest
+from freezegun import freeze_time
 from packaging.version import Version
 
 from commitizen import bump
@@ -13,9 +14,16 @@ conversion = [
     (("1.2.3+1.0.0", "v$version"), "v1.2.3+1.0.0"),
     (("1.2.3+1.0.0", "v$version-local"), "v1.2.3+1.0.0-local"),
     (("1.2.3+1.0.0", "ver$major.$minor.$patch"), "ver1.2.3"),
+    (("1.2.3a1", "%y.%m.%d$prerelease"), "21.05.06a1"),
+    (("1.2.3a1", "%Y.%-m.%-d$prerelease"), "2021.5.6a1"),
+    (("1.2.3a1", "%y.%-m.$major.$minor"), "21.5.1.2"),
+    (("2021.01.31a1", "%Y.%-m.%-d$prerelease"), "2021.5.6a1"),
+    (("21.04.02", "%y.%m.%d"), "21.05.06"),
+    (("21.4.2.1.2", "%y.%-m.%-d.$major.$minor"), "21.5.6.1.2"),
 ]
 
 
+@freeze_time("2021-5-6")
 @pytest.mark.parametrize("test_input,expected", conversion)
 def test_create_tag(test_input, expected):
     version, format = test_input
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -1,8 +1,9 @@
 import re
 from collections import OrderedDict
+from datetime import datetime
 from itertools import zip_longest
 from string import Template
-from typing import List, Optional, Union
+from typing import Iterable, List, Optional, Tuple, Union
 
 from packaging.version import Version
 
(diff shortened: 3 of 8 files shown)


KyleKing:
Just following up. I think these changes should be ready to merge. The failing issues in CI are unrelated to the changes

KyleKing:
Just following up. Is there anything I can do to help merge this change?

Woile:
Hey Kyle, sorry for the late reply, I've been quite busy. Thanks for the works, looks good.

FMPOV this PR still needs some tests. Specially under the changelog area, I don't want to find myself fixing bugs on it.
I'd like to see some changelog tests where the users transition from semver to calver, and also if they start with calver from the beginning.
The same for bump.

The pipelines should also pass in order to be merged, otherwise the burden falls on us. Most of the latests PRs have passed so doing a rebase should fix the mypy errors for you.

I'd also want a new check on the `cz init` format: `commitizen/commands/init.py:96`. 
I'd use a regex to find the year or month and tell the user something like:
> seems you are using calver, for more information read: 
> link to the docs

@Lee-W any thoughts on this?

docs/bump.md
```diff
@@ -214,6 +214,31 @@ Supported variables:
 | `$patch`      | PATCH increment                            |
 | `$prerelease` | Prerelase (alpha, beta, release candidate) |
 
+Commitizen also supports [CalVer](https://calver.org/) in any combination with or without SemVer.
```
Woile, line 217: I'd add a title before this line:

```md
#### CalVer support
```

tests/test_bump_create_tag.py
```diff
@@ -13,9 +14,13 @@
     (("1.2.3+1.0.0", "v$version"), "v1.2.3+1.0.0"),
     (("1.2.3+1.0.0", "v$version-local"), "v1.2.3+1.0.0-local"),
     (("1.2.3+1.0.0", "ver$major.$minor.$patch"), "ver1.2.3"),
+    (("1.2.3a1", "%y.%m.%d$prerelease"), "21.05.06a1"),
+    (("1.2.3a1", "%Y.%-m.%-d$prerelease"), "2021.5.6a1"),
+    (("1.2.3a1", "%y.%-m.$major.$minor"), "21.5.1.2"),
```
Woile, line 19: What happens when going from calver to calver?

KyleKing: This currently fails because create_tag expects version.release to be three integers (in Version, splits on '.' and maps to int) so a CalVer of `21.5.1.2` would fail here:

https://github.com/commitizen-tools/commitizen/blob/bec20ebf433f2281c70f1eb4b0b6a1d0ed83e9b2/commitizen/bump.py#L226

This will definitely need to be fixed. I'll check how custom formats (dashes, etc.) are handled by commitizen and give this more thought

KyleKing:
Okay great thanks for the feedback! I converted this PR to a draft while I expand the test cases (After rebase, the tests are passing locally 👍🏻 )

- [ ] Make sure CI passes (pending `git push -f`)
- [ ] Resolve [conversation above for bump from calver to calver](https://github.com/commitizen-tools/commitizen/pull/385/files/b0c9fdf63456d66bbcf024d24e9f488c0f9e91d3#diff-f5fb899fe539467adba314429001b3cadf0def6e9804d288b87c89a57319a0be)
- [ ] Add tests for:
    - [ ] `changelog` for to/from calver - checks that version sequence and tags are correct
    - [ ] `bump` - to/from calver
- [ ] In `init` (see [commitizen/commands/init.py:96](https://github.com/commitizen-tools/commitizen/blob/bec20ebf433f2281c70f1eb4b0b6a1d0ed83e9b2/commitizen/commands/init.py#L96)), add if statement to check if the repository appears to be using calver

    ```py
    from datetime import datetime
    year = datetime.today().year
    # Check current year and prior two years
    # For each year, check if 4 digit year is anywhere in version (i.e. 2021.* or *.2021)
    #     Then check if starts with last two digits of current year (^21.*)
    ```

Woile:
Thanks, the proposal looks super good 👍🏻 

Here a regex I think we could use for calver for the `init` command:

```python
^(\d{4}|\d{2})\.(\d{2}|\d{1})
```

I think this would cover this cases:

```
# valid
21.05.06a1
2021.5.6a1
21.5.1.2

# invalid
234.34
1.2.3a1
21-05.06a1
```

I don't know if there's a regex for calver, I'm just assuming the following:
- calver would start with the year
- calver would include year (4 or 2) and month
- calver is separated by dot

I think that would be enough for the init, covering most of the cases.

Lee-W:
> Hey Kyle, sorry for the late reply, I've been quite busy. Thanks for the works, looks good.
> 
> FMPOV this PR still needs some tests. Specially under the changelog area, I don't want to find myself fixing bugs on it.
> I'd like to see some changelog tests where the users transition from semver to calver, and also if they start with calver from the beginning.
> The same for bump.
> 
> The pipelines should also pass in order to be merged, otherwise the burden falls on us. Most of the latests PRs have passed so doing a rebase should fix the mypy errors for you.
> 
> I'd also want a new check on the `cz init` format: `commitizen/commands/init.py:96`.
> I'd use a regex to find the year or month and tell the user something like:
> 
> > seems you are using calver, for more information read:
> > link to the docs
> 
> @Lee-W any thoughts on this?

IMO, we'll need to redesign the question follow. Depending on different bump method used, we should as different quetsions

KyleKing:
I haven't forgotten about this and have some time today to implement the changes! I was interviewing, studying for finals in my summer course, and on vacations, so all good things. Apologies for my any excess notifications as I remember how best to update a fork from upstream 😅

KyleKing:
The main problem is that CalVer is so poorly defined (i.e. `YYYY.MM.DD`, `YY.M`, `YY.MM`, `YY.MM.Build`, `YY.Major.Minor`, etc.) which makes adding CalVer a big scope creep for commitizen

There are a couple of ways to proceed:

- Right now, users could use [bumpver](https://pypi.org/project/bumpver) to increment the version following CalVer rules
    - commitizen could still be used for pre-commit checks on commit messages and to generate a Changelog using the [git tags created by BumpVer](https://github.com/mbarkhau/bumpver#vcs-parameters-gitmercurial))
- A naive approach in commitizen might be to limit users to *either* CalVer or SemVer so that CalVer users do not receive any of the bump versioning logic and the bump/SemVer logic can continue to assume between 1-3 dot-separated integers
    - I think this might be reasonable but this tradeoff adds scope creep that might distract from the core functionality of automatic SemVer that commitizen provides
- A refactoring approach might be to always use an *instance of a to-be-built Version class* (rather than a string)
    - The version instance would provide all the methods for bumping and later string formatting to encapsulate the complexity of how `tag_formats` could be configured. This approach would require a parser for the `tag_format` to retrieve the metadata from the string version (i.e. converts `v21.01.32rc1` with `v%y.%m.$major.$prerelease` to each respective component)
    - The Version instance could use any 3rd party library under the hood (such as continuing to use [`packaging.version.Version`](https://github.com/pypa/packaging/blob/ba07d8287b4554754ac7178d177033ea3f75d489/packaging/version.py)), but would likely always need to provide some additional logic to understand what values are dates (i.e. auto-generated) vs. SemVer values of interest
    - In any implementation, one problem that might arise is going between tag_formats (i.e. `$version` to `%Y.$version`)
    - While this is likely the nicest way to implement CalVer and SemVer in peaceful coexistence (and better support more complex versioning schemes, such as in #150 / #302), this would require a wide-ranging refactor
- One further workaround might be to create a plugin that can handle implementing CalVer around commitizen, but I haven't given this much thought yet

Lee-W:
both point 2 and point 3 sound good to me. are they mutually exclusive? or should they be done together?

btw I'll host a development sprint in PyCon Taiwan on 9/26 (Sun.) 09:00 (GMT+8) ~
if you could split this one to subtasks, maybe I could ask whether others are interesting in implementing this feature

Woile:
Thanks for all this deep investigation into calver @KyleKing

I've been giving a thought to a class version (maybe a protocol?), but I don't have much time now to go deep into it. Is specially interesting to me the use of a proper semver library for `rc`, `alpha`, etc.

If I understand you correctly, calVer with commitizen would only be useful to generate a changelog based on conventional commits, because libraries like bumpver already adress the bumping side, right?

KyleKing:
Thanks! 

@Lee-W, you're right that 2 & 3 aren't mutually exclusive. 2 could be implemented as an initial check for either `$` or `%` and not both characters, while 3 would be the internal refactoring for how the the Version instance is handled. I was thinking about some sort of `CzVersion` class like this example below:

```py
BUMP_MECHANISM: Callable[[CzVersion], None]

class CzVersion:

    def __init__(self, version: str, tag_format: str):
        self._version = parse_version(version, tag_format)

    @property
    def semver(self) -> Tuple[int, int, int]:
        return (self._version.major, self._version.minor, self._version.patch)

    @semver.setter
    def semver(self, major: int, minor: int, patch: int) -> None:
        self._version.major = major
        self._version.minor = minor
        self._version.patch = patch

    @semver.setter
    def bump_calver(self) -> None:
        self._version.bump_calver()

def bump_major(cz_version: CzVersion) -> None:
    cz_version.bump_calver()
    cz_version.semver = (cz_version.major + 1, 0, 0)

def bump_minor(cz_version: CzVersion) -> None:
    cz_version.bump_calver()
    cz_version.semver = (cz_version.major, cz_version.minor = 1, 0)

# Example
cz_version = CzVersion('2021.05.1.2.3', '%Y.%m.$major.$minor.$patch')
print(cz_version.semver)  # (1, 2, 3)
cz_version.semver = bump_major()
print(cz_version.semver)  # (2, 0, 0)
```

---

@Woile, I made a demo branch using bumpver to increment CalVer (`poetry run bumpver update`) and cz to update the changelog (`poetry run cz changelog`). Diff: https://github.com/KyleKing/commitizen/pull/5

Note: I tried `poetry run bumpver update --patch` and the SemVer component reset to `0.0.0` and the build incremented a lot, but that is likely an issue with the commit/tag history on commitizen (i.e. would have worked on a fresh project)

Maybe bumpver could be used by commitizen internally for parsing versions and bumping? There is a lot of good code in the package and the `update` CLI command demonstrates how that would be implemented: https://github.com/mbarkhau/bumpver/blob/master/src/bumpver/cli.py#L571-L602

KyleKing:
I have come back to this PR a few times, but I think I've settled on the idea that CalVer and SemVer need to be separate and while mixing the two could be possible, in practice tools will not work as expected and I'm not sure there are really any good reasons to do so

CalVer is perfect for packages like `certifi` which is bundling the latest version of certs and commit history is for the most part irrelevant. While packages like my `calcipy` project require much more clarity for breaking changes, which is obscured by including date information

Additionally, poetry, npm, and other managers don't work well with date-based versioning schemes where the date information is first (i.e. `^*.1.*` for `2022.1.3.1`). Not to mention there are additional issues when dashes or underscores are used or when more than three numbers are used

Overall, I think commitizen should keep focused on SemVer and this PR and issue can be closed. I'm personally staying with or returning to SemVer on all of my projects (which while they were on CalVer, they luckily have little adoption. Yay obscurity!)

Thanks for your time reviewing this PR!

PR #386, opened 2021-05-25
Fix bumping only on the first few matches

Lee-W:
## Description
Fix bumping only on the first few matches

## Checklist

## Expected behavior
Bump all version specified in version_files

1. git clone //github.com/pedrorrivero/qrand
2. git checkout 0.3.0a0
3. cz bump

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ isort = "^5.7.0"
 freezegun = "^0.3.15"
 pydocstyle = "^5.0.2"
 pre-commit = "^2.6.0"
+pytest-regressions = "^2.2.0"
 
 [tool.poetry.scripts]
 cz = "commitizen.cli:main"
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -188,13 +188,13 @@ def _bump_with_regex(version_file_contents, current_version, new_version, regex)
         right = version_file_contents[match.end() + offset :]
         line_break = _get_line_break_position(right)
         middle = right[:line_break]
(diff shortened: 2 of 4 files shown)


Lee-W:
@Woile I'm thinking of refactoring test through pytest-regressions like [this commit](https://github.com/commitizen-tools/commitizen/pull/386/commits/3f76cf109dd0031b2fb02da5efe80ad76f9020fe). What do you think? Basically, it checks whether our result in this round is different from the previous round. By doing so, we can check whether the content is exactly the same instead of check only a few lines.

pyproject.toml
```diff
@@ -70,6 +70,7 @@ isort = "^5.7.0"
 freezegun = "^0.3.15"
 pydocstyle = "^5.0.2"
 pre-commit = "^2.6.0"
+pytest-regressions = "^2.2.0"
```
Woile, line 73: Interesting!

Woile (approved):

Lee-W:
@Woile I accidentally find another bug and fixed it. Could you please check again? In `tests/test_bump_update_version_in_files/test_multiplt_versions_to_bump_without_eol_.toml` on https://github.com/commitizen-tools/commitizen/pull/386/commits/943d87360b267b3d57a57300fa97167903fb888f, our original bumping mechanism will not bump the last line. This has been fixed on https://github.com/commitizen-tools/commitizen/pull/386/commits/e0af6a2a9d2b6797343068bfba35aa18a7bc3525

Woile:
Niceee, LGTM!

PR #388, opened 2021-05-27
fix(changelog): annotated tags not generating proper changelog

### LINKED ISSUE #378: changelog generator got messed up with annotated tags
## Description
 Tags are mssing in generated tags

## Steps to reproduce
```bash
git clone https://github.com/Trim21/transmission-rpc.git

cd transmission-rpc

pip install -U commitizen

cz changelog

git tag -l

```

## Current behavior

annotated tags like `v3.2.2` and `v3.2.1` are  missing from changelog

## Desired behavior

generate a changelog with all kind of tags

## Environment

- commitizen version: 2.17.4
- python version: 3.8.9
- operating system: Windows

Woile:
Closes #378

## Checklist

## Expected behavior
Both annotated and not annotated tags should work.

```
git clone https://github.com/Trim21/transmission-rpc.git

cd transmission-rpc

pip install -U commitizen

cz changelog

git tag -l
```

## Additional context
Now both annotated and not annotated are detected. If there's no `object` from an annotated, it means we have a lightweight tag.

Please also @Trim21 let me know if this patch works for you

### DIFF
--- a/tests/commands/test_changelog_command.py
+++ b/tests/commands/test_changelog_command.py
@@ -453,3 +453,30 @@ def test_changelog_config_start_rev_option(mocker, capsys, config_path):
 
     out, _ = capsys.readouterr()
     assert out == "## Unreleased\n\n### Feat\n\n- after 0.2\n- after 0.2.0\n\n"
+
+
+@pytest.mark.usefixtures("tmp_commitizen_project")
+def test_changelog_incremental_keep_a_changelog_sample_with_annotated_tag(
+    mocker, capsys, changelog_path, file_regression
+):
+    """Fix #378"""
+    with open(changelog_path, "w") as f:
+        f.write(KEEP_A_CHANGELOG)
+    create_file_and_commit("irrelevant commit")
+    git.tag("1.0.0", annotated=True)
+
+    create_file_and_commit("feat: add new output")
+    create_file_and_commit("fix: output glitch")
+    create_file_and_commit("fix: mama gotta work")
+    create_file_and_commit("feat: add more stuff")
+    create_file_and_commit("Merge into master")
+
+    testargs = ["cz", "changelog", "--incremental"]
+
+    mocker.patch.object(sys, "argv", testargs)
+    cli.main()
+
+    with open(changelog_path, "r") as f:
+        out = f.read()
+
+    file_regression.check(out, extension=".md")
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -44,6 +44,15 @@ class GitTag(GitObject):
     def __repr__(self):
         return f"GitTag('{self.name}', '{self.rev}', '{self.date}')"
 
+    @classmethod
+    def from_line(cls, line: str, inner_delimiter: str) -> "GitTag":
+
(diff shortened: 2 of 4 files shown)


Trim21:
thanks I'll test it later. 

It should works fine if you have tested it on that repo.

tests/commands/test_changelog_command.py
```diff
@@ -229,6 +229,7 @@ def test_changelog_incremental_keep_a_changelog_sample(mocker, capsys, changelog
     testargs = ["cz", "changelog", "--incremental"]
 
     mocker.patch.object(sys, "argv", testargs)
+    # print(git.get_tags())
```
Lee-W, line 232: Should we remove this line?

tests/commands/test_changelog_command.py
```diff
@@ -453,3 +454,34 @@ def test_changelog_config_start_rev_option(mocker, capsys, config_path):
...
+    cli.main()
+
+    with open(changelog_path, "r") as f:
+        out = f.read()
+
+    assert (
+        out
+        == """# Changelog\nAll notable changes to this project will be documented in this file.\n\nThe format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),\nand this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).\n\n## Unreleased\n\n### Feat\n\n- add more stuff\n- add new output\n\n### Fix\n\n- mama gotta work\n- output glitch\n\n## [1.0.0] - 2017-06-20\n### Added\n- New visual identity by [@tylerfortune8](https://github.com/tylerfortune8).\n- Version navigation.\n\n### Changed\n- Start using "changelog" over "change log" since it\'s the common usage.\n\n### Removed\n- Section about "changelog" vs "CHANGELOG".\n\n## [0.3.0] - 2015-12-03\n### Added\n- RU translation from [@aishek](https://github.com/aishek).\n"""
+    )
```
Lee-W, line 487: Wanna try "pytest-regressions"?  You can do so by adding `file_regresion` to this test case and then change the assert function to the following.

```python
file_regression.check(out, extension=".md")
```

The first time you run these tests it will fail as no file exists to be compared. pytest-regressions will generate the file so that you'll succeed the next time you run it. One thing to note is to check whether the output file is what we expect. It will be saved in a folder named "test_changelog_command"

tests/commands/test_changelog_command.py
```diff
@@ -453,3 +454,34 @@ def test_changelog_config_start_rev_option(mocker, capsys, config_path):
...
+    create_file_and_commit("feat: add new output")
+    create_file_and_commit("fix: output glitch")
+    create_file_and_commit("fix: mama gotta work")
+    create_file_and_commit("feat: add more stuff")
+    create_file_and_commit("Merge into master")
+
+    testargs = ["cz", "changelog", "--incremental"]
+
+    mocker.patch.object(sys, "argv", testargs)
+    # print(git.get_tags())
```
Lee-W, line 478: I think we could remove this line as well.

Lee-W (approved):
Just some minor suggestion. Other parts look good to me 👍

Lee-W (approved):

PR #400, opened 2021-07-08
feat(changelog): add support for single version and version range

### LINKED ISSUE #399: generate changelog for a single tag
## Description

Hi everyone, thanks for your work.
I'd like to generate my changelog with commitzen in ci, and store them in github releases instead of a whole changelog.md in my source files.

But It looks impossible now, I can only generate the full changelog, or split them manually from a specific tag.

## Possible Solution

add a filter arguments to filter single tag like:

```bash
cz changelog --tag v0.0.5
```

## additional information

There already has been a `--start-rev` argument that may help with this. But I think it would be better to have a way select to just a single tag for ci or other things to simplify the workflow.

This is inspired by https://github.com/git-chglog/git-chglog#cli-usag

Now I'm doing it with `git-chglog "$TAG" > chglog.md` and `gh release create "$TAG" -F chglog.md`

Woile:
Closes #399 #225

## Description

Allows the creation of a changelog for the given version, supports:

```bash
cz changelog --dry-run 1.0.0..2.0.0
```

```
cz changelog --dry-run 1.0.0
```

## Checklist

1. Move to the branch of this PR
2. Run the commands given above

## TODO

- [ ] Tests
- [ ] Syntax approval: do you think this syntax would be okay?

## Additional context

Closes #399 #225


commitizen/changelog.py
```diff
@@ -281,3 +281,49 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
+            keep = False
+            try:
+                accumulator.append(tags[index + 1])
+            except IndexError:
+                pass
+            break
+    return accumulator
+
+
+def get_star_and_end_rev(
```
Lee-W, line 306: We already have a similar function here https://github.com/commitizen-tools/commitizen/blob/master/commitizen/git.py#L71 Do you think we should extend that one instead?

commitizen/cli.py
```diff
@@ -183,6 +183,12 @@
                             "useful if the changelog has been manually modified"
                         ),
                     },
+                    {
+                        "name": "version_range",
```
Lee-W, line 187: I use rev-range for `cz check`. Maybe we could unify the naming

commitizen/changelog.py
Woile: I've checked, and they seem to be similar but they are fundamentally:
- one is performing a git command, the other is operating with `List[GitTag]` objects, no command performed.
- one receives start and end, the other works with a string

I think it's better to keep them separated. I've updated the PR to use the same naming.
What do you think?

Lee-W: Sounds good to me 👍 But it seems you have a typo here `get_star_and_end_rev` -> get_start_and_end_rev

Woile:
Lots of progress, I just have to fix and add some tests for the ranges:

`0.2.0..0.3.0`
`0.3.0..0.4.0`

Lee-W:
I might need some more time to fully review it but I just found out the reason of these failing tests. The date of [tmp_git_project](https://github.com/commitizen-tools/commitizen/blob/master/tests/conftest.py#L9) is decided by the time we run the tests which cannot be easily pinned by freezegun. Not sure whether there's work around for this case. If not, we might need to use traditional assertion instead of pytest-regression in the tests that depends on date.

Woile:
@Lee-W could you help me with this one? No clue why it's failing and they are passing locally. I patched the gittag date to avoid the problem with git timestamp and it works locally

tests/utils.py
```diff
@@ -19,3 +20,17 @@ def create_file_and_commit(message: str, filename: Optional[str] = None):
     Path(f"./{filename}").touch()
     cmd.run("git add .")
     git.commit(message)
+
+
+def wait_for_tag():
```
Woile, line 25: I had to add this func unfortunately

Woile:
@Lee-W can u take a look? I think is ready

Lee-W:
I'm a bit busy these days. Will try to take a look ASAP. 🙏

commitizen/changelog.py
```diff
@@ -281,3 +281,71 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
         output_lines.append(new_content)
     return output_lines
+
+
+def get_smart_tag_range(
+    tags: List[GitTag], start: str, end: Optional[str] = None
+) -> List[GitTag]:
+    """Smart because it finds the N+1 tag.
+
+    This is because we need to find until the next tag
```
Lee-W, line 291: May I know why do we need to find the next tag? For the case that `end` does not exist?

commitizen/changelog.py
```diff
@@ -281,3 +281,71 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
+            try:
+                accumulator.append(tags[index + 1])
+            except IndexError:
+                pass
+            break
+    return accumulator
+
+
+def get_start_and_end_rev(
+    tags: List[GitTag], version: str, tag_format: str, create_tag: Callable
```
Lee-W, line 313: May I know why do we need to make `create_tag` a `Callable` instead of using the one in `bump`? Are we going to extend it?

commitizen/changelog.py
```diff
@@ -281,3 +281,71 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
+        end = version
+
+    end_tag = create_tag(end, tag_format=tag_format)
+
+    start_tag = None
+    if start:
+        start_tag = create_tag(start, tag_format=tag_format)
+
+    tags_range = get_smart_tag_range(tags, start=end_tag, end=start_tag)
+    if len(tags_range) == 0:
```
Lee-W, line 335: Should we use `if not len(tags_range)` instead ?

commitizen/changelog.py
Lee-W: I also check the function `bump.create_tag`. It looks like `normalize_tag` to me as it does not create a new tag.

commitizen/git.py
```diff
@@ -39,11 +39,15 @@ class GitTag(GitObject):
     def __init__(self, name, rev, date):
         self.rev = rev.strip()
         self.name = name.strip()
-        self.date = date.strip()
+        self._date = date.strip()
 
     def __repr__(self):
         return f"GitTag('{self.name}', '{self.rev}', '{self.date}')"
 
+    @property
```
Lee-W, line 47: May I know why do we need this `property`?

tests/commands/test_changelog_command.py
```diff
@@ -514,3 +516,311 @@ def test_changelog_incremental_with_release_candidate_version(
...
+
+    testargs = ["cz", "changelog"]
+    mocker.patch.object(sys, "argv", testargs)
+    with pytest.raises(NotAllowed):
+        cli.main()
+
+
+@pytest.mark.usefixtures("tmp_commitizen_project")
+@pytest.mark.freeze_time("2022-02-13")
+@mock.patch("commitizen.git.GitTag.date", "2022-02-13")
```
Lee-W, line 537: Should we use `mocker` fixture in the test function? Or, would it cause any problem?

tests/commands/test_changelog_command.py
```diff
@@ -514,3 +516,311 @@ def test_changelog_incremental_with_release_candidate_version(
...
+    wait_for_tag()
+
+    create_file_and_commit("feat: after 0.2.0")
+    create_file_and_commit("feat: another feature")
+
+    testargs = ["cz", "bump", "--yes"]
+    mocker.patch.object(sys, "argv", testargs)
+    cli.main()
+
+    # time.sleep(1)
```
Lee-W, line 559: nitpick: we could remove the commented out code

tests/commands/test_changelog_command.py
```diff
@@ -514,3 +516,311 @@ def test_changelog_incremental_with_release_candidate_version(
...
+    mocker.patch.object(sys, "argv", testargs)
+    cli.main()
+
+    testargs = ["cz", "changelog", "0.2.0..0.3.0"]
+    mocker.patch.object(sys, "argv", testargs)
+    cli.main()
+    with open(changelog_path, "r") as f:
+        out = f.read()
+
+    print(out)
```
Lee-W, line 691: nitpick: maybe we should remove unused `print`

Lee-W (commented):
It's so much work 💪 Hope I can be as productive during next week's development sprint. I've left some comments and questions.

commitizen/changelog.py
Woile: Because we expect to receive the tag and 1 more in order to create the range of commits

commitizen/git.py
Woile: In order to mock it, the original value in `self.date` is now `self._date` and becomes the internal value

commitizen/changelog.py
Woile: I think you meant `if not tags_range:` right? updating...

commitizen/changelog.py
Woile: I renamed it to `normalize_tag`. I guess we can use the one directly in tag 👍🏻 
I don't recall why I was using a callable to be honest 🤔

commitizen/changelog.py
```diff
@@ -281,3 +282,71 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
+    start_tag = None
+    if start:
+        start_tag = normalize_tag(start, tag_format=tag_format)
+
+    tags_range = get_smart_tag_range(tags, start=end_tag, end=start_tag)
+    if not tags_range:
+        return None, None
+
+    start_rev: Optional[str] = tags_range[-1].name
+    end_rev = end_tag
```
Lee-W, line 340: nitpick: May I know do we need both `end_rev` and `end_tag` variables?

commitizen/changelog.py
```diff
@@ -281,3 +282,71 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
+
+    # check if it's the first tag created
+    # and it's also being requested as part of the range
+    if start_rev == tags[-1].name and start_rev == start_tag:
+        return None, end_rev
+
+    # when they are the same, and it's also the
+    # first tag crated
```
Lee-W, line 348: typo: created

commitizen/changelog.py
```diff
@@ -281,3 +282,71 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
...
+    except ValueError:
+        end = version
+
+    end_tag = normalize_tag(end, tag_format=tag_format)
+
+    start_tag = None
+    if start:
+        start_tag = normalize_tag(start, tag_format=tag_format)
+
+    tags_range = get_smart_tag_range(tags, start=end_tag, end=start_tag)
```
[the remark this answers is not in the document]
Woile: Somehow I find it very hard to reason about it, I have renamed to oldest-newest. Does it make more sense now?

commitizen/changelog.py
Woile: I think it's mostly as an alias, to make it more descriptive. The previous line is `start_rev` and to be consistent I added an end_rev. Otherwise `*_rev` and `*_tag` would be mixed. I can change it if you prefer so

Lee-W (approved):
Sorry for taking so long. Let's merge it!

Issue #402, opened 2021-07-11
Changelog change type sections are ordered "incorrectly"


lobotmcj:
## Description
The default value for `change_type_order` uses lower-case "feat", "perf", etc instead of "Feat", "Perf", and is therefore ignored in changelog output

## Steps to reproduce
1. `mkdir /tmp/cztest && cd /tmp/cztest`
2. `git init`
3. `touch foo && git add foo && git commit -m "feat: first commit"`
4. `touch foo2 && git add foo2 && git  commit -m "perf: second commit"`
![image](https://user-images.githubusercontent.com/32695581/125208418-317dd500-e260-11eb-9267-cbe9f1ed75b0.png)
5. `cz ch --dry-run`
![image](https://user-images.githubusercontent.com/32695581/125208460-81f53280-e260-11eb-9fb1-cee495e408dc.png)
6. `echo -e "[tool.commitizen]\nchange_type_order = [\"Feat\", \"Perf\"]" > pyproject.toml`
7. `cz ch --dry-run`
![image](https://user-images.githubusercontent.com/32695581/125208547-29726500-e261-11eb-9d43-7c2dc592f178.png)

## Current behavior
The default change_type_order is ineffective and pre #323 LIFO ordering is used in the changelog

## Desired behavior
Changelog sections should be ordered according to https://github.com/commitizen-tools/commitizen/blob/master/commitizen/defaults.py#L43 

## Environment
- commitizen version: 2.17.12
- python version: 3.6.9
- operating system: Linux

lobotmcj:
I'm not positive, but as far as I can tell, this only is in effect when using cz_conventional_commits and is caused by the change_type_map which maps "perf" to "Perf", etc here: https://github.com/commitizen-tools/commitizen/blob/aa0debe9ae5939afb54de5f26c7f0c395894e330/commitizen/cz/conventional_commits/conventional_commits.py#L35-L40

Lee-W:
closed by #403

PR #403, opened 2021-07-11
fix: capitalize types in default change_type_order

lobotmcj:
## Description
See #402 

## Checklist

## Expected behavior
See #402 

See #402 

## Additional context
See #402


Woile:
LGTM, may I ask a new test case to be added for this?
@Lee-W what do you think?

Lee-W:
LGTM as well. But might need some test case to verify the behavior

Lee-W:
@Woile I've rebased it and add test cases for it. Could you please take a look? Thanks!

commitizen/commands/changelog.py
```diff
@@ -45,7 +45,9 @@ def __init__(self, config: BaseConfig, args):
             self.config.settings.get("change_type_map") or self.cz.change_type_map
         )
         self.change_type_order = (
-            self.config.settings.get("change_type_order") or self.cz.change_type_order
+            self.config.settings.get("change_type_order")
+            or self.cz.change_type_order
+            or defaults.change_type_order
```
Woile, line 50: What would happen if you have a custom cz rule, without setting `change_type_order`? I see the `defaults` would be used, but would this explode?
It would always go into this line https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/changelog.py#L159

Do we have a test case for that situation?

lobotmcj:
> @Woile I've rebased it and add test cases for it. Could you please take a look? Thanks!

thank you, sir

commitizen/commands/changelog.py
Lee-W: we've not yet have a test case for it. I'll add one. Wondering why do you think it would explode? for the case that users use different change types?

Woile (approved):
Sorry for the mega long delay, I'm approving so you can merge when you feel like @Lee-W

Lee-W (approved):
@Woile I've rebased the latest changes and fix some of the error encountered. If no further comment added, I plan to merge this one these days.

Woile:
Sounds good

Issue #380, opened 2021-05-06
Improve selection prompt UX; Add single-character shortcut keys mnemonics 


webstoreportal:
We should aim for a quick win in keystroke reduction here without detracting from existing slow prompts.

* Allow user to confirm an option by first-letter (or second letter, in name clash) as a keyboard shortcut

https://github.com/commitizen-tools/commitizen/blob/32916abd2210e8c2c8852b42f51fcbbb01181f4b/commitizen/commands/commit.py#L50

https://github.com/commitizen-tools/commitizen/blob/32916abd2210e8c2c8852b42f51fcbbb01181f4b/commitizen/cz/conventional_commits/conventional_commits.py#L42

per Python Questionary dependency docs 
https://questionary.readthedocs.io/en/stable/pages/api_reference.html#questionary.Choice.shortcut_key

Slightly surprised no one has taken the opportunity to improve developer quality of life here. In commitizen-tools and cz-cli. Is there another alternative workflow we are missing out on?

Balancing forced verbose arrow navigation prompts vs command-line direct parsing

Adding the keyboard shortcut could be a happy medium solution

webstoreportal:
Wonder if this would affect customisable templates elsewhere

https://commitizen-tools.github.io/commitizen/customization/#2-customize-through-customizing-a-class

Lee-W:
Hi @webstoreportal Thanks for the suggestion!
@Woile I'm good with this feature. What do you think?

Woile:
We have key shortcuts now, I think this PR can be closed, what do you think?

Lee-W:
Yep, let's close it

PR #406, opened 2021-07-17
Add keyboard shortcuts to prompt

zevisert:
## Description

See the suggestion in #380.

- feat(prompt): add keyboard shortcuts with config option
  > Also added chore option to conventional_commits choices

This PR adds a config option called `use_shortcuts` which enables shortcuts in conventional commit choices. It defaults to false, so it'd be opt-in for now. In #380, the suggestion was to add the `use_shortcuts: bool` kwarg to questionary.prompt in [commands/commit.py](https://github.com/zevisert/commitizen/blob/65eb4f8b53f2e050b5ddbafa4b32439ef2f1a13d/commitizen/commands/commit.py#L50), but questionary passes this on to every question, and subsequently throws for prompt types that don't support `use_shortcuts`. The solution was to add `use_shortcuts` to the first question in [cz/conventional_commits/conventional_commits.py](https://github.com/zevisert/commitizen/blob/65eb4f8b53f2e050b5ddbafa4b32439ef2f1a13d/commitizen/cz/conventional_commits/conventional_commits.py#L48) instead.

## Checklist

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -45,16 +45,23 @@ class ConventionalCommitsCz(BaseCommitizen):
                 "type": "list",
                 "name": "prefix",
                 "message": "Select the type of change you are committing",
+                "use_shortcuts": self.config.settings['use_shortcuts'],
                 "choices": [
                     {
                         "value": "fix",
                         "name": "fix: A bug fix. Correlates with PATCH in SemVer",
+                        "key": "x",
                     },
                     {
                         "value": "feat",
                         "name": "feat: A new feature. Correlates with MINOR in SemVer",
+                        "key": "f",
                     },
-                    {"value": "docs", "name": "docs: Documentation only changes"},
+                    {
+                        "value": "docs",
+                        "name": "docs: Documentation only changes",
+                        "key": "d",
+                    },
                     {
                         "value": "style",
                         "name": (
@@ -62,6 +69,7 @@ class ConventionalCommitsCz(BaseCommitizen):
                             "meaning of the code (white-space, formatting,"
                             " missing semi-colons, etc)"
                         ),
+                        "key": "s",
                     },
                     {
                         "value": "refactor",
@@ -69,16 +77,19 @@ class ConventionalCommitsCz(BaseCommitizen):
                             "refactor: A code change that neither fixes "
                             "a bug nor adds a feature"
                         ),
+                        "key": "r",
                     },
                     {
                         "value": "perf",
                         "name": "perf: A code change that improves performance",
+                        "key": "p",
                     },
                     {
                         "value": "test",
                         "name": (
                             "test: Adding missing or correcting " "existing tests"
                         ),
+                        "key": "t",
                     },
                     {
                         "value": "build",
@@ -86,6 +97,7 @@ class ConventionalCommitsCz(BaseCommitizen):
                             "build: Changes that affect the build system or "
                             "external dependencies (example scopes: pip, docker, npm)"
                         ),
+                        "key": "b",
                     },
                     {
                         "value": "ci",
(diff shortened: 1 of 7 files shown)


zevisert:
I'll add tests later if you like the change. I noticed there wasn't complete approval in #380 yet.

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -45,54 +45,76 @@ def questions(self) -> List[Dict[str, Any]]:
...
                     {
                         "value": "ci",
                         "name": (
                             "ci: Changes to our CI configuration files and "
                             "scripts (example scopes: GitLabCI)"
                         ),
+                        "key": "c",
+                    },
+                    {
+                        "value": "chore",
```
Lee-W, line 111: We decide not to add "chore" as this is not required by conventional commit https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

Lee-W (changes requested):

Lee-W:
@zevisert Thanks for you contribution! This feature looks good to me @Woile What do you think?

Woile:
This is great! I like it!

zevisert:
Great, I'll follow up with some tests and docs in the next day or two - and sure I'll drop that chore option. I didn't mean to submit that here, that's just from me fiddling around.

zevisert:
Hey sorry for the delay! I rebased out the chore option, and added a test to ensure all of the options in any select questions in cz_conventional_commits has a keyboard key defined. Actually mocking out and asserting the shortcuts render in the terminal when the config option is enabled was a little too complicated IMO. I also updated the config tests to be aware of the new config option.

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -45,55 +45,68 @@ def questions(self) -> List[Dict[str, Any]]:
                 "type": "list",
                 "name": "prefix",
                 "message": "Select the type of change you are committing",
+                "use_shortcuts": self.config.settings['use_shortcuts'],
```
Lee-W, line 48: How about make it `settings['cz_conventional_commits']['use_shortcuts']`? This configuration might not make sense in other cz rules.

Lee-W:
Thanks for your update! But might need your help to reformat the code through `./scripts/format`

commitizen/cz/conventional_commits/conventional_commits.py
zevisert: Sure, conventional is the only pre-defined ruleset that uses a list select, but keep in mind that customized cz rules may also wish to use shortcuts.

Lee-W: Yep, that sounds reasonable. Once the documentation is updated. I think we're good to merge it. Thanks for this awesome feature!

zevisert: I moved this check from here to a filter in the actual commit, so that it automatically covers customized cz rules. Now use_shortcuts can apply to cz_customize in config files, and to classes defining `questions()` of their own.

zevisert:
Docs are updated, and I moved where we check the config for `use_shortcuts` to the `Commit` class, so that any `BaseCommitizen` class automatically gets access to this new feature. Thinking about my team and I, since we like to have `chore` included in our custom conventional commits rules.

zevisert:
Also, I mentioned it in the docs, but when there's no `key` provided, questionary conveniently falls back to numbering the options. This means that 3rd party templates aren't broken by this new setting! 🎉 

eg.; Using the [customize settings from the docs](https://commitizen-tools.github.io/commitizen/customization/#1-customize-in-configuration-file) with use_shortcuts enabled: 
```shell
 > poetry run cz commit
? Select the type of change you are committing (Use shortcuts or arrow keys)
 » 1) feature: A new feature.
   2) bug fix: A bug fix.
```

Lee-W (approved):

Lee-W:
didn't expect that handy feature. it's great. let's merge this one!

Issue #407, opened 2021-07-25
commitizen/commands/init.py:54: error: unused 'type: ignore' comment


karrukola:
## Description
When executing the scripts to run the tests the following error is reported: commitizen/commands/init.py:54: error: unused 'type: ignore' comment

## Steps to reproduce
Checkout master (tag v2.17.13 at time of writing)
1. Run ./scripts/test
2. Unit tests pass
3. 67 files will be left unchanged
4. mypy step fails -->

## Current behavior
```
commitizen/commands/init.py:54: error: unused 'type: ignore' comment
Found 1 error in 1 file (checked 59 source files)
```

## Desired behavior
this mypy error could be handled

**Screenshots**
None

## Environment

- commitizen version: 2.17.13
- python version: Python 3.8.10
- operating system: Linux

PR #408, opened 2021-07-25
fix(commands.init.py): Remove type ignore directive in Init._ask_config_path()

karrukola:
As it appears now choices=config_files does not cause any type mismatch with mypy so we can include this line in mypy's analysis.

Changelog generated with cz changelog command.

This should fix issue #407.

## Description
Removed the directive causing mypy to fail

## Checklist

- [ x ] Add test cases to all the changes you introduce
- [ x ] Run `./script/format` and `./script/test` locally to ensure this change passes linter check and test
- [ x ] Test the changes on the local machine manually
- [ x ] Update the documentation for the changes

## Expected behavior
No issues reported by running ./scripts/test

Run ./scripts/test on master; notice the failure.
Run ./scripts/test with this PR; failure is gone and no other issues are reported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+### Fix
+
+- **commands.init.py**: Remove type ignore directive in Init._ask_config_path()
 
 ## v2.17.13 (2021-07-14)
 
@@ -160,16 +165,14 @@
 ### Feat
 
 - add yaml as a config option
-
-### feat
-
 - **config**: add support for the new class YAMLConfig at the root of the confi internal package
+- **init**: add support for yaml config file at init
+
+### Fix
+
+- **YAMLConfig**: add a TypeError exception to handle in _parse_settings method
 
 ## v2.10.0 (2020-12-02)
-
-### Feat
-
-- **commitizen/cli**: add the integration with argcomplete
 
 ## v2.9.0 (2020-12-02)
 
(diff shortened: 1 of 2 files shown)


Woile:
Hey @karrukola thanks for the catch! I'll ask you to remove the changes to the changelog 🙏🏻 it's automatically generated once we merge the PR.

Regards

Lee-W:
Hi @karrukola , I'm thinking of making it a `style` commit instead of a `fix` one as this does not affect the functionality of commitizen. What do you think?

karrukola:
Hello @Woile, OK I will update the PR. I hope I won't have to forge anything... ...BTW perhaps there should be also an update of the _contributing_ section?

Hello @Lee-W, I have no strong opinion as this your project, with your rules. If you are not happy with `fix`, how about `test`? At the end of the day I got the error by trying to run _the unit tests_ via the script.

Woile:
Thanks, we will update the CONTRIBUTING guidelines, I didn't realize. `test` would also be fine

Lee-W:
@karrukola I'll say `style` might still be a better one if you agree we shouldn't use `fix`. Although named as `./script/test`, we run style checks as well. This is trigger by this line https://github.com/commitizen-tools/commitizen/blob/master/scripts/test#L12

karrukola:
PR updated.

Woile:
I think you have to squash the commits, otherwise it will still trigger a version bump.
In order to squash run the following:
```
git rebase -i origin/master
```
from bottom to top, change `pick` to `squash`, the one at the top select edit. Save and close.
Run `git commit --amend` and change the commit message to `style: ...`
run `git rebase --continue` to finish

Woile:
Already fixed this on another PR, closing. Thanks!

PR #420, opened 2021-09-11
fix(commit): correct the stage checker before commiting

### LINKED ISSUE #418: File stage checker not working as expected
## Description

If we have changed files not staged, and not staged files, then I don't get the "No files added to staging!".

## Steps to reproduce

Assuming you have a project with commitizen installed and the project is tracked by git, then we can do:
1. Edit an existing file
2. Run `cz commit`

## Current behavior

The questionnaire will pop up **without** erroring out with `No files added to staging!` since no files are yet added to stage.

## Desired behavior

The desired behaviour should be what the error says ("No files added to staging!"), so if no files are added to the stage, then the error should pop.

## Environment

- commitizen version: 2.17.11
- python version: Python 3.7.9
- operating system: Linux

This is an application bug, so it happens with any Python version and OS.

## Extra

This issue is related to #45 (fixed in #49), which was the original issue related to this problem, but in my opinion, was never fixed the right way. I am not randomly opening this issue, I am doing it because I bumped into this frustrating problem all the time where:

- I run `cz commit`
- Fill the questionnaire (~1 minute)
- Pre-commit runs (~1 minute)
- Get an error because the stage is empty
- Then, I have to stage the files, and run `cz commit --retry`

It seems that this was deliberately done in the [code](https://github.com/commitizen-tools/commitizen/blob/ec38ce3dcf19b4c0f704079e8ca8b0f1e3555b8e/commitizen/git.py#L159) since just this command is needed:

```
git diff --no-ext-diff --cached --name-only
```

The no-cached one will pick up files that are unstaged. In fact, in the more recent versions of  `git` one can use `--staged` flag which is an alias for `--cached`.

I understand if this was picked up from [here](https://github.com/commitizen/cz-cli/blob/62fc778a6207ec73a999e10fe7d1452a5f5dc05a/src/commitizen/staging.js#L9-L9) in [cz-cli](https://github.com/commitizen/cz-cli), but it has the same problem in there, and [this PR](https://github.com/commitizen/cz-cli/pull/818) is a proposal for the fix in that repo (not yet merged).

If this is a recognized bug, and we are willing to fix it, I offer myself to open the PR if needed.

Thank you for the great work and this amazing tool!

Regards,
Xavi M.

xaviml:
fixes #418

## Checklist

This part of the code was not originally tested, since changing the code did not break the existing tests. If interested, I could add tests for this with real git calls.

I did not add this to the documentation or README since it fixes a behaviour that was supposed to happen already.

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -155,9 +155,8 @@ def find_git_project_root() -> Optional[Path]:
 
 def is_staging_clean() -> bool:
     """Check if staing is clean."""
-    c = cmd.run("git diff --no-ext-diff --name-only")
-    c_cached = cmd.run("git diff --no-ext-diff --cached --name-only")
-    return not (bool(c.out) or bool(c_cached.out))
+    c = cmd.run("git diff --no-ext-diff --cached --name-only")
+    return not bool(c.out)
 
 
 def is_git_project() -> bool:
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -154,7 +154,7 @@ def find_git_project_root() -> Optional[Path]:
 
 
 def is_staging_clean() -> bool:
-    """Check if staing is clean."""
+    """Check if staging is clean."""
     c = cmd.run("git diff --no-ext-diff --cached --name-only")
     return not bool(c.out)
 
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -70,7 +70,7 @@ def test_get_commits_author_and_email():
     create_file_and_commit("fix: username exception")
     commit = git.get_commits()[0]
 
-    assert commit.author is not ""
+    assert commit.author != ""
     assert "@" in commit.author_email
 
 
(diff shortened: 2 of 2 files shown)


Woile:
Hi @xaviml thanks for the contribution! 
I do not understand the issue, so I'd like to ask you a test where it would fail before this PR and fixed after. It would also work as a way to document the behavior, and by seen it, I think I would understand better 👍🏻 

Sorry for the inconvinience 😅

xaviml:
Hi @Woile ,

Of course, I totally understand. Is there any test example in the test suite that actually test with the real git command? Is it okay if I do so? Otherwise, I don't see a way to test this properly since it only calls the git command.

Regards,
Xavi M.

Woile:
You can write it in https://github.com/commitizen-tools/commitizen/blob/master/tests/test_git.py

You can call git with this:

https://github.com/commitizen-tools/commitizen/blob/79b31566ab1e66560856ae1e813f702264aba06b/tests/test_git.py#L148

By using this fixture you'll have a git folder ready:
https://github.com/commitizen-tools/commitizen/blob/79b31566ab1e66560856ae1e813f702264aba06b/tests/test_git.py#L60

Example:

```python
def test_git_something():
   c = cmd.run("git something")
```

xaviml:
Hi @Woile,

Thank you for pointing me to the right test file. I did the following:
- Rename existing is_staging_clean test to `test_is_staging_clean_when_adding_file`
- Add `test_is_staging_clean_when_updating_file`

The new test fails in the second `assert` with the old code. The new code now checks correctly if stage is clean or not.

The bug was (as described in #418) that the is_stage_clean worked when adding a file (as it was tested), but not when modifying an existing file (it was not tested).

Let me know if anything needs changing.

Regards,
Xavi M.

Woile:
I see it now, wow I thought this was working.
Thanks!

Issue #318, opened 2020-12-21
I/O exception and does not commit


LukeSavefrogs:
## Description
Prints an I/O exception and continues. At the end doesn't commit to git but only prints the commit text

## Steps to reproduce
Type `cz commit` or `cz c` and continue

## Current behavior
```PowerShell
→ Test_Mattutini_PICO: [master ≡ +0 ~2 -0 !]› cz c
? Select the type of change you are committing fix: A bug fix. Correlates with PATCH in SemVer
? What is the scope of this change? (class or file name): (press [enter] to skip)
 C2C

Unhandled exception in event loop:
  File "c:\python38\lib\asyncio\proactor_events.py", line 768, in _loop_self_reading
    f.result()  # may raise
  File "c:\python38\lib\asyncio\windows_events.py", line 808, in _poll
    value = callback(transferred, key, ov)
  File "c:\python38\lib\asyncio\windows_events.py", line 457, in finish_recv
    raise ConnectionResetError(*exc.args)

Exception [WinError 995] Operazione di I/O terminata a causa dell'uscita dal thread oppure della richiesta di un'applicazione
Press ENTER to continue...
? Write a short and imperative summary of the code changes: (lower case and no period)
```

## Environment

- commitizen version: 2.11.1
- python version: Python 3.8.3
- operating system: Windows

Woile:
I think it's related to https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023
And happens because of windows and because of our dependency with `prompt-toolkit`.
Seems like a python bug: https://github.com/python/cpython/commit/8d68e59f11267ded765d4af6d71a49784a12fad5

Bogay:
According to this comment, It seems that this bug is fixed at python? But in a newer version (e.g. 3.8.7)
https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023#issuecomment-776072636

At least it works on my windows (I use python 3.8.7 instead of 3.8.3)

Lee-W:
@LukeSavefrogs in this case, I don't think we'll be able to fix the older version on our side. 
@Bogay Could you please add a FAQ page and add it as part of the documentation so that others will know they need to upgrade their python version for this to work? Thanks!

Bogay:
@Lee-W I found there exist 2 FAQ section: one in README and one in documentation, and the documentation one covers README one. Should I also removed the README's FAQ secion? (or leave the documentation link there)

Lee-W:
[...]

PR #431, opened 2021-09-26
doc(#318): add documentation about WinError 995

Bogay:
## Description

Add some description related to #318 so that others may find the solution on documentation.

## Checklist


docs/faq.md
```diff
@@ -73,3 +73,16 @@ If you are using conventional commits in your git history, then you could swap o
...
+## How to handle revert commits?
+
+```sh
+git revert --no-commit <SHA>
+git commit -m "revert: foo bar"
+```
+
+## I got `Exception [WinError 995] The I/O operation ...` error
+
+This error is cause by a python bug on windows and it should be fixed in some newer version. (e.g. 3.8.7) So if you encounter this error. Upgrade python version may solve this.
```
Lee-W, line 86: I might reword it to

```
This error was caused by a Python bug on Windows. It's been fixed and tested on Python 3.8.7. Upgrade your Python version might solve this issue.
```

Lee-W (commented):
Other than the minor documentation reword. Everything looks good. Thanks!

docs/faq.md
Bogay: But 3.8.7 is not the only version that include this fix. According to Python's changelog, [3.8.6rc1](https://docs.python.org/3.8/whatsnew/changelog.html#python-3-8-6-release-candidate-1) and [3.9.0rc2](https://docs.python.org/3.9/whatsnew/changelog.html#python-3-9-0-release-candidate-2) should be the accurate versions first contain https://github.com/python/cpython/pull/22017.  
I am afraid that just say it has been fixed and tested on Python 3.8.7 might mislead readers this error would not happened on Python 3.9. Should I write these information to the doc?

docs/README.md
```diff
@@ -175,23 +175,11 @@ See [Third-Party Commitizen Templates](third-party-commitizen.md).
...
-
-See more discussion in issue [#142](https://github.com/commitizen-tools/commitizen/issues/142) and [#36](https://github.com/commitizen-tools/commitizen/issues/36)
-
-### How to handle revert commits?
-
-```sh
-git revert --no-commit <SHA>
-git commit -m "revert: foo bar"
-```
+See [FAQ - commitizen](https://commitizen-tools.github.io/commitizen/faq/).
```
Bogay, line 178: @Lee-W Should I replace the URL with filename? It seems that filenames will be rendered as URL in generated documentation.
For example, `contributing.md` will turn into https://commitizen-tools.github.io/commitizen/contributing/

docs/faq.md
Lee-W: Yep, that would definitely be super helpful! Thanks for digging into it

docs/README.md
Lee-W: Yep, that would be even much better. 💯 didn't notice this one.

Bogay: Should I also replace those URLs inside other files? But I am not sure how to reference files not in the same directory.  
For example, near the end of this line, `docs/tutorials/gitlab_ci.md` refers to `docs/bump.md`.

https://github.com/commitizen-tools/commitizen/blob/061f47c1434f2df2da49579ff92e489001ecd3ec/docs/tutorials/gitlab_ci.md#L13

There seems to exist 2 valid path? `../bump.md` and `/docs/bump.md`, I am not sure which one is better and whether they will work in the final documentation.

Lee-W: I would prefer `../bump.md` in this case. If my understanding is correct, when deployed to gh pages `/docs/bump.md` becomes `https://commitizen-tools.github.io/docs/bump.md`

Lee-W: And yes, if you found any other problem, it would be better if you could fix them as well

Bogay: I have updated all the links I can find in documentation, and I found that absolute path like `/docs/bump.md` is not support by MkDocs, the documentation is [here](https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages).

docs/README.md
```diff
@@ -175,23 +175,11 @@ See [Third-Party Commitizen Templates](third-party-commitizen.md).
 
 ## FAQ
```
Lee-W, line 176: I think we could remove this section. as the content has been moved to FAQ page

docs/config.md
```diff
@@ -134,9 +134,9 @@ commitizen:
 | `customize`      | `dict` | `None`                      | **This is only supported when config through `toml`.** Custom rules for committing and bumping. [See more](customization)                                     |
 | `use_shortcuts`  | `bool` | `false`                     | If enabled, commitizen will show keyboard shortcuts when selecting from a list. Define a `key` for each of your choices to set the key. [See more](shortcuts) |
 
-[version_files]: https://commitizen-tools.github.io/commitizen/bump/#version_files
```
Lee-W, line 137: This syntax doesn't seem to work. Could you please check again?

docs/tutorials/writing_commits.md
```diff
@@ -38,5 +38,5 @@ understand what happened.
 | `fix(commands): bump error when no user provided` | `fix: stuff` |
 | `feat: add new commit command` | `feat: commit command introduced` |
 
-[customization]: https://commitizen-tools.github.io/commitizen/customization/
+[customization]: customization.md
```
Lee-W, line 41: This does not work either

Lee-W (changes requested):
Hi, sorry for the late review. As PyCon TW approaches, I get less time to devote to this project. Here are some errors I found on the changes.

You could test out through

```sh
poetry run mkdocs build
poetry run mkdocs serve
```

Would be great if you could add these commands to "how to contribute" so that others can have a better idea on how to test documentation changes

Lee-W:
Also, you might need to rebase the latest master branch into this branch to pass the test fail

docs/README.md
Bogay: OK, should I also remove "Third-Party Commitizen Templates" and "Contributing" sections?
The content is also in other pages and only one link in README.

Lee-W:
I still see the warning here.
image

Maybe you could try the following command to fix it

```sh
git remote add upstream git@github.com:commitizen-tools/commitizen.git
git fetch upstream 
git rebase -i remotes/upstream/main
```

Bogay:
@Lee-W sorry for lately fix and thank you for helping me to rebase the upstream. I have finished those changes.
But the test coverage seems to decrease a little, I have no idea about that. Did I break something?

PR #439, opened 2021-09-29
feat: add signoff parameter to commit command

manang:
command to sign off the commit, equivalent to git commit -s

This feature permits to execute "git commit -s" command

## Description
when cz commit --signoff is launched, the command git commit -s is executed

## Checklist

## Expected behavior
Signed-off-by message added to the footer of the commit

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -49,6 +49,11 @@ data = {
                         "action": "store_true",
                         "help": "show output to stdout, no commit, no modified files",
                     },
+                    {
+                        "name": "--signoff",
+                        "action": "store_true",
+                        "help": "Sign off the commit",
+                    },
                 ],
             },
             {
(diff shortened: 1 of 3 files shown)


Lee-W:
@Woile I'll say this is an ok temporary solution for additional arguments. I think we'll eventually use something like `--`, but I'm good with this. What do you think?

commitizen/cli.py
```diff
@@ -49,6 +49,11 @@
                         "action": "store_true",
                         "help": "show output to stdout, no commit, no modified files",
                     },
+                    {
+                        "name": "--signoff",
```
Woile, line 53: I'd suggest also supporting git's interface directly `-s`:

```suggestion
                        "name": ["s", "--signoff"],
```

Woile:
I like it and I'm fine with it, only thing missing is some docs 👍🏻

commitizen/cli.py
manang: ok, let me add this to the commit, and the docs
thansk

manang: I committed another version, I want to solve today:) thanks

Woile:
You got it! 🎉

PR #460, opened 2021-12-14
build: use poetry_core to build instead of poetry

dtrodrigues:
## Description
Changes the build backend from poetry to poetry_core. poetry_core is lighter weight than poetry and all that's required for the build vs project management aspect of poetry.

## Checklist

## Additional context
https://python-poetry.org/docs/pyproject/#poetry-and-pep-517
>If your pyproject.toml file still references poetry directly as a build backend, you should update it to reference poetry-core instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,5 +116,5 @@ known_first_party = ["commitizen", "tests"]
     ]
 
 [build-system]
-requires = ["poetry>=0.12"]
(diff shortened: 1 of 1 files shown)


Lee-W (approved):
I'm good with it. Seem to be the recommended way https://python-poetry.org/docs/pyproject/#poetry-and-pep-517

@Woile what do you think?

PR #471, opened 2022-01-14
docs: removing customization file contradiction

chr15murray:
## Description
Customising via the configuration file is clearly possible when using yaml files since you show an example for this. This PR updates the doc to remove the warning that states this is not possible. Given that yaml was the only missing file type in the original warning it seems safe to remove it completely.

## Checklist

- [n/a] Add test cases to all the changes you introduce
- [n/a] Run `./script/format` and `./script/test` locally to ensure this change passes linter check and test
- [n/a] Test the changes on the local machine manually
- [*] Update the documentation for the changes

## Expected behavior
N/A - Docs updated only

N/A - Docs updated only

## Additional context
N/A - Docs updated only

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -3,8 +3,6 @@ We have two different ways to do so.
 
 ## 1. Customize in configuration file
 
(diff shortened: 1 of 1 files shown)


Lee-W (approved):
Thanks for fixing the doc!

PR #473, opened 2022-01-16
fix(bump): raise non zero error code when there's no elegible commit to bump

Woile:
## Description

See error in https://github.com/commitizen-tools/commitizen-action/issues/18

## Checklist

- [ ] ~Update the documentation for the changes~

## Expected behavior

An error code should be returned when the bump can not succeed

Clone this repo https://github.com/now8-org/app

and run `cz bump`

## Additional context

1. Is there a reason why we weren't rising an error code in this situation? 🤔 
2. Am I missing any use case?

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -170,7 +170,10 @@ class Bump:
             out.write(information)
 
         if increment is None and new_tag_version == current_tag_version:
-            raise NoneIncrementExit()
+            raise NoneIncrementExit(
+                "[NO_COMMITS_TO_BUMP]\n"
+                "The commits found are not elegible to be bumped"
+            )
 
         # Do not perform operations over files or git.
         if dry_run:
--- a/commitizen/exceptions.py
+++ b/commitizen/exceptions.py
@@ -54,8 +54,8 @@ class DryRunExit(ExpectedExit):
     pass
 
 
-class NoneIncrementExit(ExpectedExit):
-    pass
+class NoneIncrementExit(CommitizenException):
+    exit_code = ExitCode.NO_COMMITS_FOUND
 
 
 class NoCommitizenFoundException(CommitizenException):
--- a/tests/commands/test_bump_command.py
+++ b/tests/commands/test_bump_command.py
@@ -8,8 +8,10 @@ import commitizen.commands.bump as bump
 from commitizen import cli, cmd, git
 from commitizen.exceptions import (
     BumpTagFailedError,
+    CommitizenException,
     CurrentVersionNotFoundError,
     DryRunExit,
+    ExitCode,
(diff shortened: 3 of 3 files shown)


commitizen/commands/bump.py
```diff
@@ -170,7 +170,10 @@ def __call__(self):  # noqa: C901
             out.write(information)
 
         if increment is None and new_tag_version == current_tag_version:
-            raise NoneIncrementExit()
+            raise NoneIncrementExit(
+                "[NO_COMMITS_TO_BUMP]\n"
```
Lee-W, line 174: nitpick: do we need the header here?

Lee-W (approved):

commitizen/commands/bump.py
Woile: Because it's raising the same error code as no commits found, I thought about outputing similar error with information

Lee-W: got it. Thanks! Then, I think we're ready to merge it 🎉

Woile:
I think I see the side effect of this. Now the pipelines throw an error when no change is found. 
But I guess is an edge we have to properly polish

Lee-W:
yep, after a second thought, this could be a problem. but i've not yet get a better idea

Lee-W:
look like we'll need to handle it on our side as well 🤔
maybe we'll need to add documentation for it?
https://github.com/commitizen-tools/commitizen/runs/4960945320?check_suite_focus=true

Woile:
I'm starting to think that I made a mistake here by using the same error code. 

On one hand you have new commits that you expect to be ignored (those commits starting with things like `docs`, `ci`, etc). 
And on the other hand, you have the "no commits" situation, where the pipeline is triggered and there's no new commits (this is probably because of a bad CI design, like triggering the pipeline when a new PR is made).

Maybe we can change the error code for this and introduce a new flag: `--no-raise` with error codes.
It would terminate the program but with error code `0`.
Example:
``` 
cz bump --no-raise 3,5,8
```
That would make it quite flexible, no?

Lee-W:
Yep! Sounds like a great solution! I was using `$?` in Makefile to decide whether I should ignore the error on other projects. Wondering if `--no-raise` is used in other project 🤔 Also, I think it might be great if we could support string input as well.

e.g., 

```sh
cz [any operation] --no-raise NO_COMMITS,NO_BUMP
```

or should we go with 

```sh
cz [any operation] --no-raise NO_COMIITS --no-raise NO_BUMP
```

I have no strong preference between these 2. Or we could support both? 🤔

Lee-W:
I just create an issue for this https://github.com/commitizen-tools/commitizen/issues/485. let's discuss there 💪

PR #479, opened 2022-01-28
Ignore packages that are not plugins

mmartinortiz:
When a package starting with a plugin name is found, but it is not a plugin, it becomes safely ignored.

## Description

When a packages has a name that matches a plugin, but it is not a plugin, it is safely ignored and reported to the user.

Background: The company where I work, and where I use `commitizen` for releasing package versions, is called CZ. And because we are not very imaginative, we use to call the packages that we develop for internal use something like `cz_`.

This collides with the way that `commitizen` discover plugins, and the discovery process raises an exception when a package starts with the name `cz_` is found but is not implemented as a plugin.

This MR proposes a way to ignore those packages that collide in name but are not plugins. This proposal tries to be as agnostic as possible, not very Pythonic, but I am open to suggestions.

## Checklist

## Expected behavior

When the CLI searches for plugins, those packages that have a name `cz_` but lack the `discover_this` attribute, will be ignored and the user will get a warning

1. Create a package with a name starting with `cz_`. It can be located under a `src` folder
2. Install `commitizen`
3. Run `commitizen changelog --help`. Without this patch, an exception will be risen.

### DIFF
--- a/commitizen/cz/__init__.py
+++ b/commitizen/cz/__init__.py
@@ -1,21 +1,40 @@
 import importlib
 import pkgutil
-from typing import Dict, Type
+import warnings
+from pathlib import Path
+from typing import Dict, Iterable, Type
 
 from commitizen.cz.base import BaseCommitizen
 from commitizen.cz.conventional_commits import ConventionalCommitsCz
 from commitizen.cz.customize import CustomizeCommitsCz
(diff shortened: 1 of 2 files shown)


mmartinortiz:
I'm sorry, I'm afraid that I messed up my own repository after a wrong click...

PR #480, opened 2022-01-28
Safely ignore modules named `cz_` that are not plugins

mmartinortiz:
When a package starting with a plugin name is found, but it is not a plugin, it becomes safely ignored.

## Description

When a packages has a name that matches a plugin, but it is not a plugin, it is safely ignored and reported to the user.

Background: The company where I work, and where I use `commitizen` for releasing package versions, is called CZ. And because we are not very imaginative, we use to call the packages that we develop for internal use something like `cz_`.

This collides with the way that `commitizen` discover plugins, and the discovery process raises an exception when a package starts with the name `cz_` is found but is not implemented as a plugin.

This MR proposes a way to ignore those packages that collide in name but are not plugins. This proposal tries to be as agnostic as possible, not very Pythonic, but I am open to suggestions.

## Checklist

## Expected behavior

When the CLI searches for plugins, those packages that have a name `cz_` but lack the `discover_this` attribute, will be ignored and the user will get a warning

1. Create a package with a name starting with `cz_`. It can be located under a `src` folder
2. Install `commitizen`
3. Run `commitizen changelog --help`. Without this patch, an exception will be risen.

### DIFF
--- a/commitizen/cz/__init__.py
+++ b/commitizen/cz/__init__.py
@@ -1,7 +1,6 @@
 import importlib
 import pkgutil
 import warnings
-from pathlib import Path
 from typing import Dict, Iterable, Type
 
 from commitizen.cz.base import BaseCommitizen
@@ -10,7 +9,7 @@ from commitizen.cz.customize import CustomizeCommitsCz
 from commitizen.cz.jira import JiraSmartCz
 
 
-def discover_plugins(path: Iterable[Path] = None) -> Dict[str, Type[BaseCommitizen]]:
+def discover_plugins(path: Iterable[str] = None) -> Dict[str, Type[BaseCommitizen]]:
     """Discover commitizen plugins on the path
 
     Args:
--- a/commitizen/cz/__init__.py
+++ b/commitizen/cz/__init__.py
@@ -23,7 +23,7 @@ def discover_plugins(path: Iterable[str] = None) -> Dict[str, Type[BaseCommitize
     for finder, name, ispkg in pkgutil.iter_modules(path):
         try:
             if name.startswith("cz_"):
-                plugins[name] = importlib.import_module(name).discover_this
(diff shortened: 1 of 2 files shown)


mmartinortiz:
Hi,

This is a terrible PR made at the wrong time of the day from the wrong operating system. I'll fix this and update the PR in the next days.

mmartinortiz:
This pull request is ready to review. Please let me know if you require some additional changes.

Woile:
Looks good to me, @Lee-W ?

commitizen/cz/__init__.py
```diff
@@ -1,21 +1,39 @@
...
+
+    Returns:
+        Dict[str, Type[BaseCommitizen]]: Registry with found plugins
+    """
+    plugins = {}
+    for finder, name, ispkg in pkgutil.iter_modules(path):
+        try:
+            if name.startswith("cz_"):
+                plugins[name] = importlib.import_module(name).discover_this  # type: ignore
+        except AttributeError as e:
```
Lee-W, line 27: @Woile do you think we should go with the exit code method in this case as well?

Woile: What do you mean? like a code to easily identify where this warning is coming from?
I'm not sure it's needed, the warning should point to the line in the code AFAIK 🤔

Lee-W: like what we define [here](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/exceptions.py#L6) so that user can decide whether to suppress the warning. but I'm ok with this PR now. we could create another one if we think that's necessary. it won't break anything if we do that in the future

Lee-W (approved):

commitizen/cz/__init__.py
Lee-W: I just approved it. If you agree with not handling this at this moment, I think we're good to merge it

PR #486, opened 2022-02-14
First take at `ConventionalCommitsAppenderCz`

stdedos:
## Checklist

## Expected behavior

I would like to have `ConventionalCommitsCz`, but also be able to extend it locally, as per project requirements/wishes - also to fit local workflows (i.e. use `fixup!`, `squash!`, or throwaway commits without the extra pain of disabling the local commit hooks)

As indicated by the PR, this is just a RFC: I would like to get the project's insight on how this could be achieved, since these changes touch on internal committizen parts that you might not necessarily want changed.
Also, I failed to find a tutorial on how to create my own plugin.
Finally, I've never used `poetry` before; and therefore everything is broken/unknown to me.

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -5,6 +5,7 @@ from typing import Dict, Optional
 
 from commitizen import factory, git, out
 from commitizen.config import BaseConfig
+from commitizen.cz import ConventionalCommitsAppenderCz
 from commitizen.exceptions import (
     InvalidCommandArgumentError,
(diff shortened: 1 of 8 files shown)


Woile:
Could this be started as a separate project? Like one of the ones listed here: https://commitizen-tools.github.io/commitizen/third-party-commitizen/

Lee-W:
As suggested by @woile , this could be a separate one. I'll close this.

Issue #492, opened 2022-03-08
option to specify newline types


CanisHelix:
### Description

I have pre-commit hooks to enforce LF line endings, this fails on every commitizen commit as it writes CRLF on windows. An option to configure this would be good.

### Possible Solution

Hacky, but I have customized my local commitizen and found all instances of `open` and added `newline='\n'` to them if they are `w` opens. And option in the .cz.yaml to specify this would be the optimal method.

Lee-W:
@Woile I'm good with it and think it might be helpful. What do you think?

Woile:
I'm also good with it as an option in the configuration

PR #493, opened 2022-03-13
feat: determine newline to write with Git

pan93412:
## Description

Determine the newline character to write with `git config core.eof`.

## Checklist

## Expected behavior
It will adopt the newline character from the configuration key `core.elf` of Git while writing to any file.

See tests.

## Additional context

Help me cc [CanisHelix](https://github.com/CanisHelix) check this commit. Thanks!

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -1,9 +1,32 @@
 import os
+from enum import Enum
+from os import linesep
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import List, Optional
 
 from commitizen import cmd
+
+UNIX_EOL = "\n"
+WINDOWS_EOL = "\r\n"
+
+
+class EOLTypes(Enum):
+    """The EOL type from `git config core.eol`."""
+
+    LF = "lf"
+    CRLF = "crlf"
+    NATIVE = "native"
+
+    def get_eol_for_open(self) -> str:
+        """Get the EOL character for `open()`."""
+        map = {
+            EOLTypes.CRLF: WINDOWS_EOL,
+            EOLTypes.LF: UNIX_EOL,
+            EOLTypes.NATIVE: linesep,
+        }
+
+        return map[self]
 
 
 class GitObject:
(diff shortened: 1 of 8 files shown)


Woile:
LGTM.Is there an issue associated? What's the problem that it's trying to solve?

pan93412:
> LGTM.Is there an issue associated? What's the problem that it's trying to solve?

I think it is https://github.com/commitizen-tools/commitizen/issues/492.

Woile (approved):
LGTM. Could you rebase it? And if possible just one commit containing the feature added instead of multiple independent "feats". Something like:
`feat: detect OS specific EOF`

Thanks!

commitizen/git.py
```diff
@@ -164,3 +187,31 @@ def is_git_project() -> bool:
     if c.out.strip() == "true":
         return True
     return False
+
+
+def get_eol_style() -> EOLTypes:
+    c = cmd.run("git config core.eol")
```
gpongelli, line 193: Can you improve this part of code running the git command only at the first call of `smart_open`, then saving `eol` for future `smart_open` calls ?

Lee-W: Yep, it could be an improvement. But might not be an urgent one. I don't think it would be a huge overhead

Lee-W:
I notice there's a conflict. Could you please rebase and resolve it? If you don't have the time to do so. I'm happy to work on it next week :)

commitizen/git.py
pan93412: It may be a little hard – every repositories have their different configuration, and cache the configuration with their repository hash may be slower than run this command directly:

```
$ hyperfine "git config core.eol" -i -N -w 1000 -r 5000

Benchmark 1: git config core.eol
  Time (mean ± σ):       3.0 ms ±   0.4 ms    [User: 1.4 ms, System: 0.7 ms]
  Range (min … max):     2.6 ms …  10.5 ms    5000 runs
 
  Warning: Ignoring non-zero exit code.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

It took about 700us to run `git config core.eol`, which may not be a noticeable overhead when looking into other Git commands that will be executed.

pan93412:
@Lee-W resolved ;) thanks!

Lee-W (approved):

Lee-W:
@pan93412 Sorry for taking so long. Thanks for your contribution! Let's merge it 🎉

PR #497, opened 2022-04-11
Make cz schema pattern more precise

### LINKED ISSUE #449: cz check does not detect incorrectly formatted commit message
## Description

Looks like cz does not detect properly incorrect commit messages, it only parses first line?

## Steps to reproduce
1. create file named BAD_MSG with the following content:
```text
ci: check commit message on merge
testing with more complex commit mes

age with error
```
as you can see there is no empty line between first and second line.

2. run `cz check --commit-msg-file BAD_MSG`

## Current behavior
```
Commit validation: successful!
```

## Desired behavior
```
commit validation: failed!
please enter a commit message in the commitizen format.
commit "": "CI_COMMIT_MESSAGE"
pattern: (build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)(\(\S+\))?!?:(\s.*)
```

## Environment

cz version -r
```
Commitizen Version: 2.20.0
Python Version: 3.8.10 (default, Sep 28 2021, 16:10:42) 
[GCC 9.3.0]
Operating System: Linux
```

Bogay:
Closes #449 

## Description

Currently the schema pattern of `cz` only checks the first line of commit message, so that some invalid format cannot be detected by `cz check`. I update the pattern to ensure there is a blank line between subject and body (if exists) and add some comments to make the regex a little more readable.

Meanwhile, because the inconsistent implementation of `Check._get_commits` for `--commit-msg-file` and `--message`, give a commit message file missing the blank line will also pass the validation (the missing line will be added by `Check`).

## Checklist

## Expected behavior

Miss blank line between subject and body will not pass `cz check`.

```bash
cat << EOF > /tmp/BAD_MSG                                                                                                                                          
ci: check commit message on merge
testing with more complex commit mes

age with error
EOF
cz check --commit-message-file /tmp/BAD_MSG
```

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -191,8 +191,11 @@ class ConventionalCommitsCz(BaseCommitizen):
 
     def schema_pattern(self) -> str:
         PATTERN = (
-            r"(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)"
-            r"(\(\S+\))?!?:(\s.*)"
+            r"(?s)"  # To explictly make . match new line
+            r"(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)"  # type
+            r"(\(\S+\))?!?:"  # scope
+            r"( [^\n\r]+)"  # subject
+            r"((\n\n.*)|(\s*))?$"
         )
         return PATTERN
 
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -82,20 +82,18 @@ class Check:
         out.success("Commit validation: successful!")
 
     def _get_commits(self):
+        msg = None
         # Get commit message from file (--commit-msg-file)
         if self.commit_msg_file is not None:
             # Enter this branch if commit_msg_file is "".
             with open(self.commit_msg_file, "r", encoding="utf-8") as commit_file:
                 msg = commit_file.read()
+        # Get commit message from command line (--message)
+        elif self.commit_msg:
+            msg = self.commit_msg
+        if msg is not None:
             msg = self._filter_comments(msg)
-            msg = msg.lstrip("\n")
-            commit_title = msg.split("\n")[0]
-            commit_body = "\n".join(msg.split("\n")[1:])
-            return [git.GitCommit(rev="", title=commit_title, body=commit_body)]
-        elif self.commit_msg is not None:
-            # Enter this branch if commit_msg is "".
-            self.commit_msg = self._filter_comments(self.commit_msg)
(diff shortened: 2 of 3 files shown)


Woile:
Wow! @Lee-W what do you think of this feature?

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -190,8 +190,11 @@ def schema(self) -> str:
 
     def schema_pattern(self) -> str:
         PATTERN = (
-            r"(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)"
-            r"(\(\S+\))?!?:(\s.*)"
+            r"(?s)"  # To explictly make . match new line
```
Lee-W, line 193: I'm not familiar with re enough to understand what this means. Could you please describe a bit more on why this is added?

Lee-W: rest of the stuff looks good to me  👍

Bogay: `(?s)` is an inline flag, which is the same as `re.DOTALL` ([ref](https://docs.python.org/3/library/re.html#re.DOTALL)). This is needed because I add a `$` to make this schema pattern to match the whole commit message instead of matching the first line only. If the flag is not set, `\n\n.*` only allows one line of message body.

You can use the following code to verify that behavior:
```python
import re

PATTERN = (
    r"(?s)"
    r"(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)"  # type
    r"(\(\S+\))?!?:"  # scope
    r"( [^\n\r]+)"  # subject
    r"((\n\n.*)|(\s*))?$"
)

# Shouldn't be `None`
print(
    re.match(
        PATTERN,
        (
            "refactor(docs): remove README.rst and use docs/README.md\n\n"
            "By removing README.rst, we no longer need to maintain "
            "two document with almost the same content\n"
            "Github can read docs/README.md as README for the project."
        ),
    ),
)
# Should be None
print(
    re.match(PATTERN, "feat: an awesome feature\nwith an ill-formated commit message"),
)
```

Bogay: @Lee-W Should I add more explanation to that regex pattern?

Lee-W: nope, but I think we could add a link as comment to this issue.

---

Thanks for the explanation! Now I understand it. Just want to make sure, did you forget to remove the "(?s)" in this example? If I remove "($s)", then the result is just like what you describe.

Lee-W (approved):

Lee-W:
@Woile I'm good with this one and plan to merge it these days if no further comment added. Thanks!

Issue #502, opened 2022-04-21
cz bump fails if pre-commit reformats CHANGELOG.md


yajo:
### Description

When I use `cz bump --changelog`, it tries to commit the updated changelog, but fails because of pre-commit. Then I get stuck.

### Steps to reproduce

1. Add have a [pre-commit](https://pre-commit.com/) config that applies [prettier](https://prettier.io/) over markdown files.
2. Configure it with 4 spaces.
3. Add some changes that produce a bump.
4. `cz bump --changelog`

### Current behavior

```
> cz bump
bump: version 0.1.0 → 0.1.1
tag to create: v0.1.1
increment detected: PATCH

git.commit error: "prettier.................................................................Failed
- hook id: prettier
- files were modified by this hook

CHANGELOG.md"

> cz bump
Tag v0.1.1 could not be found. 
Possible causes:
- version in configuration is not the current version
- tag_format is missing, check them using 'git tag --list'

? Is this the first tag created? No
[NO_COMMITS_FOUND]
No new commits found.
```

### Desired behavior

It should do at least 2 commit attempts, just in case the 1st one fails due to a pre-commit hook.

Otherwise, it should revert version bump diff, to let me fix the changelog manually and issue `cz bump` again.

### Environment

Commitizen Version: 2.24.0
Python Version: 3.10.4 (main, Mar 25 2022, 00:00:00) [GCC 11.2.1 20220127 (Red Hat 11.2.1-9)]
Operating System: Linux

PR #505, opened 2022-04-28
fix: respect pre-commit reformats when bumping

yajo:
## Description

Fix https://github.com/commitizen-tools/commitizen/issues/502. Just attempt to commit twice if the 1st commit fails.

## Checklist

## Expected behavior

Compatible with prettier.

Changelog is now compatible with prettier + editorconfig.


Woile:
Thanks!

@Lee-W ? LGTM

Kurt-von-Laven:
Automatically retrying something as potentially permanent as a commit seems quite risky considering Commitizen has no idea what caused the error. We also run Prettier and many other linters via pre-commit and have never had it automatically reformat anything generated by Commitizen. What types of changes is Prettier making?

yajo:
> Automatically retrying something as potentially permanent as a commit seems quite risky considering Commitizen has no idea what caused the error [...] its ramifications may be somewhat unpredictable

In reality there's no risk. If something fails because of reasons the 1st time (e.g.: disk is full, a linter fails...), it will fail with the same reasons the 2nd time.

_But_ if the reason is that a formatter changed the format of the CHANGELOG.md file, then doing the retry will be simply making cz compatible with any pre-commit toolkit out there.

I maintain a similar patch in Copier and it's worked perfectly for years.

I don't think we really need any option to implement this. AFAICS  It's a sane fix that works out of the box without any side effects.

> What types of changes is Prettier making? Indentation I'm guessing from your issue?

That's one of them. Another one is when a commit contains illegal (or imperfect) markdown in the title. For example, [this commit](https://gitlab.com/moduon/mrchef/-/commit/8655cd330c63b7875cf8275bd146dc004fd6238b) produced a line like:

```md
- use _mrchef origin only privately
```

and [Prettier modified it to add a `\` before the `_`](https://gitlab.com/moduon/mrchef/-/commit/b59003bddd0cd2821a2acf4761d7eafadc4ea2b5#ab09011fa121d0a2bb9fa4ca76094f2482b902b7_1_5) (apart from indentation).

Both use cases are exercised in the supplied test. Just remove the patch and run the test, you'll see the failure.

Kurt-von-Laven:
Consider what would happen if a tool (or even just a confused human) were attempting to parse the output though or something else (e.g., an alert regarding the error) was triggered an extra time unintentionally. It would be desirable to have the ability to configure Commitizen in a manner that would avoid alerting twice per error in such cases. Some commit failures are ephemeral, such as a pre-commit hook that conducts a security audit or otherwise hits the network for any reason whatsoever. A really frustrating situation could arise in the presence of a totally unrelated but buggy pre-commit hook that snuck in because it always worked when retried, but sometimes didn't work on the first try.

I meant to say that `n` would default to 0 when `--retry` is not present, but 1 when it is. I feel less strongly about the former point though. It sounds from your experience like retrying once is suitable in the vast majority of cases. In any case it might be nice to log a message saying that a retry is occuring (and how to disable it) so that it is clear which tool is triggering this behavior.

yajo:
If I understand you properly, you're worried that if, for example, somebody sets up a pre-commit hook that launches a rocket 🚀  every time they attempt to commit, we'd be sending two rockets 🚀 🚀 instead of one?

That's very unlikely to happen. And if it happens, it's probably more a bug in the use of pre-commit than a bug in commitizen. I don't think it's a use case that should worry us. The user would be sending 2 rockets anyway whenever prettier reformats another file.

Just think about this: if committing changes to CHANGELOG.md fails the 1st time, why would someone ever want to run `cz bump --changelog --retry=0`? It's gonna fail anyway, and you already have the (2nd) failure logs. 🤷🏼‍♂️ 

OTOH, if committing changes to CHANGELOG.md fails the 2nd time, why would you ever want to run `cz bump --changelog --retry=0`? You already have the logs of the 2nd failure, those will tell you the problem; besides, you also have the dirty changes in your repo. Why would you need more info?

The point is: if your use case was working before this patch, it will continue working (1st commit attempt passes OK). And if it was failing, it *could* get fixed (if it was a reformat), or it *could* continue failing (if it was a real failure). So there's nothing we lose with this fix (unless you send rockets with commits, of course). 😁

I'll add some logs for the retry however. That seems an easy way to tell what's happening behind the scenes, in case it matters for some users.

yajo:
Logs added. In debug mode, you get full error.

Kurt-von-Laven:
> If I understand you properly, you're worried that if, for example, somebody sets up a pre-commit hook that launches a rocket rocket every time they attempt to commit, we'd be sending two rockets rocket rocket instead of one?

Not at all. A "launch" button that fires two rockets instead of one violates the user design principle of least surprise even if it notifies the user it did so after the fact in the fine print (although that's clearly better than nothing).

> The user would be sending 2 rockets anyway whenever prettier reformats another file.

I don't know that I follow the metaphor, but if they wrote code that handles a commit failure in some manner, it would only be triggered once unless they themselves committed twice. Prettier doesn't automatically retry commits. Everyone accepts that two rockets will fire if they click "launch" twice.

> Just think about this: if committing changes to `CHANGELOG.md` fails the 1st time, why would someone ever want to run `cz bump --changelog --retry=0`? It's gonna fail anyway, and you already have the (2nd) failure logs. 🤷🏼‍♂️

> OTOH, if committing changes to CHANGELOG.md fails the 2nd time, why would you ever want to run cz bump --changelog --retry=0? You already have the logs of the 2nd failure, those will tell you the problem; besides, you also have the dirty changes in your repo. Why would you need more info?

> The point is: if your use case was working before this patch, it will continue working (1st commit attempt passes OK). And if it was failing, it could get fixed (if it was a reformat), or it could continue failing (if it was a real failure). So there's nothing we lose with this fix (unless you send rockets with commits, of course).

I am not following this, but the simplest reason someone would want to retry 0 times is that they have no reason to expect that retrying would ever be constructive (e.g., they aren't running auto-formatters as pre-commit hooks that may potentially clash with Commitizen). They may not want to spend time sorting through unexpected duplicate logs and/or time + money (e.g., build credits) re-running all of their other pre-commit hooks to no avail. They may not want an ephemeral error to be swallowed without even knowing what happened or that anything even had happened. And again, this retry logic can mask a bug in another hook that sometimes fails on the first attempt, but never on the second. On the other hand, they may not have any other pre-commit hooks, in which case retrying seems quite bizarre. There are a huge number of scenarios in computing where something can fail and be retried, and it's (a) quite rare for that retry behavior not to be configurable by the developer in polished APIs, and (b) not retrying is by-and-large the default, because it often results in more robust systems overall.

yajo:
Just try to think in one practical example where having this fix is worse than not having it. If you can't find any, it means that will just never happen. 🤷🏼‍♂️

Can't we just merge and fix that theoretical bug once your theories really happen somewhere? 🙄

Kurt-von-Laven:
In the most common case, retrying is not productive and may even be detrimental.

yajo:
I'll add the cli option.

yajo:
Option added. Now, if you don't run it with `cz bump --retry`, the old behavior is kept.

Kurt-von-Laven:
Thank you!

yajo:
So is there anything else left to merge?

Woile:
The feature seems to cover specific situation: `Maybe pre-commit reformatted some files? Retry once`.
It's important to add documentation about the use cases, so we know later why it's being introduced.

To avoid confusions, it would be nice to document the feature 👍🏻 thanks!

yajo:
Documented.

commitizen/commands/bump.py
```diff
@@ -209,16 +213,22 @@ def __call__(self):  # noqa: C901
...
         c = git.commit(message, args=self._get_commit_args())
+        if self.retry and c.return_code != 0 and self.changelog:
+            # Maybe pre-commit reformatted some files? Retry once
+            logger.debug("1st git.commit error: %s", c.err)
+            logger.info("1st commit attempt failed; retrying once")
+            cmd.run(f"git add {changelog_cmd.file_name} {' '.join(version_files)}")
+            c = git.commit(message, args=self._get_commit_args())
         if c.return_code != 0:
-            raise BumpCommitFailedError(f'git.commit error: "{c.err.strip()}"')
+            raise BumpCommitFailedError(f'2nd git.commit error: "{c.err.strip()}"')
```
Lee-W, line 231: Could we add a test case that the 2nd failed?

tests/test_bump_create_commit_message.py
```diff
@@ -20,3 +24,47 @@ def test_create_tag(test_input, expected):
...
+        # Prettier inherits editorconfig
+        Path(".editorconfig").write_text(
+            """
+            [*]
+            indent_size = 4
+            """
+        )
+        cmd.run("git add -A")
+        cmd.run("git commit -m 'fix: _test'")
+        cmd.run("pre-commit install")
```
Lee-W, line 59: Is it possible that `pre-commit` is not installed system wide and thus cause the failure of this line?

Lee-W (commented):
Overall, everything looks good to me! Thanks everyone's hard work on this. I love this "--retry" flag. Just left some questions. I think we're almost ready to merge this one.

tests/test_bump_create_commit_message.py
yajo: No, it's a dev dependency: https://github.com/commitizen-tools/commitizen/blob/713ce195afbfd449eb67c2fcd792b68e0b6048b4/pyproject.toml#L75

commitizen/commands/bump.py
yajo: Test added.

Lee-W: Thanks!

Lee-W (approved):
LGTM. Thanks @yajo for your contribution! @Woile I'd like to merge this one if no further comment is added these days

PR #511, opened 2022-05-04
Ensure At Least One Blank Line Between Old and New Changelog Content

### LINKED ISSUE #509: Ensure At Least One Blank Line Above When Updating Changelog
### Description

If you initialize the changelog with some additional documentation (e.g., "This changelog is auto-generated by the Python-based
[Commitizen](https://commitizen-tools.github.io/commitizen)."), then the first time you release, the auto-generated text appears immediately below the existing text, which looks fairly ugly. Subsequent releases have one blank line of whitespace between them. It's simple enough to manually add the extra blank line at the top in oneself (ideally before first running `cz bump`), but it would add a bit of polish to automatically insert the extra blank line on the first release just like all of the rest.

### Possible Solution

If the changelog already exists, the first release is no handled no differently than the rest. If there is no changelog, the current behavior should remain unchanged, meaning don't start adding an extra blank line at the top of the file.

### Additional context

I am happy to open a pull request if there are no concerns regarding this minor change. This makes it possible to include an initial changelog with some documentation in a template repository without violating a Prettier pre-commit hook.

### Additional context

_No response_

Kurt-von-Laven:
## Description

Ensure there is at least one blank line separating old and new content when appending incrementally to the changelog. Log entries for each release are already separated with blank lines, so separate pre-existing content from the first release with a blank line as well. If the pre-existing content ends with a blank line, then don't add an extra one.

I also fixed some unrelated pre-existing issues with `.pre-commit-config.yaml` in order to be able to comply with the contributing guidelines.

Please let me know which documentation I should update if you feel this change warrants mention. I wasn't sure whether to classify this as a tiny feature or a bug fix.

Closes #509.

## Checklist

## Expected behavior

If the changelog already says "This changelog is auto-generated by the Python-based
[Commitizen](https://commitizen-tools.github.io/commitizen).", then when the first release is added by Commitizen, it doesn't ram right up against the pre-existing content.

1. Create `CHANGELOG.md`, and add some content to it that doesn't end in an extra blank line (besides the one at the end of the file).
2. Make a feature commit.
3. Run `cz changelog --incremental`.
4. The changelog is updated with a note for the new feature appended to the bottom of the file. There is exactly one blank line between the pre-existing and new content.

### DIFF
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -235,7 +235,7 @@ def get_metadata(filepath: str) -> Dict:
     }
 
 
-def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
+def incremental_build(new_content: str, lines: List[str], metadata: Dict) -> List[str]:
     """Takes the original lines and updates with new_content.
 
     The metadata governs how to remove the old unreleased section and where to place the
@@ -253,7 +253,7 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
     unreleased_end = metadata.get("unreleased_end")
     latest_version_position = metadata.get("latest_version_position")
     skip = False
-    output_lines: List = []
+    output_lines: List[str] = []
     for index, line in enumerate(lines):
         if index == unreleased_start:
             skip = True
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -3,7 +3,8 @@ version = "2.24.0"
 tag_format = "v$version"
 version_files = [
   "pyproject.toml:version",
-  "commitizen/__version__.py"
+  "commitizen/__version__.py",
+  ".pre-commit-config.yaml:rev.\\s+(?=[^\\n]+Commitizen)"
```
Lee-W, line 7: This is neat!

commitizen/changelog.py
```diff
@@ -270,16 +270,14 @@ def incremental_build(new_content: str, lines: List, metadata: Dict) -> List:
         if skip:
             continue
 
-        if (
-            isinstance(latest_version_position, int)
```
Lee-W, line 274: Wondering why this condition is removed

Issue #451, opened 2021-11-05
Add --commit-args command line option to cz commit


adamcunnington-mlg:
Given `cz commit` wraps `git commit`, it would be great to be able to pass arbitrary args to the underlying git commit. This is a common pattern, for example, pip-tools' pip-compile has a `--pip-args` option.

Specifically, I need to be able to pass `--no-verify` to some commits to occasionally brute force supress hooks but I think you can just allow the user to pass a string of arbitrary `--commit-args` to cz commit which you then pass through to git commit.

Bogay:
I am interesting to implementing this feature.
But some of `git` arguments might conflict with `cz` arguments?
For example, `cz c --commit-args -m foo` should be invalid because the commit message will be generated by `cz`?
I am not sure how to process these cases. Just ignore them or print a warning?

Woile:
It seems similar to this discussion:
https://github.com/commitizen-tools/commitizen/issues/248

Which of all the proposed methods would be best?
Alternatively we can manually add one by one as they are required (which might not be the best)

Bogay:
@Woile I read the discussion of #248 and I think `--` is a good idea.
About handing invalid arguments, I come up with two approach:
1. Directly pass all arguments after `--` to `git commit` and let `git` handle them. Leave the responsibility of giving correct args to the user. But there are so many edge cases might brake `cz`'s behavior, which affects the UX.
2. Add a whitelist to filter which argument can use. For example, `-S/--gpg-sign`, `-s/--signoff`, `-n/--no-verify`. At least I saw issues which discuss about these features. But this is less flexible if the whitelist is hard-coded in source.

Lee-W:
I like the first solution a bit more. @Woile what do you think?

adamcunnington-mlg:
Hi all, thanks for the consideration on this ticket. Really keen for this - is anyone picking it up?

Lee-W:
@adamcunnington-mlg AFAIK, no one is picking it. Feel free to create a PR if you're interested in it. Thanks!

adamcunnington-mlg:
I've not had chance to pick this up but would really love this

SCrocky:
Is anybody working on this?

Lee-W:
AFAIK, there's no one working on it. Feel free to take it :)

adamcunnington-mlg:
@Lee-W did anyone review @SCrocky's MR?

Lee-W:
[...]

Issue #247, opened 2020-08-13
[Feature Request] Respect --allow-empty flag


alecgerona:
## Description
Respect git flag `--allow-empty` when using the cli tool.

## Possible Solution
Allow empty commits using the cli tool.

## Related Issue
Related to #49

Kurt-von-Laven:
Copying the request from linked issue here. It would be ideal to support `--allow-empty-message` as well.

Kurt-von-Laven:
Is #485 sufficiently general that its implementation would additionally address this issue for free?

Lee-W:
I think #485 is solving another issue but could be a workaround of this one. This one https://github.com/commitizen-tools/commitizen/issues/451#issuecomment-1092067170 might potentially be the general solution of this issue

Kurt-von-Laven:
Indeed, being able to pass arbitrary flags to `git commit` seems ideal!

PR #512, opened 2022-05-05
feat(check): Add `--allow-abort` option

### LINKED ISSUE #424: Aborted Commit Considered to Fail Validation When Run As commit-msg Hook
## Description

Git interprets an empty commit message to mean that a commit should be aborted, but calls all commit-msg hooks anyways. This causes pre-commit hooks that run as commit-msg hooks but don't guard against empty commit messages to spew errors en masse when the commit was already being intentionally aborted.

## Steps to reproduce
1. Configure commitizen to run as a commit-msg hook using pre-commit as described below.
2. `touch foo && git add && git commit`
3. Save an empty commit message.

For step 1, add the following to `.pre-commit-config.yaml`:

```
yaml
  - repo: https://github.com/commitizen-tools/commitizen
    rev: v2.18.1
    hooks:
      - id: commitizen
        stages:
          - commit-msg
```

## Current behavior

[Pre-commit follows git's lead and calls commitizen when the commit message is empty when run as a commit-msg hook](https://github.com/pre-commit/pre-commit/issues/2052). This leads to the following error:

```
commitizen check..............................................................Failed
- hook id: commitizen
- exit code: 14

commit validation: failed!
please enter a commit message in the commitizen format.
commit "": "# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch foo
#
# Changes to be committed:
#       ,,,
pattern: (build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)(\(\S+\))?!?:(\s.*)
```

The commit is still correctly aborted of course.

## Desired behavior

I expected the commitizen commit-msg hook to silently report success in this case since there is no commit message to check.

## Environment

- commitizen version: 2.18.1
- python version: Python 3.9.5
- operating system: Linux

Kurt-von-Laven:
## Description

Empty commit messages indicate to Git that the commit should be aborted. Displaying an error message when the commit is already being aborted typically only creates confusion. Add an `--allow-abort` argument to the check command and `allow_abort` config variable, both defaulting to false for backwards compatibility. When the commit message is empty, determine the outcome based on the `allow_abort` setting alone, ignoring the pattern.

Closes #424.

## Checklist

## Expected behavior

No error message is displayed when aborting a commit locally, unless the commit .

1. Run `cz commit`.
2. Save and close the commit message file that appears with an empty commit message.
3. No error occurs if you add `allow_abort = true` to your config, but an error occurs otherwise.

## Additional context

Would it make sense for `--allow-abort` to also be added as an arg to the `commit` command? How do folks feel about the argument/setting name and overall approach? See also #247 for a related feature proposal. Since the time of that discussion, I realized that naming a Commitizen argument `--allow-empty-message` could prove extremely confusing, because passing it to Commitizen would not, by itself, cause commits with empty messages to be committed like the Git argument of the same name. It would also interact poorly with the proposal to allow ferrying arbitrary arguments to `git commit` in #451, because it would be easy to accidentally pass the argument to Git, resulting in commits with empty messages no longer being aborted.


Woile:
LGTM but `cz check` is not my speciality, can u take a look @Lee-W ?

commitizen/cli.py
```diff
@@ -240,6 +240,12 @@
                         "help": "commit message that needs to be checked",
                         "exclusive_group": "group1",
                     },
+                    {
+                        "name": ["--reject-abort"],
+                        "action": "store_true",
+                        "default": False,
```
Lee-W, line 246: If I'm not misunderstood, the original behavior is `True`, right? If that's the case, I'd prefer to use that as the default option, unless we're to bump our major version

docs/check.md
```diff
@@ -46,3 +50,12 @@ $ cz check --commit-msg-file COMMIT_MSG_FILE
 
 In this option, COMMIT_MSG_FILE is the path of the temporal file that contains the commit message.
 This argument can be useful when cooperating with git hook, please check [Automatically check message before commit](auto_check.md) for more information about how to use this argument with git hook.
+
+### Reject Abort
+
```
Lee-W, line 55: nitpick: maybe we could remove this additional blank line

commitizen/commands/check.py
```diff
@@ -79,15 +83,15 @@ def __call__(self):
 
     def _get_commits(self):
         # Get commit message from file (--commit-msg-file)
-        if self.commit_msg_file:
+        if self.commit_msg_file is not None:
```
Lee-W, line 86: Wondering why you prefer to use `is not None` in this case?

commitizen/commands/check.py
```diff
@@ -79,15 +83,15 @@ def __call__(self):
...
             with open(self.commit_msg_file, "r", encoding="utf-8") as commit_file:
                 msg = commit_file.read()
             msg = self._filter_comments(msg)
             msg = msg.lstrip("\n")
             commit_title = msg.split("\n")[0]
             commit_body = "\n".join(msg.split("\n")[1:])
             return [git.GitCommit(rev="", title=commit_title, body=commit_body)]
-        elif self.commit_msg:
+        elif self.commit_msg is not None:
```
Lee-W, line 94: Same here

Lee-W (commented):
Hi @Kurt-von-Laven , sorry for late reviewing. Your work is fantastic! Just left a few questions. But I think we're almost ready to mrege

Kurt-von-Laven (commented):
Would it make sense for `--allow-abort` to also be added as an arg to the commit command?

EDIT: I think I answered this question for myself: no.

Kurt-von-Laven:
I am not able to reproduce the failure on Python 3.10 locally on Python 3.10.4. It appears to be unrelated to this pull request, so I think the build may simply need to be re-run. Please let me know if anyone sees a connection I am missing.

commitizen/cli.py
Lee-W: Sounds good 👍

commitizen/commands/check.py
Lee-W: Got it 👍

commitizen/commands/check.py
Lee-W: In that case, could you please add a comment here. I think we might miss it in the future

Lee-W (approved):
Except for the minor comment update, this PR looks good to me. @Woile I'm planning on merging it this weekend if no other new comments added

Lee-W:
hmmm. we might need to take a deeper look at the 3.10 case

commitizen/commands/check.py
Kurt-von-Laven: Good idea; done. Also, removing `is not None` on this line (but not the one above) will currently cause tests to fail.

Kurt-von-Laven (commented):
I am still unable to reproduce the build failure locally on Python 3.10.4.

Kurt-von-Laven:
Looks like the Python 3.10 build indeed simply needed to be re-run since all I changed was the two comments.

Lee-W (approved):
LGTM. Let's merge it now!

PR #514, opened 2022-05-16
fix(pre-commit): Correct pre-commit Hook Definition

Kurt-von-Laven:
## Description

Use new `--allow-abort` option. Prevent pre-commit hook from complaining when a commit is aborted by default, but allow users to override this option in their pre-commit config by specifying it in `args`, not `entry`. Move `--commit-msg-file` option from `entry` to `args` since it has to be the last option.

Confine hook to `commit-msg` stage. The Commitizen pre-commit hook runs `cz check`, which only makes sense to do on a commit message, not at other Git hook stages. By default the hook is currently run on staged files `pre-commit`, which is nonsensical unless the files in the repository themselves contain Git commit messages. The hook is an implausible candidate even for the `prepare-commit-msg` stage since this stage is intended for hooks that modify the commit message rather than check it. Specifying the hook stage centrally makes it less error-prone to use and saves one line of configuration for users of the pre-commit hook.

Change minimum pre-commit version from 0.15.4 to 1.4.3. pre-commit 1.4.3, released 2018-01-02, is the minimum pre-commit version at which `language_version: python3` is translated to the correct py launcher call on Windows.

Don't require serial execution. `require_serial` is not pertinent to `commit-msg` hooks or hooks that themselves execute serially. The option is intended to prevent fork-bombing that can occur when pre-commit runs a hook that itself spawns many processes many times in parallel.

## Checklist

## Expected behavior

The commit hook continues to work as before but without requiring that the `commit-msg` stage be specified in your `.pre-commit-config.yaml` or giving a misleading error message when the commit is aborted by saving an empty commit message in your editor.

1. Replace your Commitizen hook config in your `.pre-commit-config.yaml` with the following:

      ```yaml
      repos:
        - repo: https://github.com/Kurt-von-Laven/commitizen
          rev: allow-abort
          hooks:
            - id: commitizen
      ```

1. Ensure that you have pre-commit 1.4.3 or higher installed.
1. Install `commit-msg` hooks if you haven't already via: `pre-commit install --hook-type commit-msg`.
1. Commit some changes to your repository.
1. If the commit message was invalid, the hook will fail, and otherwise it will succeed. Empty commit messages now fall in the latter category, and the commit aborts cleanly without a misleading error.

## Additional context

Follows on #512, which added the `--allow-abort` option to `cz check`. Reverts #135, which removed `stages: [commit-msg]`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,4 +4,4 @@
   entry: cz check --commit-msg-file
   language: python
   language_version: python3
-  minimum_pre_commit_version: "0.15.4"
+  minimum_pre_commit_version: "1.4.3"
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,8 @@
 - id: commitizen
   name: commitizen check
   description: "Check whether the commit message follows commiting rules"
-  entry: cz check --commit-msg-file
+  entry: cz check
+  args: [--allow-abort, --commit-msg-file]
   stages: [commit-msg]
   language: python
   language_version: python3
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,6 +2,7 @@
   name: commitizen check
   description: "Check whether the commit message follows commiting rules"
   entry: cz check --commit-msg-file
+  stages: [commit-msg]
   language: python
   language_version: python3
   minimum_pre_commit_version: "1.4.3"
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,5 +4,4 @@
   entry: cz check --commit-msg-file
   language: python
   language_version: python3
-  require_serial: true
   minimum_pre_commit_version: "0.15.4"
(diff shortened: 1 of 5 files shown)


Kurt-von-Laven:
Please let me know what action I should take regarding the Codecov report. My instinct is to consider it a false positive and ignore it since I believe it is a consequence of increasing the length of the pre-commit hook definition length by one net line. Although, maybe the ideal thing would be to configure Codecov not to scan `.pre-commit-hooks.yaml`?

Lee-W:
I just browse through the code. Everything looks good, but I might a bit more time to test it out. I think we'll be able to merge it this week. Thank you so much for your active contributions 🙏

> Please let me know what action I should take regarding the Codecov report. My instinct is to consider it a false positive and ignore it since I believe it is a consequence of increasing the length of the pre-commit hook definition length by one net line. Although, maybe the ideal thing would be to configure Codecov not to scan `.pre-commit-hooks.yaml`?

I'm good with ignore the warning here.
@Woile I don't think I have the permission to do the check. Could you please check it when you have time? Thanks!

Lee-W:
just tested. let's merge it 🎉

PR #517, opened 2022-05-22
feat(pre-commit): Add commitizen-branch hook

Kurt-von-Laven:
## Description

Check all commit messages on the current branch. This is useful for checking commit messages after the fact (e.g., pre-push or in CI) since the existing hook only works at commit time. Expand the documentation of the pre-existing commitizen hook to clarify the relationship between them. I wasn't sure whether it would be appropriate to add a section to [this documentation](https://commitizen-tools.github.io/commitizen/auto_check/).

Remove no longer needed `commit-msg` stage from self-use of `commitizen` pre-commit hook in `.pre-commit-config.yaml`. In v2.27.1, the `commitizen` pre-commit hook began specifying that it runs on the `commit-msg` stage in the hook definition in `.pre-commit-hooks.yaml`, so it is no longer necessary to specify the hook stage when using the hook in `.pre-commit-config.yaml`.

## Checklist

## Expected behavior

The `commitizen` hook continues to work as before, allowing empty commit messages. The `commitizen-branch` hook checks all commit messages on your branch, and doesn't allow empty commit messages.

1. Replace your Commitizen hook config in your `.pre-commit-config.yaml` with the following:

      ```yaml
      repos:
        - repo: https://github.com/Kurt-von-Laven/commitizen
          rev: pre-commit-hook
          hooks:
            - id: commitizen
            - id: commitizen-branch
              stages: [push]
      ```

1. Ensure that you have pre-commit 1.4.3 or higher installed.
1. Install `commit-msg` and `pre-push` hooks if you haven't already via: `pre-commit install --hook-type commit-msg pre-push`.
1. Commit some changes to your repository.
1. If the commit message was invalid, the `commitizen` hook will fail, and otherwise it will succeed.
1. Commit a change with an empty commit message: `git commit --allow-empty-message -m ""`.
1. Attempt to push your branch.
1. The push fails with an error regarding the empty commit message.

## Additional context
Follows on #512 and #514.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -101,7 +101,7 @@ cz commit -s
 ```
 
 ### Integrating with Pre-commit
-Commitizen can lint your commit message for you with `cz check`.
+Commitizen can lint your commit messages for you with `cz check`.
 You can integrate this in your [pre-commit](https://pre-commit.com/) config with:
 
 ```yaml
@@ -111,13 +111,22 @@ repos:
     rev: master
     hooks:
       - id: commitizen
+      - id: commitizen-branch
+        stages: [push]
 ```
 
-After the configuration is added, you'll need to run
+After the configuration is added, you'll need to run:
 
 ```sh
-pre-commit install --hook-type commit-msg
+pre-commit install --hook-type commit-msg pre-push
 ```
+
+If you aren't using both hooks, you needn't install both stages.
+
+| Hook              | Recommended Stage |
+| ----------------- | ----------------- |
+| commitizen        | commit-msg        |
+| commitizen-branch | pre-push          |
 
 Read more about the `check` command [here](check.md).
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,6 @@ repos:
     rev: v2.31.0 # automatically updated by Commitizen
     hooks:
       - id: commitizen
-        stages: [commit-msg]
 
   - repo: local
     hooks:
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,9 +1,27 @@
 - id: commitizen
   name: commitizen check
-  description: "Check whether the commit message follows commiting rules"
+  description: >
+    Check whether the current commit message follows commiting rules. Allow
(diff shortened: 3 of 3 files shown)


docs/README.md
```diff
@@ -111,12 +111,14 @@ repos:
...
       - id: commitizen
+      - id: commitizen-branch
+        stages: [push]
 ```
 
 After the configuration is added, you'll need to run
 
 ```sh
-pre-commit install --hook-type commit-msg
+pre-commit install --hook-type commit-msg pre-push
```
Lee-W, line 121: Should we separate the steps of adding these hooks? e.g., `commitizen` hook might need need `pre-push`

Lee-W (commented):
@Woile Overall I'm good with this feature. One of the things come to my mind is whether we really need 2 `cz check` hooks? It seems users can config such behavior on their `.pre-commit-config.yaml` as well

docs/README.md
Kurt-von-Laven: I clarified the relationship between the hooks and the stages by adding a table. Let me know what you think.

Kurt-von-Laven:
Users can certainly write their own pre-commit hooks, but I consider them tricky to write and test since many implementations I see that haven't already been reviewed by the author of pre-commit get details wrong. I also believe many users will not be aware that the commit-msg hook relies on the developer to have installed the pre-commit hooks correctly since it can't be run in CI. It is a good hook in that it offers real-time feedback, but it is also trivially circumvented (most likely by accident). Tools like [pre-commit.ci](https://pre-commit.ci/), [pre-commit/pre-commit-action](https://github.com/pre-commit/pre-commit-action), and [ScribeMD/pre-commit-action](https://github.com/ScribeMD/pre-commit-action) make it trivial to run the proposed pre-push hook in CI, which is not possible with the existing commit-msg hook.

Kurt-von-Laven:
The test failures seem unrelated to this pull request.

docs/README.md
Lee-W: It's great! Thanks @Kurt-von-Laven

Lee-W (approved):
@Woile I'm good with this one. Could you please take a look? Thanks! As for the failed tests, I'll send a separate PR to fix it.

PR #522, opened 2022-06-09
Sign tag

gpongelli:
## Description
Add tag sign feature for bump command.

## Checklist

## Expected behavior

When added the `--signed-tag`, bump command will sign the tag with preconfigured GPG private key.

1. run `cz bump --signed-tag <tag>`
2. check with `git tag -v <tag>` that the stdout answer starts with `gpg: Signature made` string

A new method to check if the tag is signed was added to do same step of point 2 above.


gpongelli:
pipeline error seems nothing related to my changes, please let me know how to retry the jobs or how to fix them.

gpongelli:
pipeline fixed, @Woile @Lee-W please review this PR, thanks.

tests/commands/test_bump_command.py
```diff
@@ -66,6 +66,24 @@ def test_bump_minor_increment_annotated(commit_msg, mocker):
...
 
+    _is_signed = git.is_signed_tag("0.2.0")
+    assert _is_signed is False
+
+
+@pytest.mark.parametrize("commit_msg", ("feat: new file", "feat(user): new file"))
+@pytest.mark.usefixtures("tmp_commitizen_project_with_gpg")
+def test_bump_minor_increment_signed(commit_msg, mocker):
+    create_file_and_commit(commit_msg)
+    testargs = ["cz", "bump", "--yes", "--signed-tag"]
```
Woile, line 77: Should we use something like in git commit instead? I think they use `-s` and some others, @Lee-W ?

commitizen/cmd.py
```diff
@@ -20,4 +20,10 @@ def run(cmd: str) -> Command:
     )
     stdout, stderr = process.communicate()
     return_code = process.returncode
-    return Command(stdout.decode(), stderr.decode(), stdout, stderr, return_code)
+    return Command(
+        stdout.decode("iso-8859-1"),
```
Lee-W, line 24: Could you elaborate why do we need to decode `iso-8859-1`? Thanks!

tests/commands/test_bump_command.py
Lee-W: @Woile I'm not sure I get your point. But using `cz bump ` to test here should be fine

Woile: I was asking about the `flag` name, should we use something like `git commit -s`? In the current implementation is using `--signed-tag` or `-st`

gpongelli: I put `--signed-tag` and `-st` to mimic what's done for annotated tag.

No problem to change to `-s` like git for the short name, let me know.

Is long name `--signed-tag` ok?

Woile: Makes sense, I'm not really sure what would be correct. 

For example, do you think it would be possible to sign an anottated tag and a lightweight tag? Could a user do:

```
cz bump --anotated-tag -s
```
and also
```
cz bump -s
```

commitizen/cmd.py
gpongelli: I changed to iso-8859-1 because my gpg's stdout returns latin characters that fails to be decoded.

The character I get into stdout is "à" .

Probably the best way could be doing multiple decode stages in case of exception, like "first decode with utf, if it fails decode with iso-8859, if it fails decode with... " , but it is a different task/issue than this PR.

tests/commands/test_bump_command.py
gpongelli: A tag can be lightweight, or annotated, or signed.

This is why into tag method the `_opt` is not appended but substituted (cannot link the row, so paste here):
```python
    if annotated:
         _opt = f"-a {tag} -m"
     if signed:
         _opt = f"-s {tag} -m"
```

I think git cannot accept both arguments, but I'm not sure.

gpongelli: looking at git documentation, there are:
- `-s` and `--sign` for [git tag](https://git-scm.com/docs/git-tag#Documentation/git-tag.txt--s)
- `-S` `--gpg-sign` for [git commit](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt)

If commitizen should use same git options' name, I think also commit command should be aligned.

Let me know if you agree and I'll align both commands.

Woile: yes, i agree

Woile:
Awesome work @gpongelli 💪🏻

commitizen/cli.py
```diff
@@ -155,10 +155,16 @@
                         "action": "store_true",
                     },
                     {
-                        "name": ["--annotated-tag", "-at"],
+                        "name": ["--annotate", "-a"],
```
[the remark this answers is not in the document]
gpongelli: latest commit reverts the breaking change.

Woile (approved):
@Lee-W looks good to me, if you are ok with it we merge

gpongelli:
@Woile what do you think about [this comment](https://github.com/commitizen-tools/commitizen/pull/522#discussion_r913551098) ?

commitizen/cmd.py
gpongelli: on windows I've just checked, it depends on the shell:

with Powershell I get nothing
image

Meanwhile with MINGW I get something
image

this is why I currently use chardet when I've to decode something, and I'm using it on windows and on linux.

gpongelli: latest commits group add chardet and its stub for mypy .

Lee-W: I'm good with using chardet. Looks like a good idea.

commitizen/cmd.py
```diff
@@ -20,4 +22,10 @@ def run(cmd: str) -> Command:
     )
     stdout, stderr = process.communicate()
     return_code = process.returncode
-    return Command(stdout.decode(), stderr.decode(), stdout, stderr, return_code)
+    return Command(
+        stdout.decode(chardet.detect(stdout)["encoding"] or "utf-8"),
```
Lee-W, line 26: What's the expected output of `chardet.detect` if it cannot detect anything? Will it be something like `{'encoding': None, 'confidence': 0.99}`?

scripts/test
```diff
@@ -9,5 +9,5 @@ ${PREFIX}pytest -n 3 --cov-report term-missing --cov-report=xml:coverage.xml --c
 ${PREFIX}black commitizen tests --check
 ${PREFIX}isort --check-only commitizen tests
 ${PREFIX}flake8 commitizen/ tests/
-${PREFIX}mypy commitizen/ tests/
+${PREFIX}mypy --install-types --non-interactive commitizen/ tests/
```
Lee-W, line 12: could you describe a bit more on why these arguments are added? Thanks!

gpongelli: it is to avoid [such kind of pipeline errors](https://github.com/commitizen-tools/commitizen/runs/7410357144?check_suite_focus=true#step:6:86) because of chardet's missing stubs for mypy.

Lee-W: if that's the case, should we add types-chardet directly to our dev dependency in `pyproject.toml`? Or what would be the main difference between adding `type-chardet` and using `--install-types`

commitizen/cmd.py
gpongelli: > What's the expected output of `chardet.detect` if it cannot detect anything? Will it be something like `{'encoding': None, 'confidence': 0.99}`?

Yes, it returns None, that's why I've added the `or` clause.

It also returns None for empty content `b''`, like the stderr content when all goes ok, so the `or` is mandatory to avoid `decode` failures.

Let me know if this "chardet-enabled" version is ok or if you prefer the previous one (with "iso..." encoding).

scripts/test
gpongelli: you're right, I did some test and locally mypy wanted that command instead of managing all through pyproject .

I'll do a commit to see if pipeline works having package into pyproject, if not I'll revert it.

gpongelli: `types-chardet` into pyproject works

Lee-W: nice! then, let's use `types-chardet`

commitizen/cmd.py
Lee-W: just confirm it. I like the chatdet idea.

Lee-W (approved):
Thanks for your contribution! let's merge it and release a new version 🎉🎉🎉

PR #540, opened 2022-08-01
git.get_commits error check

### LINKED ISSUE #518: `cz changelog` is not reporting git errors (`git.py` missing `cmd.return_code` checks)
### Description

`cz commit` worked just fine, but `cz changelog` resulted in the error:
> "No commits found"

I later came to find the calls to git were silently failing in `git.get_tags` due to an unsupported version of git.
It's not very obvious the issue is with git.
Might we want some kind of warnings/exceptions [here](https://github.com/commitizen-tools/commitizen/blob/afa0d93544405bb0a871b5072e0a814720d933be/commitizen/git.py#L123)"?

### Steps to reproduce

1. unsupported git (e.g. 1.8.3.1) (normally we use a newer version)
2. `cz changelog` (`cz commit` worked)

### Current behavior

Error handling could be improved.

### Desired behavior

Maybe a warning or `raise` would help users out when `cmd.run( ...git...)` fails?

### Screenshots

_No response_

### Environment

CentOS7

bhelgs:
## Description
`git.get_commits` could potentially fail silently. 

The first commit is the fix. 
See the git commit message for more info on all following commits (broken up for easier review). 
image

The final commit also improves error checking in `git.get_tags`  (closes #518).

## Checklist

## Expected behavior
Fail fast behavior and better reports to the user when something goes wrong with git.

Some of the tests showed the command was failing for unexpected reasons (e.g. invalid commit range `git log None..None`).

#### Example of a silent failure that is now caught:
Without this change users with really old versions of git will see is "No commits found" but after the change: <b>
`GitCommandError: fatal: unrecognized argument: --author-date-order`
* reason:  `commands/changelog.py` calls `git.get_commits` with `--author-date-order`  but this argument is not valid for older versions of it. 

1. ran tests locally and formatted.

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -105,27 +105,12 @@ def get_commits(
     start: Optional[str] = None,
     end: str = "HEAD",
     *,
-    log_format: str = "%H%n%s%n%an%n%ae%n%b",
-    delimiter: str = "----------commit-delimiter----------",
     args: str = "",
 ) -> List[GitCommit]:
     """Get the commits between start and end."""
-    git_log_cmd = (
-        f"git -c log.showSignature=False log --pretty={log_format}{delimiter} {args}"
-    )
-
-    if start:
-        command = f"{git_log_cmd} {start}..{end}"
-    else:
-        command = f"{git_log_cmd} {end}"
-    c = cmd.run(command)
-    if c.return_code != 0:
-        raise GitCommandError(c.err)
-    if not c.out:
-        return []
-
+    git_log_entries = _get_log_as_str_list(start, end, args)
     git_commits = []
-    for rev_and_commit in c.out.split(f"{delimiter}\n"):
+    for rev_and_commit in git_log_entries:
         if not rev_and_commit:
             continue
         rev, title, author, author_email, *body_list = rev_and_commit.split("\n")
@@ -236,3 +221,22 @@ def get_eol_style() -> EOLTypes:
 def smart_open(*args, **kargs):
     """Open a file with the EOL style determined from Git."""
     return open(*args, newline=get_eol_style().get_eol_for_open(), **kargs)
+
+
+def _get_log_as_str_list(start: Optional[str], end: str, args: str) -> List[str]:
+    """Get string representation of each log entry"""
+    delimiter = "----------commit-delimiter----------"
+    log_format: str = "%H%n%s%n%an%n%ae%n%b"
+    git_log_cmd = (
+        f"git -c log.showSignature=False log --pretty={log_format}{delimiter} {args}"
+    )
+    if start:
+        command = f"{git_log_cmd} {start}..{end}"
+    else:
+        command = f"{git_log_cmd} {end}"
+    c = cmd.run(command)
+    if c.return_code != 0:
+        raise GitCommandError(c.err)
+    if not c.out:
+        return []
+    return c.out.split(f"{delimiter}\n")
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import List, Optional
 
-from commitizen import cmd
+from commitizen import cmd, out
 from commitizen.exceptions import GitCommandError
 
 UNIX_EOL = "\n"
@@ -150,8 +150,13 @@ def get_tags(dateformat: str = "%Y-%m-%d") -> List[GitTag]:
         f'%(object)"'
     )
     c = cmd.run(f"git tag --format={formatter} --sort=-creatordate")
+    if c.return_code != 0:
+        raise GitCommandError(c.err)
 
-    if c.err or not c.out:
+    if c.err:
+        out.warn(f"Attempting to proceed after: {c.err}")
+
+    if not c.out:
         return []
 
     git_tags = [
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -6,6 +6,7 @@ from tempfile import NamedTemporaryFile
 from typing import List, Optional
 
 from commitizen import cmd
+from commitizen.exceptions import GitCommandError
 
 UNIX_EOL = "\n"
 WINDOWS_EOL = "\r\n"
@@ -118,6 +119,8 @@ def get_commits(
     else:
         command = f"{git_log_cmd} {end}"
     c = cmd.run(command)
+    if c.return_code != 0:
+        raise GitCommandError(c.err)
     if not c.out:
         return []
 
--- a/docs/exit_codes.md
+++ b/docs/exit_codes.md
@@ -30,3 +30,4 @@ These exit codes can be found in `commitizen/exceptions.py::ExitCode`.
 | NotAllowed                  | 20        | `--incremental` cannot be combined with a `rev_range`                                                       |
 | NoneIncrementExit           | 21        | The commits found are not eligible to be bumped                                                             |
 | CharacterSetDecodeError     | 22        | The character encoding of the command output could not be determined                                        |
+| GitCommandError             | 23       | Unexpected failure while calling a git command                                                              |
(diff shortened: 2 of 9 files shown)


docs/exit_codes.md
```diff
@@ -5,7 +5,7 @@ Commitizen handles expected exceptions through `CommitizenException` and returns
 These exit codes can be found in `commitizen/exceptions.py::ExitCode`.
 
 | Exception                   | Exit Code | Description                                                                                                 |
-| --------------------------- | --------- | ----------------------------------------------------------------------------------------------------------- |
```
bhelgs, line 8: I wonder why my formatter did this. Not sure if this is better.

tests/commands/test_changelog_command.py
```diff
@@ -20,10 +21,12 @@ def test_changelog_on_empty_project(mocker):
...
     mocker.patch.object(sys, "argv", testargs)
 
-    with pytest.raises(NoCommitsFoundError) as excinfo:
+    with pytest.raises(GitCommandError):
         cli.main()
 
-    assert "No commits found" in str(excinfo)
+    #  git will error out with something like:
+    #  "your current branch 'XYZ' does not have any commits yet"
+    #  is it wise to assert the exception contains a message like this?
```
Lee-W, line 29: If we need to compare the error message of external tool (e.g., `git`), we'll need to take `LC_ALL` into consideration. Otherwise, people with difference `LC_ALL` will not pass the test

Lee-W: We could try something like changing the git commit command to `LC_ALL=en_US.UTF-8 git commit`

bhelgs: you bring up a really good point. Are you happy as-is? How would you like to proceed with this PR?

Lee-W: I notice you join the discussion [here](https://github.com/commitizen-tools/commitizen/pull/543#issuecomment-1203038161). Maybe we could discuss this kind of issue there 🤔 

> How would you like to proceed with this PR?

Yep, I'm ok with asserting the message. 

> How would you like to proceed with this PR?

We'll need a way to resolve the language issue. After that, I think we're almost ready to merge

bhelgs: What about this approach of retrying the command?

1. `execinfo` captures the command that failed.
2. test retries executing the command with `LC_ALL`.
3. assert the retry matches the expected `en_US` output of the command.

``` python
    with pytest.raises(GitCommandError) as execinfo:
         cli.main()

     c = cmd.run("LC_ALL=en_US.UTF-8 " + execinfo.value.cmd)
     assert "does not have any commits yet" in c.err
```

tests/commands/test_changelog_command.py
```diff
@@ -676,12 +688,13 @@ def test_changelog_from_rev_single_version_not_found(
     with pytest.raises(NoCommitsFoundError) as excinfo:
         cli.main()
 
-    assert "No commits found" in str(excinfo)
+    assert "Could not find a valid revision" in str(excinfo)
```
Lee-W, line 691: Do we need to use the approach above as well?

tests/commands/test_changelog_command.py
Lee-W: I'm good with this approach :) 

@Woile what do you think. Different from what's discussed on #543  `LC_ALL` part only happens in test cases.

tests/commands/test_changelog_command.py
bhelgs: In this case `commitizen/changelog.py` is what raised the exception message (in English) so no need to have `LC_ALL`:

image

tests/commands/test_changelog_command.py
bhelgs: I wonder if I should have just added `LC_ALL=en_US.UTF-8` to the entire testing environment.

bhelgs: that would be quite a bit more straightforward than this retry strategy.

bhelgs: assuming we don't mock out that specific git call entirely.

Lee-W: If it's just for testing, we could use [pytest-env](https://pypi.org/project/pytest-env/) to achieve it. If it's for the functional code, maybe we'll need to think of a better solution like we did on #543

tests/commands/test_changelog_command.py
Lee-W: ah got it. Thanks @bhelgs !

tests/commands/test_changelog_command.py
bhelgs: @Lee-W After much thought, this test probably isn't providing much on the `command` level.
I think all we want to be concerned about is that we handle an empty git log properly. 

I have refactored `get_commits` so that it calls a `get_log` function that we can test as it's own unit. https://github.com/commitizen-tools/commitizen/pull/540/commits/4317d9f3aba9cb5edba702f0076155118578fad1

bhelgs: As in:

**tests/test_git.py**
``` python
 def test_get_log_as_str_list_empty():
     """ensure an exception or empty list in an empty project"""
     try:
         gitlog = git._get_log_as_str_list(start=None, end="HEAD", args="")
     except exceptions.GitCommandError:
         return
     assert len(gitlog) == 0, "list should be empty if no assert"
```

replaces: <br>

**tests/commands/test_changelog_command.py**
``` python
 def test_changelog_on_empty_project(mocker):
     testargs = ["cz", "changelog", "--dry-run"]

     with pytest.raises(GitCommandError):
         cli.main()

     #  git will error out with something like:
     #  "your current branch 'XYZ' does not have any commits yet"
     #  is it wise to assert the exception contains a message like this?
```

bhelgs: And then no need for the `LC_ALL` at all since the function being tested is so small in comparison to before when we tested the whole command

tests/commands/test_changelog_command.py
```diff
@@ -15,17 +15,6 @@
 from tests.utils import create_file_and_commit, wait_for_tag
 
 
-@pytest.mark.usefixtures("tmp_commitizen_project")
```
Lee-W, line 18: May i know why is this test removed?

Lee-W (commented):
Thanks @bhelgs ! I'm good with most of the stuff. Just leave one question and we might need your help to rebase from the master branch. I think we're almost ready to merge 💪

tests/commands/test_changelog_command.py
bhelgs: This test would be a duplicate of `test_git.py::test_get_log_as_str_list_empty`. 
Also, it seems unnecessary for the changelog command/test to enforce a specific exception since it has no way of handling such an exception.

commitizen/git.py
```diff
@@ -105,27 +105,12 @@ def get_commits(
...
-        command = f"{git_log_cmd} {end}"
-    c = cmd.run(command)
-    if c.return_code != 0:
-        raise GitCommandError(c.err)
-    if not c.out:
-        return []
-
+    git_log_entries = _get_log_as_str_list(start, end, args)
     git_commits = []
-    for rev_and_commit in c.out.split(f"{delimiter}\n"):
```
bhelgs, line 125: I wonder if this `\n` was working properly for Windows.

tests/commands/test_changelog_command.py
Lee-W: ah got it. Thanks for your explanation!

Lee-W (approved):
@Woile I'm good with this one as well. Could you please take a look when you have time? Thanks!

commitizen/git.py
Lee-W: Our CI runs on Windows machine as well so I think it should be fine 🤔 But both maintainers do not have windows machine to test it out. Might need others' help to verify it.

Woile:
Thanks a lot!

Issue #542, opened 2022-08-02
failed to check signed commit for user uses different language


Lee-W:
### Description

as title

### Steps to reproduce

1. `export LC_ALL=zh_TW.UTF-8`
2. `./script/test`

The test will fail because https://github.com/commitizen-tools/commitizen/blob/master/commitizen/git.py#L148 checks only the English version output

### Current behavior

`is_signed_tag` function checks only works for ones whose system language is English.

### Desired behavior

`is_signed_tag` function should temporarily set the language to English for commtizen to check

### Environment

Commitizen Version: 2.29.2
Python Version: 3.8.5 (default, Oct  1 2020, 16:07:27)
[Clang 12.0.0 (clang-1200.0.32.2)]
Operating System: Darwin

PR #543, opened 2022-08-02
use error_code to do "is_signed_tag" check to avoid language issue

Lee-W:
## Description
as title

## Checklist

## Expected behavior
pass all the test cases even when `LC_ALL` is set to non-English

1. `export LC_ALL=zh_TW.UTF-8`
2. `./script/test`

## Additional context
#542

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -1,5 +1,4 @@
 import os
-import re
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import List, Optional
@@ -143,11 +142,7 @@ def tag_exist(tag: str) -> bool:
 
 
 def is_signed_tag(tag: str) -> bool:
-    c = cmd.run(f"git tag -v {tag}")
-    _ret = False
-    if re.match("gpg: Signature made [0-9/:A-Za-z ]*", c.err):
-        _ret = True
-    return _ret
+    return cmd.run(f"git tag -v {tag}").return_code == 0
 
 
 def get_latest_tag_name() -> Optional[str]:


Woile:
I wonder if there's a language agnostic way of doing this, instead of forcing english into the equation 🤔

bhelgs:
@Woile 

> I wonder if there's a language agnostic way of doing this, instead of forcing english into the equation 🤔

* `git tag -a testtag`
* `git tag -v testtag || echo $?`

The above return code is 1... so it appears the exit code is 1 for annotated tag and 0 for signed tag.

bhelgs:
side note, `tmp_commitizen_project_with_gpg` is creating a key every time the unit tests run. I wonder if there is a way to make them temporary.

Lee-W:
> side note, tmp_commitizen_project_with_gpg is creating a key every time the unit tests run. I wonder if there is a way to make them temporary.

Maybe we could use a context manager to remove the created key?

gpongelli:
> @Woile
> 
> > I wonder if there's a language agnostic way of doing this, instead of forcing english into the equation 🤔
> 
>     * `git tag -a testtag`
> 
>     * `git tag -v testtag || echo $?`
> 
> 
> The above exit code is 1... so it appears the exit code is 1 for annotated tag and 0 for signed tag.


[link](https://git-scm.com/docs/git-verify-tag) `git verify-tag <tagname>` (that also print things to stdout) and then `echo $?` :
* 0 means signed tag
* 1 means not signed

Lee-W:
Thanks @bhelgs and @gpongelli ! Let me update the code with the methods you provided!

Lee-W:
@Woile I think this one is now ready to be reviewed 🎉

gpongelli:
> > side note, tmp_commitizen_project_with_gpg is creating a key every time the unit tests run. I wonder if there is a way to make them temporary.
> 
> Maybe we could use a context manager to remove the created key after running testing?

PR #547  handles this ;-)

PR #545, opened 2022-08-03
fix(cmd): improve character encoding detection for sub-commands

### LINKED ISSUE #544: `UnicodeDecodeError` if commit messages contain Unicode characters
### Description

If I run
```
cz changelog
```
and the commit messages contain Unicode characters like 🤦🏻‍♂️ (which is an eight-byte utf-8 sequence: `\xf0\x9f\xa4\xa6 \xf0\x9f\x8f\xbb`) then I get the following traceback
```
  File "/.../.venv/bin/cz", line 8, in <module>
    sys.exit(main())
  File "/.../.venv/lib/python3.10/site-packages/commitizen/cli.py", line 389, in main
    args.func(conf, vars(args))()
  File "/.../.venv/lib/python3.10/site-packages/commitizen/commands/changelog.py", line 143, in __call__
    commits = git.get_commits(
  File "/.../.venv/lib/python3.10/site-packages/commitizen/git.py", line 98, in get_commits
    c = cmd.run(command)
  File "/.../.venv/lib/python3.10/site-packages/commitizen/cmd.py", line 32, in run
    stdout.decode(chardet.detect(stdout)["encoding"] or "utf-8"),
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/encodings/cp1254.py", line 15, in decode
    return codecs.charmap_decode(input,errors,decoding_table)
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 1689: character maps to <undefined>
```
The result of `chardet.detect()` here

https://github.com/commitizen-tools/commitizen/blob/2ff9f155435b487057ce5bd8e32e1ab02fd57c94/commitizen/cmd.py#L26

is:
```
{'encoding': 'Windows-1254', 'confidence': 0.6864215607255395, 'language': 'Turkish'}
```
An interesting character encoding prediction with a low confidence, which in turn picks the incorrect codec and then decoding the `bytes` fails. Using `decode("utf-8")` works fine. It looks like issue https://github.com/chardet/chardet/issues/148 is related to this.

I think the fix would be something like this to replace [these lines of code](https://github.com/commitizen-tools/commitizen/blob/2ff9f155435b487057ce5bd8e32e1ab02fd57c94/commitizen/cmd.py#L23-L31):
```python
stdout, stderr = process.communicate()
return_code = process.returncode
try:
    stdout_s = stdout.decode("utf-8")  # Try this one first.
except UnicodeDecodeError:
    result = chardet.detect(stdout)  # Final result of the UniversalDetector’s prediction.
    # Consider checking confidence value of the result?
    stdout_s = stdout.decode(result["encoding"])
try:
    stderr_s = stderr.decode("utf-8")  # Try this one first.
except UnicodeDecodeError:
    result = chardet.detect(stderr)  # Final result of the UniversalDetector’s prediction.
    # Consider checking confidence value of the result?
    stderr_s = stderr.decode(result["encoding"])
return Command(stdout_s, stderr_s, stdout, stderr, return_code)
```

### Steps to reproduce

Well I suppose you can add a few commits to a local branch an go crazy with much text and funky unicode characters (emojis with skin tones, flags, etc.), and then attempt to create a changelog.

### Current behavior

`cz` throws an exception.

### Desired behavior

`cz` creates a changelog.

### Screenshots

_No response_

### Environment

```
> cz version
2.29.3
> python --version
Python 3.10.5
> uname -a
Darwin pooh 18.7.0 Darwin Kernel Version 18.7.0: Mon Feb 10 21:08:45 PST 2020; root:xnu-4903.278.28~1/RELEASE_X86_64 x86_64 i386 Darwin
```

jenstroeger:
## Description

Fixes #544 

## Checklist

## Expected behavior

See issue #544: `cz` now creates the changelog without failing

### DIFF
--- a/commitizen/cmd.py
+++ b/commitizen/cmd.py
@@ -12,6 +12,14 @@ class Command(NamedTuple):
     return_code: int
 
 
+def _try_decode(bytes_: bytes) -> str:
+    try:
+        return bytes_.decode("utf-8")
+    except UnicodeDecodeError:
+        result = chardet.detect(bytes_)
+        return bytes_.decode(result["encoding"] or "utf-8")
+
+
 def run(cmd: str) -> Command:
     process = subprocess.Popen(
         cmd,
@@ -23,8 +31,8 @@ def run(cmd: str) -> Command:
     stdout, stderr = process.communicate()
     return_code = process.returncode
     return Command(
-        stdout.decode(chardet.detect(stdout)["encoding"] or "utf-8"),
-        stderr.decode(chardet.detect(stderr)["encoding"] or "utf-8"),
+        _try_decode(stdout),
+        _try_decode(stderr),
         stdout,
         stderr,
         return_code,


jenstroeger:
@Woile I’m unsure why that one test failed when locally it worked on `git push`. Unless something jumps out at you right now, I’ll look into it later.

commitizen/cmd.py
```diff
@@ -12,6 +12,14 @@ class Command(NamedTuple):
     return_code: int
 
 
+def _try_decode(bytes_):
+    try:
+        return bytes_.decode("utf-8")
+    except UnicodeDecodeError:
+        result = chardet.detect(bytes_)
+        return bytes_.decode(result["encoding"] or "utf-8")
```
gpongelli, line 20: Is the “or” part needed? Doing `decode(“utf-8”)` here could raise unhandled exception.

jenstroeger: Well, in case of
```
{'encoding': None, 'confidence': 0.0, 'language': None}
```
the `decode()` function should receive the `"utf-8"` encoding arg. Incidentally, that’s the default encoding too ([docs](https://docs.python.org/3/library/stdtypes.html#bytes.decode)) for all Python versions that commitizen supports ([3.6 and later](https://github.com/commitizen-tools/commitizen/blob/2ff9f155435b487057ce5bd8e32e1ab02fd57c94/pyproject.toml#L48)).

Leaving it there or removing it both work, I’m happy with either way.

commitizen/cmd.py
```diff
@@ -12,6 +12,14 @@ class Command(NamedTuple):
     return_code: int
 
 
+def _try_decode(bytes_):
```
Woile, line 15: could you add the type hint here? `bytes_: bytes`?

commitizen/cmd.py
Woile: Should we have a minimum threshold? If the confidence is low, what's the point? I would only return if it makes sense, something like:

```python
    try:
        return bytes_.decode("utf-8")
    except UnicodeDecodeError as e:
        result = chardet.detect(bytes_)
        if result["confidence"] > 0.8:
            return bytes_.decode(result["encoding"] or "utf-8")
        raise e
```

Thoughts?

jenstroeger: I agree with checking the confidence value, the interesting part will be to pick the threshold here 😉

@Woile do you want to reraise the original [UnicodeDecodeError](https://docs.python.org/3/library/exceptions.html#UnicodeDecodeError) if the confidence is too low, or define a new [custom exception](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/exceptions.py) (e.g. `CharacterSetDecodeError` which carries the misidentified charset)?

There’s also the case where a high confidence can still  fail to decode, which we should consider.

commitizen/cmd.py
jenstroeger: Yup, I didn’t know that the code is required to have type hints 👍🏼

jenstroeger: @Woile I amended the initial commit.

What’s the general recommendation for this repo — 
- any number of PR commits and you’ll squash on merge, or 
- careful/useful commits and you’ll merge the PR with its history, or
- keep amending the initial commit?

commitizen/cmd.py
gpongelli: > Well, in case of
> 
> ```
> {'encoding': None, 'confidence': 0.0, 'language': None}
> ```
> 
> the `decode()` function should receive the `"utf-8"` encoding arg. 

Since I’m using chardet, I’ve never received such dict as answer, but ok to keep that. 


> > Doing `decode(“utf-8”)` here could raise unhandled exception.
> 
> And any other encoding could also raise an exception, too. Which is why @Woile [mentioned](https://github.com/commitizen-tools/commitizen/issues/544#issuecomment-1203810504) to propagate an error and I agree with that.

I did not see that issue, I agree with that comment.

commitizen/cmd.py
Woile: A combination of the last 2. Think which commits you want to be reflected in the changelog. As of now is good 👍🏻

commitizen/cmd.py
jenstroeger: > Since I’m using chardet, I’ve never received such dict as answer, but ok to keep that.

Easy to enjoy:
```python
>>> import chardet
>>> chardet.detect("ö".encode())
{'encoding': 'TIS-620', 'confidence': 0.99, 'language': 'Thai'}
>>> chardet.detect("🤦🏻‍♂️".encode())
{'encoding': None, 'confidence': 0.0, 'language': None}
```

> I did not see that issue, I agree with that comment.

Thanks. I’ll change the code accordingly.

Woile: @jenstroeger I like the idea of a custom exception. And you are right that it can still fail. So maybe we can skip the threshold directly and return the failure:

```python
try:
     bytes_.decode("utf-8")
except UnicodeDecodeError as e:
    result = chardet.detect(bytes_)
    try:
        return bytes_.decode(result["encoding"])  # here we don't need to try utf-8 again
    except UnicodeDecodeError as e:
        raise CharacterSetDecodeError("Could not decode this char", e)
```

gpongelli: Looking for chardet alternatives,  [charset_normalizer](https://github.com/Ousret/charset_normalizer) seems promising (py versions, supported languages and also emoji), I’ll try it doing some checks in next future.


I cannot try it at the moment, but if you can try the same strings that led to your errors and if it works correctly, I support its use instead of chardet.

Woile: @jenstroeger do you have some time to update the code? otherwise let me know and I can take over 👍🏻

jenstroeger: I’ll do that tonight or tomorrow, it’s been busy today…

gpongelli: I did [this PR](https://github.com/commitizen-tools/commitizen/pull/550) using [charset_normalizer](https://github.com/Ousret/charset_normalizer) instead of chardet , can you try that branch and see if the issue still happens ?

PR #546, opened 2022-08-03
feat(ConventionalCommitsCz): allow to override defaults from config

### LINKED ISSUE #535: generate defaults for cz_customize pyproject.toml
### Description

I want to add a type but, in order to do so, I need to define the entire commitizen sections of my pyproject.toml as outlined here: https://commitizen-tools.github.io/commitizen/customization/.

It would be much easier to have the option to start with the defaults filled in (e.g. default questions, default message_template,etc...) and then modify from there. As it is now, I am copying values out of the ConventionalCommitsCz

### Possible Solution

_No response_

### Additional context

_No response_

### Additional context

_No response_

amitlevy21:
Before fixing tests and adding extra ones, I would like to receive
partial feedback about my changes to see if this aligns with the desired
code changes by maintainers.

As addressed on #535, when using customize commitizen, if we want to
customize a small attribute, we need to redefine all commitizens options
to make our custom class work.

For example:

```diff
diff --git a/cz.yaml b/cz.yaml
index f2e19a9..302e961 100644
--- a/cz.yaml
+++ b/cz.yaml
@@ -1,6 +1,18 @@
 commitizen:
   annotated_tag: true
   bump_message: 'bump: $current_version -> $new_version [skip ci]'
-  name: cz_conventional_commits
+  name: cz_customize
   update_changelog_on_bump: true
   version: 0.11.0
+
+  customize:
+    bump_pattern: '^(fix|feat|docs|style|refactor|test|build|ci)'
+    bump_map:
+      fix: PATCH
+      feat: PATCH
+      docs: PATCH
+      style: PATCH
+      refactor: PATCH
+      test: PATCH
+      build: PATCH
+      ci: PATCH
diff --git a/t b/t
new file mode 100644
index 0000000..e69de29
diff --git a/t2 b/t2
new file mode 100644
index 0000000..e69de29

```

making the following change on a repo would cause an unexpected
behavior:

```python
+ bash -c cz commit

  File "/home/amit/.local/bin/cz", line 8, in <module>
    sys.exit(main())
  File "/home/amit/.local/lib/python3.10/site-packages/commitizen/cli.py", line 382, in main
    args.func(conf, vars(args))()
  File "/home/amit/.local/lib/python3.10/site-packages/commitizen/commands/commit.py", line 74, in __call__
    m = self.prompt_commit_questions()
  File "/home/amit/.local/lib/python3.10/site-packages/commitizen/commands/commit.py", line 49, in prompt_commit_questions
    for question in filter(lambda q: q["type"] == "list", questions):
  File "/home/amit/.local/lib/python3.10/site-packages/commitizen/commands/commit.py", line 49, in <lambda>
    for question in filter(lambda q: q["type"] == "list", questions):
KeyError: 'type'
```

From my best understanding, this error happens because I didn't defined
question section in config, though I'm ok with using
ConventionalCommitsCz default ones.

This commit extends ConventionalCommitsCz to read from config and
fallbacks to defaults if some are not provided.

By adding this change, potentially customize commitizen can be
deprecated.

A warning message will be printed when using `cz_customize`

Closes #535.

## Checklist

### DIFF
--- a/commitizen/cz/base.py
+++ b/commitizen/cz/base.py
@@ -1,37 +1,14 @@
 from abc import ABCMeta, abstractmethod
-from typing import Callable, Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional
 
-from prompt_toolkit.styles import Style, merge_styles
+from prompt_toolkit.styles import Style
 
-from commitizen import git
+from commitizen import defaults, git
 from commitizen.config.base_config import BaseConfig
 from commitizen.defaults import Questions
 
 
 class BaseCommitizen(metaclass=ABCMeta):
-    bump_pattern: Optional[str] = None
-    bump_map: Optional[Dict[str, str]] = None
-    default_style_config: List[Tuple[str, str]] = [
-        ("qmark", "fg:#ff9d00 bold"),
-        ("question", "bold"),
-        ("answer", "fg:#ff9d00 bold"),
-        ("pointer", "fg:#ff9d00 bold"),
-        ("highlighted", "fg:#ff9d00 bold"),
-        ("selected", "fg:#cc5454"),
-        ("separator", "fg:#cc5454"),
-        ("instruction", ""),
-        ("text", ""),
-        ("disabled", "fg:#858585 italic"),
-    ]
-
-    # The whole subject will be parsed as message by default
-    # This allows supporting changelog for any rule system.
-    # It can be modified per rule
-    commit_parser: Optional[str] = r"(?P<message>.*)"
-    changelog_pattern: Optional[str] = r".*"
-    change_type_map: Optional[Dict[str, str]] = None
-    change_type_order: Optional[List[str]] = None
-
     # Executed per message parsed by the commitizen
     changelog_message_builder_hook: Optional[
         Callable[[Dict, git.GitCommit], Dict]
@@ -42,8 +19,28 @@ class BaseCommitizen(metaclass=ABCMeta):
 
     def __init__(self, config: BaseConfig):
         self.config = config
-        if not self.config.settings.get("style"):
-            self.config.settings.update({"style": BaseCommitizen.default_style_config})
+        self.style = Style(self.config.settings.get("style", defaults.style))
+        self.bump_pattern: Optional[str] = self.config.settings.get(
+            "bump_pattern", defaults.bump_pattern
+        )
+        self.bump_map: Optional[Dict[str, str]] = self.config.settings.get(
+            "bump_map", defaults.bump_map
+        )
+        self.change_type_order: Optional[List[str]] = self.config.settings.get(
+            "change_type_order", defaults.change_type_order
+        )
+        self.change_type_map: Optional[Dict[str, str]] = self.config.settings.get(
+            "change_type_map", defaults.change_type_map
+        )
+        self.commit_parser: Optional[str] = self.config.settings.get(
+            "commit_parser", defaults.commit_parser
+        )
+        self.changelog_pattern: Optional[str] = self.config.settings.get(
+            "changelog_pattern", defaults.changelog_pattern
+        )
+        self.version_parser = self.config.settings.get(
+            "version_parser", defaults.version_parser
+        )
 
     @abstractmethod
     def questions(self) -> Questions:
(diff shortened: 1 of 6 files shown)


commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -172,39 +176,47 @@ def message(self, answers: dict) -> str:
...
     def example(self) -> str:
-        return (
+        return self.config.settings.get(
+            "example",
             "fix: correct minor typos in code\n"
             "\n"
             "see the issue for details on the typos fixed\n"
             "\n"
-            "closes issue #12"
+            "closes issue #12",
```
amitlevy21, line 185: not sure why formatter insisted on comma here

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -172,39 +176,47 @@ def message(self, answers: dict) -> str:
...
     def schema(self) -> str:
-        return (
+        return self.config.settings.get(
+            "schema",
             "<type>(<scope>): <subject>\n"
             "<BLANK LINE>\n"
             "<body>\n"
             "<BLANK LINE>\n"
-            "(BREAKING CHANGE: )<footer>"
+            "(BREAKING CHANGE: )<footer>",
```
amitlevy21, line 195: and here

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -148,9 +140,21 @@ def questions(self) -> Questions:
                 ),
             },
         ]
-        return questions
+
+        # TODO: How would filter functions would be handled from config?
+        return self.config.settings.get("questions", questions)
```
amitlevy21, line 145: would like to consult on this

commitizen/cz/conventional_commits/conventional_commits.py
Lee-W: AFAIK, that's how black works. It prefer adding a following comma to avoid the error that you add another item without adding this comma in the future.

commitizen/cz/conventional_commits/conventional_commits.py
Lee-W: same above

commitizen/cz/conventional_commits/conventional_commits.py
Lee-W: yes, what was it?

Lee-W (commented):
Thanks for your contributions! 

I'm not sure whether we're to deprecate CustomizeCommitsCz now. We'll do it eventually. But I think it might be better for us to keep it but add a deprecation warning (e.g., we're going to deprecate it on 3.0) @Woile What do you think?

commitizen/cz/conventional_commits/conventional_commits.py
amitlevy21: Since questions can have filter functions, how would they be handled read from a config file? From my understanding, this was also not handled in the past

Lee-W: yes, this was not handled. IMO, if we're to build a complicate customization, we should fallback to implement a CZ class. Using config file is just a handy way to achieve easy ones

amitlevy21:
@Lee-W Hi, made some changes and fixed tests, Can you take another look?

commitizen/cz/base.py
```diff
@@ -42,8 +19,28 @@ class BaseCommitizen(metaclass=ABCMeta):
 
     def __init__(self, config: BaseConfig):
         self.config = config
-        if not self.config.settings.get("style"):
-            self.config.settings.update({"style": BaseCommitizen.default_style_config})
+        self.style = Style(self.config.settings.get("style", defaults.style))
+        self.bump_pattern: Optional[str] = self.config.settings.get(
```
Lee-W, line 23: I would say the `defaults.bump_pattern` is something only for `conventional commit` that I want to move out from default and move it to conventional commit. Thus, I'd like to keep the default value of this as `None`

Lee-W: This behavior should be applied to all the following default values which did not exist in `BaseCommitizen`

commitizen/cz/jira/jira.py
```diff
@@ -1,12 +1,20 @@
...
 from commitizen.cz.base import BaseCommitizen
 from commitizen.defaults import Questions
 
 __all__ = ["JiraSmartCz"]
 
 
 class JiraSmartCz(BaseCommitizen):
+    def __init__(self, config: BaseConfig):
+        super().__init__(config)
+        self.bump_map = None
```
Lee-W, line 13: I think the values here is the ideal default values in `BaseCommitizen` and we should move the `defaults` to `ConventionalCommitsCz`

Lee-W (commented):
Hi @amitlevy21 , sorry for the super late reply. I just took a look and left some comments. Let me know if you have any though on it. Thanks!

Lee-W:
As this has no update since 2023 and there might be lots of conflicts, I'm going to close this one. Feel free to reopen it, if we still want this improvement in 🙂

PR #550, opened 2022-08-05
refactor(tool): use charset_normalizer instead of chardet

gpongelli:
## Description
move from chardet to charset_normalizer to avoid recent issue on wrong encoding.

## Checklist

## Expected behavior
commitizen is able to decode command's strings without issue on `decode` .

## Additional context

https://github.com/commitizen-tools/commitizen/pull/545

https://github.com/commitizen-tools/commitizen/issues/544

### DIFF
--- a/commitizen/cmd.py
+++ b/commitizen/cmd.py
@@ -1,7 +1,7 @@
 import subprocess
 from typing import NamedTuple
 
-import chardet
+from charset_normalizer import from_bytes
 
 
 class Command(NamedTuple):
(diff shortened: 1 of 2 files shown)


jenstroeger:
In the future I’d appreciate it if people communicate the duplicated effort, so that I don’t have to waste my time writing a PR only so it will be replaced by another.

Woile:
I think your change is still valid. It's worth changing the chardet library in this PR. But we can still have the same issue as before, or I'm missing something? Sorry for the inconvenience

gpongelli:
Hi Jens, sorry but adding chardet comes from my PR and, after all the recent issues made by it, I felt unhappy so I’ve searched and tested an alternative.

I hope it works, and probably could be useful to add some more tests to avoid regression.

Issue #414, opened 2021-08-21
Option to allow `fixup!` and `squash!` in commit messages


lobotmcj:
## Description
It would be nice if `cz check` didn't fail for fixup and squash commits, e.g., `git commit --fixup <rev>` or `git commit --squash <rev>`

## Possible Solution
Add an option for this and, if true, prepend the commit message check with something like `r"((fixup|squash)\!)?\s"`

Lee-W:
I'm good with this feature @Woile What do you think?

Yajo:
I don't think this is related to git itself. It's just a matter of allowing fixup commits in the regexp.

For example, having the commit-msg pre-commit hook installed, there should be a way to let this *not* fail:

```sh
> git commit -a --fixup 57a2044f1290cba0ca6530e6e77cd5da243e74b1
commitizen check.........................................................Failed
- hook id: commitizen
- exit code: 14

commit validation: failed!
please enter a commit message in the commitizen format.
commit "": "fixup! feat: allow easy working"
pattern: (build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)(\(\S+\))?!?:(\s.*)
```

Lee-W:
if that's the case, I'm ok with it. @Woile what do you think?

Woile:
fine by me as well :+1:

Lee-W:
Thanks @wombelix I think it's a great workaround

Lee-W:
I think this is supported by default in the latest version. Let's close it. Thanks @bearomorphism  for reminding me

PR #551, opened 2022-08-08
feat: Ignore fixup! & squash!

a5r0n:
## Description

#414 

## Checklist

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -109,6 +109,8 @@ class Check:
             commit_msg.startswith("Merge")
             or commit_msg.startswith("Revert")
             or commit_msg.startswith("Pull request")
+            or commit_msg.startswith("fixup!")
+            or commit_msg.startswith("squash!")
         ):
             return True
         return bool(re.match(pattern, commit_msg))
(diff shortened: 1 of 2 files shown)


bhelgs:
Disclaimer: I am not a maintainer. 


I wonder what happens to the changelog if you forget to rebase these out of existence. Is it (correctly) ignoring them?

a5r0n:
the `validate_commit_message`  called only when u run `cz check` (like the pre-commit do).
so i think the changelog be not impacted.

Lee-W (approved):
@a5r0n Thanks for contribution! I'm good with it. @Woile could you please check whether you're good with it. Thanks!

eldipa:
Could be possible to load the prefixes to ignore from the config file? I'm thinking that `Merge`, `Revert`, `fixup!` are quite standard but they may not fit in any project and a simple solution could load the prefixes from a file (leaving the current prefixes `Merge`, `Revert` and so on as default).

What do you think?

Lee-W:
Hi @eldipa , I think that's a good idea! But I wish to keep this PR as it is. We could create another issue / PR for tracking the feature you mention. Thanks!

Lee-W:
@woile I'm thinking of merging this one this week. Please let me know if we need more discussion on this one :)

Issue #191, opened 2020-05-23
[Feature Request] Line length warning


Lee-W:
## Description
It's recommended to have only 72 characters. We could add a warning or a hint if the generated title might exceed 72.

## Possible Solution
Add a line hint like vim

![image](https://user-images.githubusercontent.com/5144808/82725484-eaf1b200-9d0f-11ea-9f05-5f762f2fc9b3.png)

Woile:
You mean for the commit subject? Where would we add this?

Remember that this is not a problem for the changelog. I'm not sure why the line length was chosen to be 72.

Lee-W:
The first line of our generated commit. e.g., `feat(new): awesome feature`.
I'm thinking of make the functions in `commitizen/utils.py` and this funtionality something like `commitizen-plugins`. But it's still a rough idea.

I guess the number is from [here](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting)

Woile:
IMO we should add it as a config parameter (`max_subject_length`), disabled by default, or set to `0` which would mean: anything.

I like the plugins idea, not sure how to hook everything haha but it's interesting

Lee-W:
This could be my next target after #128 haha. Also, if my memory serves me right, you've mentioned implementing changelog-hooks (like email-hooks or slack-hooks). Do you have any idea how we should group them. Add them into to this repo as extra deps? Or, create another repo like **commitizen-changelog-hooks**?

Woile:
I was thinking more like separate packages, so people can choose which one to install:

```
pip install cz-slack-hook cz-email-hook
```

and then in your `cz.toml`
```bash
hooks = [
  "slack-hook",
  "email-hook"
] 
```

jtpavlock:
I think a reasonable default would be 72 characters for the commit title, as that's the [max github displays in the commit history](https://github.com/commitizen-tools/commitizen/commits/master). Perhaps the title and body should be able to be configured differently. I don't care as much about a character limit on the body, as github or other's should wrap the content to fit the screen.

Lee-W:
Here's a way to implement similar functionality https://github.com/commitizen-tools/commitizen/issues/331#issue-780669838. Not yet tried this functionality on cz-cli

iknowright:
Hi there!
I found a feasible solution for line length warning. It is done by `questionary`'s text validator, which will show a red color background warning at the bottom-left corner of the console.

the sample code is:
```
import questionary
[...]

PR #557, opened 2022-08-14
feat: check if commit message is no more than 72 characters

kevin1kevin1k:
## Description
Check if the length of commit message (the `prefix(scope): subject` part) does not exceed 72 characters.
Currently only supports ConventionalCommitsCz; JiraSmartCz and CustomizeCommitsCz not implemented.
An imperfect workaround for #191.

## Checklist

## Expected behavior

If run `cz commit` with `-cl` or `--check-length` argument, the commit would fail when `prefix(scope): subject` is longer than 72 chars.
Without the argument, the default behavior is to accept longer messages.

1. `cz commit -cl`
2. ENTER (type prefix: fix)
3. ENTER (empty scope)
4. 01234567890123456789012345678901234567890123456789012345678901234567 (for subject)
5. ENTER (empty body)
6. ENTER (not breaking change)
7. ENTER (no footer)

Then cz should exit with message `Length of commit message exceeded limit (73/72)` and error code 23.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1461,28 +1461,28 @@ jupyter = ["ipywidgets (>=7.5.1,<9)"]
 
 [[package]]
 name = "ruff"
-version = "0.3.7"
+version = "0.4.0"
 description = "An extremely fast Python linter and code formatter, written in Rust."
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "ruff-0.3.7-py3-none-macosx_10_12_x86_64.macosx_11_0_arm64.macosx_10_12_universal2.whl", hash = "sha256:0e8377cccb2f07abd25e84fc5b2cbe48eeb0fea9f1719cad7caedb061d70e5ce"},
-    {file = "ruff-0.3.7-py3-none-macosx_10_12_x86_64.whl", hash = "sha256:15a4d1cc1e64e556fa0d67bfd388fed416b7f3b26d5d1c3e7d192c897e39ba4b"},
-    {file = "ruff-0.3.7-py3-none-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:d28bdf3d7dc71dd46929fafeec98ba89b7c3550c3f0978e36389b5631b793663"},
-    {file = "ruff-0.3.7-py3-none-manylinux_2_17_armv7l.manylinux2014_armv7l.whl", hash = "sha256:379b67d4f49774ba679593b232dcd90d9e10f04d96e3c8ce4a28037ae473f7bb"},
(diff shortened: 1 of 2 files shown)


commitizen/commands/commit.py
```diff
@@ -29,6 +29,8 @@ def __init__(self, config: BaseConfig, arguments: dict):
         self.config: BaseConfig = config
         self.cz = factory.commiter_factory(self.config)
         self.arguments = arguments
+        print(self.config)
```
Lee-W, line 32: I'd suggest we don't print anything not directly related to `cz commit`

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -167,7 +169,15 @@ def message(self, answers: dict) -> str:
         if footer:
             footer = f"\n\n{footer}"
 
-        message = f"{prefix}{scope}: {subject}{body}{footer}"
+        message = f"{prefix}{scope}: {subject}"
+        message_len = len(message)
+        MESSAGE_LEN_LIMIT = 72
```
Lee-W, line 174: Could we make it configurable?

commitizen/cz/jira/jira.py
```diff
@@ -44,7 +45,7 @@ def questions(self) -> Questions:
         ]
         return questions
 
-    def message(self, answers) -> str:
+    def message(self, answers: dict, check_length: Optional[bool] = False) -> str:
```
Lee-W, line 48: Maybe we could implement such check on customize and jira as well 🤔

commitizen/cz/conventional_commits/conventional_commits.py
Lee-W: Also, in the comment https://github.com/commitizen-tools/commitizen/issues/191#issuecomment-1216952884, he proposes a reasonable solution. Although global variables might sometimes not be desired. I'm good with it in this case. @kevin1kevin1k @Woile  What do you think?

woile:
What's the status of this PR? Do we still want this feature?

Lee-W:
I think this feature is still what we want but do not have bandwidth to work on it at this moment

schlotter:
This PR is useful. What is missing to get it merged?

Lee-W:
> This PR is useful. What is missing to get it merged?

I think we should make it a configurable feature.

schlotter:
@kevin1kevin1k Could you please make `MESSAGE_LEN_LIMIT` configurable?

kevin1kevin1k:
Sorry for the late response.
I will find time to resolve the comments and make it configurable!

nowNick:
Hey :) 
What's the status of this PR? @kevin1kevin1k - do you think you'll have some time to wrap it up 🙇 ?

kevin1kevin1k:
Thanks for the reminder!
I've still been busy lately, but I guess I can find time to finish it up in the upcoming weeks.

PR #565, opened 2022-08-21
Self Check With New `commitizen-branch` pre-commit Hook

Kurt-von-Laven:
## Description

Check existing commit messages. Use new `commitizen-branch` hook to verify correctness of existing commit messages post-commit and pre-push. This guards against cherry-picked commits with invalid commit messages as well as empty commit messages (e.g., created with `git commit --allow-empty-message`).

Run hooks pre-commit by default. They already run pre-push, and this behavior is left unchanged. Running hooks pre-commit gives developers more immediate feedback and improves commit quality by blocking an often unnecessary series of fix commits.

When testing this change locally, `poetry install` failed with the following error:

```
Creating virtualenv commitizen-9emEp_2F-py3.10 in ~/.cache/pypoetry/virtualenvs
Installing dependencies from lock file
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.

  SolverProblemError

  Because commitizen depends on charset-normalizer (^2.1.0) which doesn't match any versions, version solving failed.

  at ~/.asdf/installs/poetry/1.1.14/lib/poetry/puzzle/solver.py:241 in _solve
      237│             packages = result.packages
      238│         except OverrideNeeded as e:
      239│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      240│         except SolveFailure as e:
    → 241│             raise SolverProblemError(e)
      242│ 
      243│         results = dict(
      244│             depth_first_search(
      245│                 PackageNode(self._package, packages), aggregate_package_nodes
```

I think `poetry update` may need to be run, but when I tried that, I encountered 51 test failures.

## Checklist

- [x] Add test cases to all the changes you introduce
- [ ] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes

## Expected behavior
The new `commitizen-branch` hook passes.

## Steps to Test This Pull Request
1. Run `poetry update` to work around issue described above.
2. Run `poetry install` to install pre-commit and other dependencies.
3. Make a test commit since presently the `commitizen-branch` hook fails when the range `origin/HEAD..HEAD` contains no commits.
4. Run `poetry run pre-commit run --all-files --hook-stage push` to run all pre-push hooks.

## Additional context
Follows on #517.

PR #567, opened 2022-08-22
fix(bump): Support regexes containing colons

### LINKED ISSUE #499: Support Regexes Containing Colons
### Description

One limitation of [this very cool feature](https://commitizen-tools.github.io/commitizen/bump/#version_files) is that if you put a `:` in your regex, anything from the colon on will be ignored (and there is no way to escape the colon).

### Possible Solution

Use [`str.partition()`](https://docs.python.org/3/library/stdtypes.html#str.partition) rather than [`str.split()`](https://docs.python.org/3/library/stdtypes.html#str.split) [here](https://github.com/commitizen-tools/commitizen/blob/605b1252c2b9bb4ce82d14e1fde26a8139ad5e23/commitizen/bump.py#L144) (and make the corresponding changes this entails).

### Additional context

_No response_

### Additional context

_No response_

Kurt-von-Laven:
## Description

Fixes #499.

The `version_files` feature uses a colon to delimit filenames from a regex to use to find the version. Make `version_files` slightly more robust by splitting on the first colon rather than all colons, thereby permitting the regex to contain a colon. Maintain the pre-existing assumption that the filename doesn't contain a colon, as this is rare and would likely cause many other problems.

Include colon in `version_files` regex. Commitizen runs bump on itself to, among other things, manage the version it uses of its own pre-commit hooks. When searching `.pre-commit-config.yaml` for the version number of Commitizen, look for `"rev:"` rather than `"rev"` to reduce the likelihood of a false positive now that colons are permitted in `version_files` regexes.

## Checklist

## Expected behavior
The version of the Commitizen pre-commit hooks is incremented along with all other references to the Commitizen version number.

1. Run `cz bump`.

## Additional context
As mentioned in #565, I am encountering some difficulties testing Commitizen locally.

### DIFF
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -141,8 +141,7 @@ def update_version_in_files(
     """
     # TODO: separate check step and write step
     for location in files:
-        filepath, *regexes = location.split(":")
-        regex = regexes[0] if regexes else None
+        filepath, _, regex = location.partition(":")
 
         with open(filepath, "r") as f:
             version_file = f.read()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ tag_format = "v$version"
 version_files = [
   "pyproject.toml:version",
   "commitizen/__version__.py",
-  ".pre-commit-config.yaml:rev.\\s+(?=[^\\n]+Commitizen)"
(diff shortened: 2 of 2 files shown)


Woile:
Wow! Thanks for the fix 🚀

PR #568, opened 2022-08-22
fix(bump): Search for version number line by line

### LINKED ISSUE #498: Version Not Bumped If Regex Right of Version
### Description

I attempted to use Commitizen to automatically increment a version number in [`.pre-commit-config.yaml`](https://pre-commit.com/#2-add-a-pre-commit-configuration).

`pyproject.toml`:
```toml
...
[tool.commitizen]
version = "0.0.0"
version_files = [
    ".pre-commit-config.yaml:Commitizen",
    "pyproject.toml:version"
]
...
```

`.pre-commit-config.yaml`:
```yaml
...
- repo: ./
  rev: 0.0.0 # automatically updated by Commitizen
  hooks:
    - id: my-custom-hook
...
```

### Steps to reproduce

1. Run `cz bump`.

### Current behavior

The version numbers in `pyproject.toml` are incremented, but not the version number in `.pre-commit-config.yaml`. My understanding of [the pertinent source code](https://github.com/commitizen-tools/commitizen/blob/605b1252c2b9bb4ce82d14e1fde26a8139ad5e23/commitizen/bump.py#L185) is that it only looks for the version number to the right of the regex.

### Desired behavior

The version number in `.pre-commit-config.yaml` is incremented just like the version numbers in `pyproject.toml`. It may be simpler to process the file line by line. Some may be tempted to match on `rev` rather than a string in the comment, such as `Commitizen`, but that would cause Commitizen to bump _every_ repo's rev that ever happens to be at the same version to the repository itself. YAML relies on vertical alignment, so I'm not aware of a way I can put a comment before the version number in my scenario.

### Workarounds
Since the current processing is not line-by-line, in my particular scenario, I can specify a regex that matches part of the preceding line and part of the line containing the version number (e.g., `repo.*\\./\\n\\s+rev`). Personally, I don't feel that it is important to retain such [undocumented](https://commitizen-tools.github.io/commitizen/bump/#version_files) functionality. Matching across lines in this manner seems likely to result in cryptic and brittle regexes (e.g., suppose a comment is added between the two lines) particularly since backslashes need to be escaped in TOML. A better approach, in my opinion, is to leverage lookahead assertions (e.g., `rev.\\s+(?=[^\\n]+Commitizen)`) to stop consuming the string before reaching the version number, but this still implies some unnecessary complexity and is prone to the pitfall of the developer assuming the matching is performed line-by-line. For instance, `rev.\\s+(?=.+Commitizen)` would match the first line matching `rev.\\s+` provided "Commitizen" appears anywhere subsequently in the file. One of our open-source repositories [demonstrates the lookahead assertion technique](https://github.com/ScribeMD/pre-commit-hooks/blob/25c3bf7fc87058729a2231e7383bde4e2907ca27/pyproject.toml#L9) in case it's helpful to look around.

### Screenshots

_No response_

### Environment

Commitizen Version: 2.23.0
Python Version: 3.10.4 (main, Apr  5 2022, 17:28:03) [GCC 11.2.0]
Operating System: Linux

Kurt-von-Laven:
## Description

Fixes #498.

Add mypy typing for `_bump_with_regex`.

Avoid the complexities of multiline regex matching and inconsistencies with the case where no regex is specified. Simplify the implementation by eliminating the need to track an offset into the file. Ensure that the version number will be found even if it is to the left of the portion of the line matching the regex.

Remove lookahead assertion from `version_files` entry for `.pre-commit-config.yaml`. Commitizen runs bump on itself to, among other things, manage the version it uses of its own pre-commit hooks. When searching `.pre-commit-config.yaml` for the version number of Commitizen, we search for "Commitizen," which is in a comment to the right of the version number. Remove the lookahead assertion for "Commitizen," and instead simply consume it when matching now that `version_files` regexes don't need to exclusively match to the left of the version number.

## Checklist

## Expected behavior
The version of Commitizen is bumped as usual in all `version_files` listed in `pyproject.toml`.

1. Run `cz bump`.

## Additional context
As mentioned in #565, I am encountering some difficulties testing Commitizen locally.

### DIFF
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -142,18 +142,12 @@ def update_version_in_files(
     # TODO: separate check step and write step
     for location in files:
         filepath, _, regex = location.partition(":")
+        if not regex:
+            regex = _version_to_regex(current_version)
 
-        with open(filepath, "r") as f:
-            version_file = f.read()
-
-        if regex:
-            current_version_found, version_file = _bump_with_regex(
-                version_file, current_version, new_version, regex
-            )
-        else:
-            current_version_regex = _version_to_regex(current_version)
-            current_version_found = bool(current_version_regex.search(version_file))
-            version_file = current_version_regex.sub(new_version, version_file)
+        current_version_found, version_file = _bump_with_regex(
+            filepath, current_version, new_version, regex
+        )
 
         if check_consistency and not current_version_found:
             raise CurrentVersionNotFoundError(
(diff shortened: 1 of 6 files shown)


Woile:
LGTM! @Lee-W feel free to merge

Woile (approved):

Issue #520, opened 2022-05-25
`test_bump_pre_commit_changelog` is failing on `master` (afa0d93) for some environments


bhelgs:
### Description

I am seeing some tests failing:
1. `test_bump_pre_commit_changelog_fails_always[True]`
2. `test_bump_pre_commit_changelog[True]`
3. `test_bump_pre_commit_changelog_fails_always[False]`

**See resolution (pre-commit was silently failing):**
https://github.com/commitizen-tools/commitizen/issues/520#issuecomment-1139207915

### Steps to reproduce

1. cloned repo
2. poetry install
3. set up hooks
4. `./scripts/test`

### Current behavior

`./scripts/test`:
```
FAILED tests/test_bump_create_commit_message.py::test_bump_pre_commit_changelog_fails_always[True] - Failed: DID NOT RAISE <class 'commitizen.exceptions.BumpCommitFailedError'>
FAILED tests/test_bump_create_commit_message.py::test_bump_pre_commit_changelog[True] - AssertionError: assert '## 0.1.1 (20...\n\n- _test\n' == '## 0.1.1 (20...\n-   \\_test\n'
FAILED tests/test_bump_create_commit_message.py::test_bump_pre_commit_changelog_fails_always[False] - Failed: DID NOT RAISE <class 'commitizen.exceptions.BumpCommitFailedError'>
```

I was able to alter [this assert in](https://github.com/commitizen-tools/commitizen/blob/afa0d93544405bb0a871b5072e0a814720d933be/tests/test_bump_create_commit_message.py#L68) `test_bump_pre_commit_changelog` to get that test to work.
![image](https://user-images.githubusercontent.com/33836927/170322750-af9a1a10-e174-425f-b8d9-bfdd9c26e7bb.png)

What could be different about my machine? I assume the `  \\` had purpose otherwise the CI would have failed a couple of days ago when this test was created.

### Desired behavior

test pass

### Environment

```
commit hash: afa0d93
platform linux -- Python 3.6.8, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
plugins: regressions-2.3.1, xdist-2.5.0, mock-2.0.0, freezegun-0.4.2, forked-1.4.0, cov-2.12.1, datadir-1.3.1
```

bhelgs:
@yajo sorry to bother you over something probably really straightforward.
Any guess what the purpose of the `  \\` was?

yajo:
[...]

PR #571, opened 2022-08-26
fix(bump): log git commit stderr and stdout during bump

bhelgs:
## Description

I notice that the the [commit command outputs both `stderr` and `stdout` of `git commit`](https://github.com/commitizen-tools/commitizen/blob/8cb1e22d72abd855f221556f2a1c8fc0ddded109/commitizen/commands/commit.py#L104). I am thinking it might be a good idea for the bump command to do the same. This way if there are any hook warnings it will show up in the bump output.

What do you think?

## Checklist

## Expected behavior
*  `cz bump --changelog` outputs: 
image

*  `scripts/test`

## Additional context
Helpful in diagnosing #520

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -230,6 +230,12 @@ class Bump:
             c = git.commit(message, args=self._get_commit_args())
         if c.return_code != 0:
             raise BumpCommitFailedError(f'2nd git.commit error: "{c.err.strip()}"')
+
+        if c.out:
+            out.write(c.out)
+        if c.err:
+            out.write(c.err)
+
         c = git.tag(
             new_tag_version,
             signed=self.bump_settings.get("gpg_sign", False)


commitizen/commands/bump.py
```diff
@@ -230,6 +230,13 @@ def __call__(self):  # noqa: C901
             c = git.commit(message, args=self._get_commit_args())
         if c.return_code != 0:
             raise BumpCommitFailedError(f'2nd git.commit error: "{c.err.strip()}"')
+
+        if c.out:
+            out.write("git commit stdout: " + c.out)
```
Woile, line 235: do we need the string prefix? or doing `out.write(c.out)` would be enough?

bhelgs: sounds good.

bhelgs: done.

Lee-W (approved):
LGTM! Thanks @bhelgs ! 

@woile   If there's no further discussion needed, I plan to merge it these days.

Issue #428, opened 2021-09-26
Run `cz init` will raise `UnicodeDecodeError`


Bogay:
## Description

The `cz init` command may fail unexpectedly under some condition.  
In this case, the error may be caused by my command output is not encoding by UTF-8. I have tried to decode the output using [CP950](https://en.wikipedia.org/wiki/Code_page_950) and after that I can see a error message: `'pre-commit' 不是內部或外部命令、可執行的程式或批次檔。`, which means that PowerShell can not find `pre-commit` to execute, because I forget to install it before run `cz init`.

## Steps to reproduce

1. Create a empty directory
2. Enter the directory
3. execute `cz init` and use the default setttings

## Current behavior

It will raise `UnicodeDecodeError`, logs are attached below:

```pwsh
? Please choose a supported config file: (default: pyproject.toml) pyproject.toml
? Please choose a cz (commit rule): (default: cz_conventional_commits) cz_conventional_commits
No Existing Tag. Set tag to v0.0.1
? Please enter the correct version format: (default: "$version")
? Do you want to install pre-commit hook? Yes
Traceback (most recent call last):
  File "c:\users\aa\appdata\local\programs\python\python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\aa\appdata\local\programs\python\python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\AA\AppData\Local\Programs\Python\Python38\Scripts\cz.exe\__main__.py", line 7, in <module>
  File "c:\users\aa\appdata\local\programs\python\python38\lib\site-packages\commitizen\cli.py", line 301, in main
    args.func(conf, vars(args))()
  File "c:\users\aa\appdata\local\programs\python\python38\lib\site-packages\commitizen\commands\init.py", line 43, in __call__
    self._install_pre_commit_hook()
  File "c:\users\aa\appdata\local\programs\python\python38\lib\site-packages\commitizen\commands\init.py", line 144, in _install_pre_commit_hook
    c = cmd.run("pre-commit install --hook-type commit-msg")
  File "c:\users\aa\appdata\local\programs\python\python38\lib\site-packages\commitizen\cmd.py", line 23, in run
    return Command(stdout.decode(), stderr.decode(), stdout, stderr, return_code)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa4 in position 13: invalid start byte
```

[...]

PR #578, opened 2022-09-07
Report hook installation fail

Bogay:
## Description

This PR make `cz` report errors if user want to install pre-commit hook and failed, which fixs #428. 

## Checklist

## Expected behavior

Run `cz init` will print error if there is something wrong.

```
❯ cz init
? Please choose a supported config file: (default: pyproject.toml) pyproject.toml
? Please choose a cz (commit rule): (default: cz_conventional_commits) cz_conventional_commits
No Existing Tag. Set tag to v0.0.1
? Please enter the correct version format: (default: "$version")
? Do you want to install pre-commit hook? Yes
pre-commit is not installed in current environement.
Installation failed. See error outputs for more information.
```

And the exit code should not be 0.

```
❯ echo $?
24
```

```
# Run `cz init` with default choices
cz init
# Check the exit code should not be 0
test $? != 0
```

## Additional context

I found that `TestPreCommitCases` has some duplicated code. Should they be extracted to methods?

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -50,10 +50,10 @@ class Init:
             ],
         ).ask()
         if hook_types:
-            if not self._install_pre_commit_hook(hook_types):
-                raise InitFailedError(
-                    "Installation failed. See error outputs for more information."
-                )
+            try:
+                self._install_pre_commit_hook(hook_types)
+            except InitFailedError as e:
+                raise InitFailedError(f"Failed to install pre-commit hook.\n{e}")
 
         out.write("You can bump the version and create changelog running:\n")
         out.info("cz bump --changelog")
@@ -120,24 +120,32 @@ class Init:
                 tag_format = "$version"
         return tag_format
 
-    def _search_pre_commit(self):
+    def _search_pre_commit(self) -> bool:
+        """Check whether pre-commit is installed"""
         return shutil.which("pre-commit") is not None
 
     def _exec_install_pre_commit_hook(self, hook_types: List[str]):
+        cmd_str = self._gen_pre_commit_cmd(hook_types)
+        c = cmd.run(cmd_str)
+        if c.return_code != 0:
+            err_msg = (
+                f"Error running {cmd_str}."
+                "Outputs are attached below:\n"
+                f"stdout: {c.out}\n"
+                f"stderr: {c.err}"
+            )
+            raise InitFailedError(err_msg)
+
+    def _gen_pre_commit_cmd(self, hook_types: List[str]) -> str:
+        """Generate pre-commit command according to given hook types"""
         if not hook_types:
             raise ValueError("At least 1 hook type should be provided.")
         cmd_str = "pre-commit install " + "".join(
             f"--hook-type {ty}" for ty in hook_types
         )
-        c = cmd.run(cmd_str)
-        if c.return_code != 0:
-            out.error(f"Error running {cmd_str}. Outputs are attached below:")
-            out.error(f"stdout: {c.out}")
-            out.error(f"stderr: {c.err}")
-            return False
-        return True
+        return cmd_str
 
-    def _install_pre_commit_hook(self, hook_types: Optional[List[str]] = None) -> bool:
+    def _install_pre_commit_hook(self, hook_types: Optional[List[str]] = None):
         pre_commit_config_filename = ".pre-commit-config.yaml"
         cz_hook_config = {
             "repo": "https://github.com/commitizen-tools/commitizen",
@@ -173,16 +181,13 @@ class Init:
             yaml.safe_dump(config_data, stream=config_file)
 
         if not self._search_pre_commit():
-            out.error("pre-commit is not installed in current environement.")
-            return False
-
+            raise InitFailedError(
+                "pre-commit is not installed in current environement."
+            )
         if hook_types is None:
             hook_types = ["commit-msg", "pre-push"]
-        if not self._exec_install_pre_commit_hook(hook_types):
-            return False
-
+        self._exec_install_pre_commit_hook(hook_types)
         out.write("commitizen pre-commit hook is now installed in your '.git'\n")
-        return True
 
     def _update_config_file(self, values: Dict[str, Any]):
         for key, value in values.items():
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -1,4 +1,6 @@
 import os
+import shutil
+from typing import Any, Dict
 
 import questionary
 import yaml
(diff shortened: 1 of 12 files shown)


Lee-W:
> I found that TestPreCommitCases has some duplicated code. Should they be extracted to methods?

Yep that would be great

tests/test_cli.py
```diff
@@ -55,11 +57,13 @@ def test_ls(mocker, capsys):
...
 
 
-def test_arg_debug(mocker):
+def test_arg_debug(mocker: MockFixture):
     testargs = ["cz", "--debug", "info"]
     mocker.patch.object(sys, "argv", testargs)
     cli.main()
-    assert sys.excepthook.keywords.get("debug") is True
+    excepthook = sys.excepthook
+    assert isinstance(excepthook, partial)
```
Lee-W, line 65: I'm a bit confused. Why is excepthook a partial?

Bogay: @Lee-W Because `sys.excepthook` is replaced by a `partial` here.
```py
    if args.debug:
        logging.getLogger("commitizen").setLevel(logging.DEBUG)
        sys.excepthook = commitizen_debug_excepthook
    elif args.no_raise:
        no_raise_exit_codes = parse_no_raise(args.no_raise)
        no_raise_debug_excepthook = partial(
            commitizen_excepthook, no_raise=no_raise_exit_codes
        )
        sys.excepthook = no_raise_debug_excepthook
```
https://github.com/commitizen-tools/commitizen/blob/master/commitizen/cli.py#L379-L387

and `keywords` is a property of `partial` object.

Maybe I should add some comments to explain this?

commitizen/commands/init.py
```diff
@@ -109,7 +113,20 @@ def _ask_tag_format(self, latest_tag) -> str:
                 tag_format = "$version"
         return tag_format
 
-    def _install_pre_commit_hook(self):
+    def _search_pre_commit(self):
+        return shutil.which("pre-commit") is not None
+
+    def _exec_install_pre_commit_hook(self):
+        cmd_str = "pre-commit install --hook-type commit-msg"
```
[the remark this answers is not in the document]
Bogay: @Lee-W
Should installing both `commit-msg` and `pre-push` hooks be the default behavior?
If we want to let user choose which type of hooks to install, maybe we can add an [checkbox](https://questionary.readthedocs.io/en/stable/pages/types.html#checkbox) for that.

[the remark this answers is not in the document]
Lee-W: Yep, I think the checkbox option is the best solution!

tests/test_cli.py
Lee-W: ah got it. Yep, it would be really helpful! Even though it was what I implemented, I still forget this 🤦‍♂️

commitizen/commands/init.py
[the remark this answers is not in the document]
Bogay: If checkbox is preferred, what should be the default choices? Both `commit-msg` and `pre-push`?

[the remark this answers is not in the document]
Lee-W: yep, I think that's reasonable

Bogay:
@Lee-W I think it's ready to be reviewed.

commitizen/commands/init.py
```diff
@@ -19,34 +21,43 @@ def __init__(self, config: BaseConfig, *args):
...
+
+        hook_types = questionary.checkbox(
+            "What types of pre-commit hook you want to install? (Leave blank if you don't want to install)",
+            choices=[
+                questionary.Choice("commit-msg", checked=True),
+                questionary.Choice("pre-push", checked=True),
+            ],
+        ).ask()
+        if hook_types:
+            if not self._install_pre_commit_hook(hook_types):
```
Lee-W, line 53: I think we should put the following code into `_exec_install_pre_commit_hook`. This seems to be the responsibility of that function. @Bogay what do you think?

```python
                raise InitFailedError(
                    "Installation failed. See error outputs for more information."
                )
```

commitizen/commands/init.py
```diff
@@ -109,17 +120,37 @@ def _ask_tag_format(self, latest_tag) -> str:
                 tag_format = "$version"
         return tag_format
 
-    def _install_pre_commit_hook(self):
+    def _search_pre_commit(self):
```
Lee-W, line 123: nitpick: missing return type anntation

commitizen/commands/init.py
```diff
@@ -109,17 +120,37 @@ def _ask_tag_format(self, latest_tag) -> str:
...
+
+    def _exec_install_pre_commit_hook(self, hook_types: List[str]):
+        if not hook_types:
+            raise ValueError("At least 1 hook type should be provided.")
+        cmd_str = "pre-commit install " + "".join(
+            f"--hook-type {ty}" for ty in hook_types
+        )
+        c = cmd.run(cmd_str)
+        if c.return_code != 0:
+            out.error(f"Error running {cmd_str}. Outputs are attached below:")
```
Lee-W, line 134: nitpick: I might suggest invoke one `out.error` and concatenation the string instead.

Lee-W (commented):
Hi @Bogay sorry for the late review. Just one minor design issue I'd like to discuss. Others are just minor nitpick. You could fix it if you have time. Otherwise we could do that in the future. We're almost ready for this merge!

commitizen/commands/init.py
```diff
@@ -109,17 +120,37 @@ def _ask_tag_format(self, latest_tag) -> str:
...
-    def _install_pre_commit_hook(self):
+    def _search_pre_commit(self):
+        return shutil.which("pre-commit") is not None
+
+    def _exec_install_pre_commit_hook(self, hook_types: List[str]):
+        if not hook_types:
+            raise ValueError("At least 1 hook type should be provided.")
+        cmd_str = "pre-commit install " + "".join(
+            f"--hook-type {ty}" for ty in hook_types
+        )
```
Bogay, line 131: I found that I forgot to add a space between `--hook-type` options...
@Lee-W Do you have any idea how to write tests for this?
Because it depends on `pre-commit`'s behavior, it's harder to test.
```suggestion
        cmd_str = "pre-commit install " + " ".join(
            f"--hook-type {ty}" for ty in hook_types
        )
```

Lee-W: we could verify whether the code generate the pre-commit command we want. e.g., mock the caller function of `cmd_str` to check

Bogay:
@Lee-W it's done 👍

Lee-W (approved):
@Bogay Sorry for taking so long. I think we're ready to merge this one!

@woile I plan to merge this later this week. Let me know if you want to take a deeper look.

Issue #570, opened 2022-08-24
Developmental releases


PsypherPunk:
### Description

We're running into issues using `commitizen` for version-bumping where, if people are working on the same repository simultaneously on different branches, our CI process will bump them to the same version, leading to a collision when tagging.

Previously, we've used a CI-specific suffix to get around this and, as per [PEP 440](https://peps.python.org/pep-0440/#developmental-releases), "Developmental releases" _"may be useful for continuous integration purposes…"_.

The [docs](https://github.com/commitizen-tools/commitizen/blob/master/docs/bump.md) currently state: _"post and dev releases are not supported yet"_ so this may be a useful feature for anyone with a similar workflow…?

### Possible Solution

I've made an initial attempt at this over [here](https://github.com/PsypherPunk/commitizen/tree/add-dev-release).

Here, `--devrelease` has been added as a flag that takes an explicit _"non-negative integer value"_ and simply uses that as part of the `.devN` component of the version, e.g. passing `--devrelease 1234` results in a developmental release segment of `.dev1234`.

It does not attempt to infer the version based on the current version: doing so would run counter to the continuous-integration use-case and potentially result in a collision as per the *Description*.

Lee-W:
Sounds like an interesting feature 🤔 @woile  what do you think?

---

In our team, we usually merge / rebase the main / master branch back to feature branch so that commitizen can still handle the version bumping

PsypherPunk:
Thanks, @Lee-W.

We tend to follow the same workflow: the above only really arises when we have more than one active branch. If we're actively pushing to those branches, because each is relative to `main` and unaware of the other, they end up vying for version numbers.

When one gets merged, as you say, any remaining branches can be rebased and it's less of a problem.

woile:
It does look good!
Could you describe a bit the workflow? I'm wondering how are users expected to automate this. From what I gather there is a `--devrelease N`, but this implies manual input. How would you detect the current `N` in another branch and do `N+1`? 
Thanks!

PsypherPunk:
Our immediate use-case is for either CircleCI and that passes the `CIRCLE_BUILD_NUM` variable to each build job, [which is a build-specific integer](https://circleci.com/docs/variables).

Passing `--devrelease ${CIRCLE_BUILD_NUM}` would allow for a branch-unique `.dev` suffix.

[...]

PR #582, opened 2022-09-13
feat: add functionality for dev-releases

PsypherPunk:
## Description
This adds a `--devrelease` flag, expecting a non-negative integer, as per [PEP 440](https://peps.python.org/pep-0440/#developmental-releases); e.g. passing `--devrelease 1234` results in a developmental release segment of `.dev1234`.

## Checklist

## Expected behavior

When calling `cz bump`, passing `--devrelease`, followed by a non-negative integer will result in the bumped version adding a `.dev` suffix.

1. Check the existing behaviour: `python3 -m commitizen bump --dry-run`; for this branch, this gives `version 2.32.5 → 2.33.0`.
1. Check the new behaviour:  `python3 -m commitizen bump --dry-run --devrelease 0`; for this branch, this gives `version 2.32.5 → 2.33.0.dev0`.

## Additional context

See also discussions on #570.

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -103,6 +103,7 @@ class Bump:
         is_yes: bool = self.arguments["yes"]
         increment: Optional[str] = self.arguments["increment"]
         prerelease: str = self.arguments["prerelease"]
+        devrelease: Optional[int] = self.arguments["devrelease"]
         is_files_only: Optional[bool] = self.arguments["files_only"]
         is_local_version: Optional[bool] = self.arguments["local_version"]
 
@@ -151,6 +152,7 @@ class Bump:
             current_version,
             increment,
             prerelease=prerelease,
+            devrelease=devrelease,
             is_local_version=is_local_version,
         )
 
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -69,6 +69,18 @@ def prerelease_generator(current_version: str, prerelease: Optional[str] = None)
     return pre_version
 
 
+def devrelease_generator(devrelease: int = None) -> str:
+    """Generate devrelease
+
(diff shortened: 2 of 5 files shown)


Lee-W (approved):
Thanks @PsypherPunk ! This looks great!

@woile I'm good with this one. Could you please check when you have time? If no further comment added, I plan to merge it later this week. Thanks!

commitizen/commands/bump.py
```diff
@@ -103,6 +103,7 @@ def __call__(self):  # noqa: C901
         is_yes: bool = self.arguments["yes"]
         increment: Optional[str] = self.arguments["increment"]
         prerelease: str = self.arguments["prerelease"]
+        devrelease: bool = self.arguments["devrelease"]
```
woile, line 106: Why is this a bool?

PsypherPunk: A leftover from an earlier attempt when it was; fix incoming...

PsypherPunk: @woile, `devrelease` should be an `Optional[int]` here, right?

Should `prerelease` be an `Optional[str]`? Comes through as `None` if `--prerelease` isn't passed—if that's correct, shall I fix that while I'm correcting this?

woile: Yes, it should be an `Optional[int]`, as specified [here](https://github.com/commitizen-tools/commitizen/pull/582/files#diff-6a6eecba6ea4aab85dcb0372fe57ce7306496e5a47186850be245bd732e8894eR147)

PsypherPunk:
Comment addressed. Had a failing test but re-tested locally without issue and re-pushed, seems to be okay.

Failed checks, however, do to an apparent drop in overall coverage…?

PR #591, opened 2022-09-18
Bump: support manual version

### LINKED ISSUE #435: Manually bump entire version, not just manually specify which part to increment
The [docs](https://commitizen-tools.github.io/commitizen/bump/#about) state that the version can be manually bumped but I believe this is incorrect as there's no other mention of this in the docs or in the CLI help beyond `--increment {MAJOR,MINOR,PATCH}` which is merely specifying the part to auto-increment.

Why do I want to manually specify the version to bump to? Well, I don't but unfortunately, dev versions are not supported - but I still want to make use of commitizen's version-updating inside of files. Otherwise, I will need to use a separate tool in conjunction, such as [bumpversion](https://github.com/c4urself/bump2version).

I want to be able to do `cz bump 1.0.1` for example (optional positional arg or `--manual <foo>` option would be fine).

A question on this though - whilst commitizen does not support dev versions, will it still be able to successfully parse a PEP440-compatible version that includes a dev version, for the sake of bumping? I.e. will commitizen be able to bump `1.0.1.dev0+foo.bar.baz` to `1.0.2` or `1.1.0` etc.?

And the ultimate related question - is there any planned support for dev versions?

jakob-keller:
## Description

Closes #435 

## Checklist

## Expected behavior

`cz bump 1.2.3` uses 1.2.3 as new version

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -183,6 +183,13 @@ data = {
                         "default": False,
                         "help": "retry commit if it fails the 1st time",
                     },
+                    {
+                        "name": "manual_version",
+                        "type": str,
+                        "nargs": "?",
+                        "help": "bump to the given version (e.g: 1.5.3)",
+                        "metavar": "MANUAL_VERSION",
+                    },
                 ],
             },
             {
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -102,7 +102,7 @@ class Bump:
         dry_run: bool = self.arguments["dry_run"]
         is_yes: bool = self.arguments["yes"]
         increment: Optional[str] = self.arguments["increment"]
-        prerelease: str = self.arguments["prerelease"]
+        prerelease: Optional[str] = self.arguments["prerelease"]
         devrelease: Optional[int] = self.arguments["devrelease"]
         is_files_only: Optional[bool] = self.arguments["files_only"]
         is_local_version: Optional[bool] = self.arguments["local_version"]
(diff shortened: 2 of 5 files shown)


jakob-keller:
Not sure, why these tests fail. Anything I need to fix?

commitizen/cli.py
```diff
@@ -183,6 +183,13 @@
                         "default": False,
                         "help": "retry commit if it fails the 1st time",
                     },
+                    {
+                        "name": "manual_version",
```
Lee-W, line 187: Should we make it `--manual-version`? 🤔

jakob-keller: I would not mind, but believe that makes no practical difference: 

1. decli/argparse normalises the argument name, i.e. bump.py uses `manual_version = self.arguments["manual_version"]` either way.
2. `cz bump --help` does not display the argument name, but the metavar value, i.e. `MANUAL_VERSION`.

Let me know, if you want me to change the PR.

woile: I prefer it like the proposal from this PR, as it is inline with the way changelog works: `cz changelog 1.2.3` would generate the changelog for that version. I would expect `cz bump 1.2.3` to bump the given version as well.

Lee-W: If that's the case, let's go with it 👍

Lee-W (approved):

woile:
Tests are not failing, it is the coverage report. But I'm not sure why it's decreasing, it doesn't look like it's decreasing

Lee-W:
> Tests are not failing, it is the coverage report. But I'm not sure why it's decreasing, it doesn't look like it's decreasing

Sometimes I just don't understand codecov 🤔

woile (approved):
Thanks for this nice feature!

Issue #590, opened 2022-09-17
Make empty commit


Dranaxel:
### Description

I wan to allow dev to create an empty commit with only cz message

### Possible Solution

Add an --allow-empty option

### Additional context

I work with a team using gitflow. I want to allow people to work freely on their feature branch and make a last conventionnally formated commit so we can merge-squash with the good message on develop branch

Lee-W:
Hi @Dranaxel , thanks for the feedback. We have a duplicate issue https://github.com/commitizen-tools/commitizen/issues/247 to track this one. Let's discuss there. Also, PRs are welcome. Thanks!

Dranaxel:
Hi @Lee-W thanks for your answer. I saw this issue but the status wasn't clear to me, sorry for the duplicate.
Does a PR with only allow-empty would fit ?

Lee-W:
Yep, I'm good with it 👍

PR #592, opened 2022-09-18
Feature/allow empty commit

Dranaxel:
## Description
Add allow to create empty commit

## Checklist

  - _One error on create commit message seem to be here before_
  - _Not needed_

## Expected behavior
Allow to invoke cz and create a commit without any change in it

## Additional context
Relates to #247 & #590

### DIFF
--- a/tests/commands/test_commit_command.py
+++ b/tests/commands/test_commit_command.py
@@ -38,7 +38,7 @@ def test_commit(config, mocker):
     success_mock = mocker.patch("commitizen.out.success")
 
     commands.Commit(config, {})()
-    success_mock.assert_called_once()
+   success_mock.assert_called_once()
 
 
 @pytest.mark.usefixtures("staging_is_clean")
--- a/tests/commands/test_commit_command.py
+++ b/tests/commands/test_commit_command.py
@@ -53,6 +53,26 @@ def test_commit_retry_fails_no_backup(config, mocker):
 
 
 @pytest.mark.usefixtures("staging_is_clean")
+def test_commit_allow_empty(config, mocker):
+    prompt_mock = mocker.patch("questionary.prompt")
+    prompt_mock.return_value = {
+        "prefix": "feat",
+        "subject": "user created",
+        "scope": "",
+        "is_breaking_change": False,
+        "body": "closes #21",
+        "footer": "",
+    }
+
+    commit_mock = mocker.patch("commitizen.git.commit")
+    commit_mock.return_value = cmd.Command("success", "", "", "", 0)
+    success_mock = mocker.patch("commitizen.out.success")
+
+    commands.Commit(config, {"allow_empty": True})()
+    success_mock.assert_called_once()
+
+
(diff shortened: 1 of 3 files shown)


tests/commands/test_commit_command.py
```diff
@@ -52,6 +52,26 @@ def test_commit_retry_fails_no_backup(config, mocker):
     assert NoCommitBackupError.message in str(excinfo.value)
 
 
+@pytest.mark.usefixtures("staging_is_clean")
+def test_commit_allow_empty(config, mocker):
+    prompt_mock = mocker.patch("questionary.prompt")
```
Lee-W, line 57: Overall the code looks good to me. Could you please explain a bit more on this test case? I'm a bit confused

Dranaxel:
@Lee-W Can you elaborate about what bother you in this test case ?

Lee-W:
I'm kinda not sure whether this is correctly tested and will need sometime to figure it out. Will try to check it this week

tests/commands/test_commit_command.py
Lee-W: Yep, this test does not quite make sense. As what we're trying is to make an empty commit, passing "prefix" and "subject" might not be what we want.

Dranaxel:
Hi, I think we have a misunderstanding here. When I say "Empty commit" I mean a commit with no changes in it but we still need a message inside.

tests/commands/test_commit_command.py
```diff
@@ -52,6 +52,26 @@ def test_commit_retry_fails_no_backup(config, mocker):
...
+    prompt_mock.return_value = {
+        "prefix": "feat",
+        "subject": "user created",
+        "scope": "",
+        "is_breaking_change": False,
+        "body": "closes #21",
+        "footer": "",
+    }
+
+    commit_mock = mocker.patch("commitizen.git.commit")
```
Lee-W, line 67: @Dranaxel Thanks for correcting me. Yes, you're right. In that case, we probably should assert whether this function is called with `--allow-empty`. This can be done by something like`assert_called_with`.

Dranaxel: Sorry i'm not sure about what you mean/how to use assert_called_with. Are you suggesting to assert the flag allow-empty was here when le command was launched ?

Lee-W: assert `--allow-empty` is actually passed into `commitizen.git.commit` when the operation above is executed

noirbizarre (approved):
Need rebase but I am OK with this one. I can definitely see use cases where you just want to force a commit without changes (like forcing rebuild to have updated dependencies).

Note: we might need the bump counter part for the same reason, aka. being able to force a release not containing commits eligible to changelog.

bearomorphism:
Commitizen has `cz commit -- --allow-empty` now. Do we still need this change?

Lee-W:
> Commitizen has `cz commit -- --allow-empty` now. Do we still need this change?

Yep, you're right. Thanks for reminding us. Let me close it

Issue #613, opened 2022-10-20
pre releases optionally start with a non-zero value


ckagerer:
### Description

I am C developer in microcontroller environment. So far I always used my own scripts/tools to increment version numbers and manually updated the changelog (if I didn't forget ;)).

I have been successfully using commitizen for my utilities in Phython for a few months now and would like to use that for embedded projects as well. But I currently have a problem that prevents me from using it. In our company the version is stored as four bytes in the embedded devices. MAJOR.MINOR[.BUGFIX][bBETA]. The BUGFIX and BETA digits are optional and are filled with zero if these digits should not be displayed by our ServiceTool.
With the BUGFIX everything still works perfectly (example: 0.3.0 -> 0.3.1 ✓).
However, with a PRERELEASE there is a conflict with our encoding (example: 0.3.0 -> 0.3.1b0 ⨯).
The only thing I would need to be able to use this great tool there would be an offset setting for prereleases so that the above example becomes this: 0.3.0 -> 0.3.1b1 ✓.

### Possible Solution

My proposed solution see PR #614 .
=> The introduction of a configuration for the offset. With this one can easily reconfigure the deviating behavior depending on the project.

woile:
Hey @ckagerer thanks for the input and the proposal. We recently introduced the `--devrelease` flag to bump. So you can do

```sh
cz bump --devrelease 1
```

Would you be able to work with that?

I think the docs could be improved, but here's the discussion that may be relevant for you: https://github.com/commitizen-tools/commitizen/issues/570#issuecomment-1243609467

ckagerer:
> Hey @ckagerer thanks for the input and the proposal. We recently introduced the `--devrelease` flag to bump. So you can do
> 
> ```shell
> cz bump --devrelease 1
> ```
> 
> Would you be able to work with that?
> 
> I think the docs could be improved, but here's the discussion that may be relevant for you: [#570 (comment)](https://github.com/commitizen-tools/commitizen/issues/570#issuecomment-1243609467)

Hi @woile,
Thanks for the quick suggestion of an alternative.
The devrelease option is unfortunately not quite suitable for my case. 

1. We need tags b{NUMBER} and not .dev{NUMBER}, because the tag is also used as folder name on our software archives. Here this style break would lead to confusion for many non-software developer colleagues. I could let an own script after `cz bump` correct the tag and the version number in the CHANGELOG, but I find this unsightly and very error-prone.
[...]

PR #614, opened 2022-10-20
feat(bump): support prereleases with a start offset

ckagerer:
This PR solves #613 

For special cases it may be necessary that prereleases cannot start with a 0, because the single characters are encoded in bytes. Here, consistent with the location for bugfixes, a 0 in the byte is interpreted as a release version. Therefore this commit introduces a generic option to change the start of the prerelease and also adapt it to such circumstances.

EXAMPLES

Previously:
0.3.0 -> PATCH beta-> 0.3.1b0

Now (with offset 0 == default):
0.3.0 -> PATCH beta-> 0.3.1b0

Now (with offset 1):
0.3.0 -> PATCH beta-> 0.3.1b1

## Checklist


Lee-W (commented):
Hi @ckagerer , thanks for your contribution and sorry for taking so long. The code looks good to me 🙌 But we might need your help to add some documentation and rebase the latest changes from the master branch.

ckagerer:
Hi @Lee-W,
Thanks for your feedback and initial review.

I have rebased the latest master branch to my feature branch.

Where do you expect the documentation to be?
More in the code or is there more documentation?

Best regards
Christian

Lee-W:
Hi @ckagerer thanks for the prompt update. I think we'll need to add the usage in https://github.com/commitizen-tools/commitizen/blob/master/docs/bump.md so that users know how to use

ckagerer:
@Lee-W
Sure, I will be happy if my PR is integrated :)

I have already extended the documentation in config.md.

Should I also include a command line argument, or is it enough just to use the configuration?

Lee-W:
@ckagerer It would be better if you could add the command line argument. Also, I notice we'll have to update the output of `cz bump --help` in https://github.com/commitizen-tools/commitizen/blob/master/docs/bump.md

ckagerer:
@Lee-W
Command line argument has been added and I have extended the documentation.
What do you think?

Lee-W:
Hi @ckagerer it looks great! I think our last step is to taking care of the CI failure.

ckagerer:
Hey @Lee-W 

Oh yes, I have ignored this so far because one of the unit tests fails locally on the master.

It seems to go wrong when installing with pip. Any experience with this?

I can try a poetry update to see if that helps.

ckagerer:
No, there are no updates at all.
It fails with the "poetry install" but I have no idea why.

Lee-W:
Do you have the error message?

ckagerer:
@Lee-W 

I think I read the message completely wrong :)

I ran the format script again and it was one line short. Now it should fit :)

PR #621, opened 2022-11-26
Feat 451 add commit args

SCrocky:
## Description
Added extra argument mechanic for `cz commit` command so that all arguments after a `--` will be treated as extra args.
Example: You can now sign commits with `cz c -- -S`

## Checklist

## Additional context
This PR should close [issue 451](https://github.com/commitizen-tools/commitizen/issues/451)


Lee-W (commented):
Sorry for late reviewing. The overall idea looks good to me! But we'll need doc and unit test for this one. After this is merged, I think we can have one branch removing the original signoff and send that PR to `v3` branch

SCrocky:
> Sorry for late reviewing. The overall idea looks good to me! But we'll need doc and unit test for this one. After this is merged, I think we can have one branch removing the original signoff and send that PR to `v3` branch

Sounds good, I'll get on the tests next.

woile:
Any update on this? We'd appreciate having the tests and rebasing 🙏🏻

SCrocky:
I have some time this weekend. I'll try to finish it then.

SCrocky:
Done reading, added both tests and doc.
@woile @Lee-W

commitizen/commands/commit.py
```diff
@@ -91,7 +91,11 @@ def __call__(self):
         signoff: bool = self.arguments.get("signoff")
 
         if signoff:
+            out.warn("signoff mechanic is deprecated, please use `cz commit -- -s` instead.")
```
Lee-W, line 94: @woile Do you think we should add a list somewhere so we won't forget to deprecate stuff in the next release?

Lee-W: @SCrocky In this implementation, if we're using the original `signoff` feature, we won't be able to use `extra_cli_args.`. Should we implement it as something like the following? (the warning is still needed)

```python
extra_cli_args = ""

if signoff:
    extra_cli_args = "-s "
```

commitizen/commands/commit.py
```diff
@@ -91,7 +91,11 @@ def __call__(self):
...
             c = git.commit(m, "-s")
+
+        if self.arguments.get("extra_cli_args"):
```
Lee-W, line 97: Do we need this if-else block? Could we handle the case in `git.commit()` when `extra_cli_args` is not provided?

docs/commit.md
```diff
@@ -4,13 +4,26 @@
...
     In particular pre-commit hooks coloring will be deactivated as discussed in [commitizen-tools/commitizen#417](https://github.com/commitizen-tools/commitizen/issues/417).
+
+
+### git options
+
+`git` command options that are not implemented by commitizen can be use via the `--` syntax for the `commit` command.
+The syntax separates commitizen arguments from `git commit` arguments by a double dash. This is the resulting syntax:
+```
+cz commit -commitizen-args -- -git-cli-args
+```
```
Lee-W, line 24:
```suggestion

```sh
cz commit -- <git-cli-args>

# e.g., cz commit --dry-run -- -a -S
```

```

File: commitizen/cli.py [hunk 4]
```diff
@@ -419,6 +420,23 @@ def main():
...
+                f"Invalid commitizen arguments were found: `{' '.join(unknown_args)}`. "
+                "Please use -- separator for extra git args"
+            )
+        # Raise error for extra-args before --
+        elif unknown_args[0] != "--":
+            pos = unknown_args.index("--")
+            raise InvalidCommandArgumentError(
+                f"Invalid commitizen arguments were found before -- separator: `{' '.join(unknown_args[:pos])}`. "
+            )
+        extra_args = " ".join(unknown_args[1:])
```
Lee-W, line 437: We might need to handle the case that the user use `--` without actually providing an argument. e.g., `cz commit --`

Lee-W (commented):
I am thrilled about this wonderful feature! I was able to spare some time to examine it. I have just provided a few comments and suggestions.

docs/commit.md
SCrocky: Can we keep `commitizen-args` ? It provides clarity IMO.

```sh
cz commit <commitizen-args> -- <git-cli-args>

# e.g., cz commit --dry-run -- -a -S
```

commitizen/commands/commit.py
SCrocky: Ah, I see it. Thanks for that.

Would this work for you?

```python
        if signoff:
            out.warn("signoff mechanic is deprecated, please use `cz commit -- -s` instead.")
            extra_args = "-s " + self.arguments.get("extra_cli_args")
        else:
            extra_args = self.arguments.get("extra_cli_args")

        c = git.commit(m, extra_args=extra_args)
```

commitizen/cli.py
SCrocky: Sure, how should we deal with it?

Warning or error?

docs/commit.md
Lee-W: Sure 🙂 Looks better 👍

commitizen/cli.py
Lee-W: Warning should be enough in this case 👍

commitizen/commands/commit.py
Lee-W: I might try something like

```python
        extra_args = self.arguments.get("extra_cli_args", "")
        if signoff:
            out.warn("signoff mechanic is deprecated, please use `cz commit -- -s` instead.")
            extra_args += " -s"

        c = git.commit(m, extra_args=extra_args)
```

Lee-W: But one thing I'm thinking is whether we should handle `cz commit -s -- -s`

SCrocky: I tested it and doesn't seem to be an issue with the git command:
```sh
git commit -s -s -m "test commit" 
[master (root-commit) b864659] test commit
 1 file changed, 0 insertions(+), 0 deletions(-)
```

But it is a bit messy. Warning or error?

Lee-W: If it doesn't cause an error, I'm good with either silently passing or warning.

mesaglio:
Is this wait somthing? I think `-S` [flag](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt) its good to use for security reasons.

SCrocky:
I've fixed rebased to master, fixed all the previous issues, added/edited tests and added some documentation.

@Lee-W ready for Round 2?

commitizen/git.py
```diff
@@ -98,13 +98,16 @@ def add(args: str = "") -> cmd.Command:
 
 
 def commit(
-    message: str, args: str = "", committer_date: str | None = None
+    message: str,
+    args: str = "",
+    extra_args: str = "",
```
Lee-W, line 103: Would it be possible for us to combine these 2? Not sure what's the difference between them

Lee-W (commented):
@woile @noirbizarre Might need some time to take a deeper look, but I think most of the parts are good.

commitizen/git.py
SCrocky: `extra_args` are arguments that we just want to pass to `git` transparently (like `-S`).

Having the `extra_args` argument separate helps with testing and clarity.

For example:`extra_args` has to start with "--" and contain at least one flag behind it. (e.g. `"-- -S"`). 

If this is contained within the `args` argument, then it becomes a little harder to know whether or not any extra args were passed.

> Would it be possible for us to combine these 2?

`git.commit` is only called with `extra_args` in [commands/commit.py `__call__` method](https://github.com/SCrocky/commitizen/blob/feat-451-add-commit-args/commitizen/commands/commit.py#L108) where it doesn't have any `args`.

So technically, if I'm not missing anything, I could just use the `args` argument to pass the extra_args. 

> Not sure what's the difference between them

I'll add a Docstring

Lee-W: > So technically, if I'm not missing anything, I could just use the args argument to pass the extra_args.

Yep, I think that's what we should do. Thanks for clarifying!

SCrocky: Done

Lee-W (approved):
Thanks, @SCrocky, for your contribution! This really helps a lot. Could you please make it ready for review?

@woile @noirbizarre I'm planning on merging it these days. Please let me know if you want to take a look

SCrocky:
> Thanks, @SCrocky, for your contribution! This really helps a lot.

No problem! I'm really glad to have finally finished it. :pray: 

> Could you please make it ready for review?

Is this addressed to me? I'm sorry, I'm not sure what the remaining steps are. Is there anything else I need to do?

Lee-W:
> > Thanks, @SCrocky, for your contribution! This really helps a lot.
> 
> No problem! I'm really glad to have finally finished it. 🙏
> 
> > Could you please make it ready for review?
> 
> Is this addressed to me? I'm sorry, I'm not sure what the remaining steps are. Is there anything else I need to do?

Ah, just found out I can do it as well

Issue #622, opened 2022-11-27
`cz ch <tag>` prints the whole changelog if `tag_format` is not set


gsalvatella:
### Description

Removing the `tag_format` config from the `pyproject.toml`  results in commitizen not being able to read the changelog for specific tags (it always prints the whole changelog). If `tag_format` is a required parameter it should be noted in the documentation. Otherwise `tag_format` should default to `$version`, since this is how it is initialized.

This can cause a lot of confusion for existing projects that are introducing commitizen in their workflows without using `cz init` and didn't add `tag_format` explicitly in the `pyproject.toml`. 

### Steps to reproduce

Give the following example setup

```bash
git init
cz init
git add . && git commit -m "feat: add pyproject"
cz bump
touch file1 && git add file1 && git commit -m "feat: add file1"
cz bump
```

Now check the changelog of the latest version

```bash
cz ch 0.2.0
cat CHANGELOG.md
## 0.2.0 (2022-11-27)

### Feat

- add file1
```

Remove `tag_format` line from `pyproject.toml`

Check again

### Current behavior

```bash
cz ch 0.2.0
cat CHANGELOG.md
## 0.2.0 (2022-11-27)

### Feat

- add file1

## 0.1.0 (2022-11-27)

### Feat

- add pyproject
```

### Desired behavior

```bash
cz ch 0.2.0
cat CHANGELOG.md
## 0.2.0 (2022-11-27)

### Feat

- add file1
```

### Environment

cz version: 2.37.0
python version: 3.10.8

PR #623, opened 2022-11-27
fix(changelog): allow rev range lookups without a tag format

gsalvatella:
The current default setting for `tag_format` is `None`. This is not a problem for the `bump` command, since the `normalize_tag` function defaults to `$version` when no `tag_format` is passed. However it is a problem for the `changelog` command, which seems to explicitly demand a `tag_format` in order to run a rev-range lookup. This creates issues like https://github.com/commitizen-tools/commitizen/issues/622.

Either a sane default needs to be set for `tag_format` or the restriction in `changelog` has to be uplifted. In this commit the latter has been chosen. A test is also implemented to check that `changelog` will always compute a rev range with the default tag format.

Fixes https://github.com/commitizen-tools/commitizen/issues/622

## Checklist

- [ ] Update the documentation for the changes -> Not needed

## Expected behavior
See https://github.com/commitizen-tools/commitizen/issues/622

See https://github.com/commitizen-tools/commitizen/issues/622

### DIFF
--- a/commitizen/commands/changelog.py
+++ b/commitizen/commands/changelog.py
@@ -137,7 +137,7 @@ class Changelog:
                 )
                 start_rev = self._find_incremental_rev(latest_tag_version, tags)
 
-        if self.rev_range and self.tag_format:
+        if self.rev_range:
             start_rev, end_rev = changelog.get_oldest_and_newest_rev(
                 tags,
                 version=self.rev_range,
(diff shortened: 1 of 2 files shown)


gsalvatella:
While this fixes the issue, I think that the root of the problem here lies in not defining a proper default for `tag_format`. Mostly all parameters should have a sane default. Otherwise they come to bite you down the line, as it happens with https://github.com/commitizen-tools/commitizen/issues/622

gsalvatella:
BTW the current failures in the tests haven't been introduced by this PR...

woile:
Thanks for the contribution. Yes, we try to have defaults, but sometimes with many contributors and needs things slip through. The problems have been fixed, could you rebase and push again?
Thank

gsalvatella:
The pipelines have been cancelled?

Lee-W:
It seems to be an error on testing. But I just retrigger it

gsalvatella:
:tada:

Issue #516, opened 2022-05-22
commitizen_emoji cannot write to CHANGELOG


adam-grant-hendry:
### Description

This is because `commitizen_emoji` uses unicode emoji characters, but `commitizen` does not specify the file encoding in `write_changelog()`, so it defaults back to `cp1252`:

https://github.com/commitizen-tools/commitizen/blob/79165119817cbd0aef25434a85dcbffa3a13acbb/commitizen/commands/changelog.py#L92

This could be fixed by specifying the `utf-8` encoding:

```python
with open(self.file_name, "w", encoding='utf-8') as changelog_file: 
   ...
```

### Steps to reproduce

```
> cz bump --changelog
bump: version 0.4.0 → 0.5.0
tag to create: 0.5.0       
increment detected: MINOR  

Traceback (most recent call last):
  File "C:\Program Files\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Program Files\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\hendra11\Code\external\cz-emoji\.venv\Scripts\cz.exe\__main__.py", line 7, in <module>
  File "C:\Users\hendra11\Code\external\cz-emoji\.venv\lib\site-packages\commitizen\cli.py", line 382, in main
    args.func(conf, vars(args))()
  File "C:\Users\hendra11\Code\external\cz-emoji\.venv\lib\site-packages\commitizen\commands\bump.py", line 215, in __call__
    changelog_cmd()
  File "C:\Users\hendra11\Code\external\cz-emoji\.venv\lib\site-packages\commitizen\commands\changelog.py", line 172, in __call__
    self.write_changelog(changelog_out, lines, changelog_meta)
  File "C:\Users\hendra11\Code\external\cz-emoji\.venv\lib\site-packages\commitizen\commands\changelog.py", line 103, in write_changelog
    changelog_file.write(changelog_out)
  File "C:\Program Files\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f389' in position 29: character maps to <undefined>
```

### Current behavior

See `Steps to reproduce`

### Desired behavior

Emojis can be logged to changelog

### Environment

- cz version: 2.27.0
- python version: 3.8.13
[...]

PR #629, opened 2022-12-03
Add Unicode Support

adam-grant-hendry:
Adds unicode support by allowing configurable encodings to be specified (defaults to `utf-8`).

## Checklist

Fixes: Issue #516

### DIFF
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -47,7 +47,7 @@ def find_increment(
 def update_version_in_files(
     current_version: str,
     new_version: str,
-    files: List[str],
+    files: list[str],
     encoding: str,
     *,
     check_consistency=False,
@@ -92,7 +92,7 @@ def _bump_with_regex(
     new_version: str,
     regex: str,
     encoding: str,
-) -> Tuple[bool, str]:
+) -> tuple[bool, str]:
     current_version_found = False
     lines = []
     pattern = re.compile(regex)
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -8,7 +8,7 @@ from string import Template
 from commitizen.defaults import MAJOR, MINOR, PATCH, bump_message
 from commitizen.exceptions import CurrentVersionNotFoundError
 from commitizen.git import GitCommit, smart_open
-from commitizen.version_schemes import DEFAULT_SCHEME, VersionScheme, Version
+from commitizen.version_schemes import DEFAULT_SCHEME, Version, VersionScheme
 
 
 def find_increment(
@@ -45,7 +45,12 @@ def find_increment(
 
 
 def update_version_in_files(
-    current_version: str, new_version: str, files: list[str], *, check_consistency=False
+    current_version: str,
+    new_version: str,
+    files: List[str],
+    encoding: str,
+    *,
+    check_consistency=False,
 ) -> None:
     """Change old version to the new one in every file given.
 
@@ -62,7 +67,11 @@ def update_version_in_files(
             regex = _version_to_regex(current_version)
 
         current_version_found, version_file = _bump_with_regex(
-            filepath, current_version, new_version, regex
+            filepath,
+            current_version,
+            new_version,
+            regex,
+            encoding,
         )
 
         if check_consistency and not current_version_found:
@@ -73,17 +82,21 @@ def update_version_in_files(
             )
 
         # Write the file out again
-        with smart_open(filepath, "w") as file:
+        with smart_open(filepath, "w", encoding=encoding) as file:
             file.write(version_file)
 
 
 def _bump_with_regex(
-    version_filepath: str, current_version: str, new_version: str, regex: str
-) -> tuple[bool, str]:
+    version_filepath: str,
+    current_version: str,
+    new_version: str,
+    regex: str,
+    encoding: str,
+) -> Tuple[bool, str]:
     current_version_found = False
     lines = []
     pattern = re.compile(regex)
-    with open(version_filepath, "r") as f:
+    with open(version_filepath, "r", encoding=encoding) as f:
         for line in f:
             if pattern.search(line):
                 bumped_line = line.replace(current_version, new_version)
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -5,7 +5,7 @@ import re
 from collections import OrderedDict
 from string import Template
 
-from commitizen.defaults import MAJOR, MINOR, PATCH, bump_message
+from commitizen.defaults import MAJOR, MINOR, PATCH, bump_message, encoding
 from commitizen.exceptions import CurrentVersionNotFoundError
 from commitizen.git import GitCommit, smart_open
 from commitizen.version_schemes import DEFAULT_SCHEME, Version, VersionScheme
@@ -48,9 +48,9 @@ def update_version_in_files(
     current_version: str,
     new_version: str,
     files: list[str],
-    encoding: str,
     *,
     check_consistency=False,
+    encoding: str = encoding,
 ) -> None:
     """Change old version to the new one in every file given.
 
@@ -71,7 +71,7 @@ def update_version_in_files(
             current_version,
             new_version,
             regex,
-            encoding,
+            encoding=encoding,
         )
 
         if check_consistency and not current_version_found:
@@ -91,7 +91,7 @@ def _bump_with_regex(
     current_version: str,
     new_version: str,
     regex: str,
-    encoding: str,
+    encoding: str = encoding,
 ) -> tuple[bool, str]:
     current_version_found = False
     lines = []
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -49,7 +49,7 @@ def update_version_in_files(
     new_version: str,
     files: list[str],
     *,
-    check_consistency=False,
+    check_consistency: bool = False,
     encoding: str = encoding,
 ) -> None:
     """Change old version to the new one in every file given.
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -213,7 +213,9 @@ def parse_title_type_of_line(value: str) -> str | None:
     return m.groupdict().get("title")
 
 
-def get_metadata(filepath: str, scheme: VersionScheme = Pep440) -> dict:
+def get_metadata(
+    filepath: str, scheme: VersionScheme = Pep440, encoding: str = "utf-8"
+) -> dict:
     unreleased_start: int | None = None
     unreleased_end: int | None = None
     unreleased_title: str | None = None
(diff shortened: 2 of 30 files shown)


adam-grant-hendry:
@Lee-W @woile Please review when you get a chance. Thanks!

.github/workflows/pythonpackage.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
-        platform: [ubuntu-latest, macos-latest, windows-latest]
+        platform: [ubuntu-20.04, macos-latest, windows-latest]
```
Lee-W, line 10: Is there any reason we need to pin it to a certain version?

Lee-W (approved):
@adam-grant-hendry Thanks for the update! Just one minor discussion needed on my side.

@woile I'm planning on merging this next week. Let me know if you need more time to take a deeper look. Thanks!

.github/workflows/pythonpackage.yml
adam-grant-hendry: No worries and thanks for reviewing! The CI workflow `actions/setup-python@v4` kept failing on `ubuntu-latest` with:

`Error: Version 3.6 with arch x64 not found`

This error was reported in

https://github.com/actions/setup-python/issues/544 

and the recommended fix is to hard code to Ubuntu 20.04 if Python 3.6 is needed since they no longer are supporting Python 3.6.

I figured this could be acceptable until we sunset Python 3.6.

Lee-W: ah, got it. Thanks for replying! I'm now encountering the same issue on my own PR haha

adam-grant-hendry: Ya, it just started happening. The team didn’t make tarballs of 3.6 for the new distro and chose not to add them since 3.6 is already deprecated.

Side note, it’s usually best to specify specific versions rather than latest for this reason. You get more repeatable results that way, so it might be good to do the same for the other OS’s.

commitizen/defaults.py
```diff
@@ -65,6 +67,7 @@ class Settings(TypedDict, total=False):
     "update_changelog_on_bump": False,
     "use_shortcuts": False,
     "major_version_zero": False,
+    "encoding": "utf-8",
```
woile, line 70: shouldn't this be

```py
"encoding": encoding,
```

woile:
LGTM, the only thing missing is documentation: https://commitizen-tools.github.io/commitizen/config/

Ideally, I think the `cz init` should ask for the encoding **if** you are on windows, this would make windows users aware of potential encoding issues

commitizen/defaults.py
adamgranthendry: Good point. Should we also change

```python
"name": "cz_conventional_commits"
```

to

```python
"name": name
```

adam-grant-hendry: Good point. Should we also change

```python
"name": "cz_conventional_commits"
```

to

```python
"name": name
```

Lee-W:
It seems CI failed on windows 🤔

adam-grant-hendry:
@Lee-W I didn't push any commits yesterday, but it is showing that I did through `Unverified` commits. By chance, was this you?

Lee-W:
Hi @adam-grant-hendry , yes I tried to rebase the master branch to this one to see if the latest update fix this branch

woile:
Is this still valid? Please rebase 🙏🏻

adam-grant-hendry:
@woile @Lee-W Apologies for taking so long: I've been away for quite some time. My gpg signature keys are outdated, but the commits are by me.

`mypy` is still unhappy with line 6 of `out.py` (See: https://github.com/python/typeshed/issues/3049). I added a `#type: ignore`, but Python 3.7 reads this as an unused type ignore (See: https://github.com/commitizen-tools/commitizen/actions/runs/5491623964/jobs/10008334198#step:5:251), so I'm stuck.

Only remaining item is to add documentation. I added `encoding` as a keyword argument to `changelog.py/get_metadata` so as to not break backwards-compatibility with the added `scheme` argument.

adam-grant-hendry:
@woile @Lee-W Everything is good now. Could you please re-review the changes and make sure I added everything you requested?

Lee-W:
@adam-grant-hendry Sure! I'll take a look these days.

commitizen/bump.py
```diff
@@ -45,7 +45,12 @@ def find_increment(
 
 
 def update_version_in_files(
-    current_version: str, new_version: str, files: list[str], *, check_consistency=False
+    current_version: str,
+    new_version: str,
+    files: list[str],
+    encoding: str,
+    *,
```
Lee-W, line 52: If possible, I'd like to keep newly added argument as keyword only argument
```suggestion
    *,
    encoding: str,
```

commitizen/changelog.py
```diff
@@ -213,7 +213,9 @@ def parse_title_type_of_line(value: str) -> str | None:
     return m.groupdict().get("title")
 
 
-def get_metadata(filepath: str, scheme: VersionScheme = Pep440) -> dict:
+def get_metadata(
+    filepath: str, scheme: VersionScheme = Pep440, encoding: str = "utf-8"
```
Lee-W, line 217: Is the default `"utf-8"` required here? if so, should we add this default elsewhere as well?

Lee-W (commented):
Hi @adam-grant-hendry , sorry for the late review. Overall the changes are good, but I think we should drop commit `3430edba7a5b1b4f59e5b7aa48e03bd15ccc32a4` and `9af6c748eb1084c5239e8d8058e7eddd59abbfb5` which downgrades the version of our GitHub actions workflow. Also, I notice some `open` and `smart_open` miss this `encoding` feature. Is this intentional? Or just accidentally missed? Thanks!

adam-grant-hendry:
> Hi @adam-grant-hendry , sorry for the late review. Overall the changes are good, but I think we should drop commit `3430edba7a5b1b4f59e5b7aa48e03bd15ccc32a4` and `9af6c748eb1084c5239e8d8058e7eddd59abbfb5` which downgrades the version of our GitHub actions workflow. Also, I notice some `open` and `smart_open` miss this `encoding` feature. Is this intentional? Or just accidentally missed? Thanks!

No worries at all. Thank you for your review!

1. Yes, good find: we should definitely drop those commits then
2. That was certainly an accidental miss on my part. I’ll review, make changes, and push up the fixes

I probably can’t get to this today, but I can start tomorrow. I’ll see if I can request your re-review by Monday morning.

Thanks again!

commitizen/changelog.py
adam-grant-hendry: Since `scheme` is a keyword parameter, a non-keyword parameter cannot come after it in the parameter list. I added `"utf-8"` so as not reorder the parameters and break backwards compatibility.

Instead, I can import `defaults` and use `encoding` defined there:

```python
from commitizen.defaults import encoding
...

def get_metadata(
    filepath: str, scheme: VersionScheme = Pep440, encoding: str = encoding
) -> dict:
    ...
```

Python still [defaults to the 8-bit 1252 character set encoding on Windows](https://docs.python.org/3/using/windows.html#utf-8-mode). Making `"utf-8"` the default encoding is a good choice and shouldn't be problematic (see `Code that may break` in [PEP 528](https://peps.python.org/pep-0528/) and [PEP 529](https://peps.python.org/pep-0529/) for impetus).

Files opened in binary mode [return contents as `bytes` objects without any decoding](https://docs.python.org/3/library/functions.html#open), so `encoding` wasn't added to those uses of `open` (nor for trivial cases like `commands/init.py/is_python_poetry()`.

I'll add `.encoding` everywhere else where it was missed.

commitizen/bump.py
adam-grant-hendry: Resolved in commit `e9d73c5f`.

commitizen/changelog.py
adam-grant-hendry: Resolved in commit `e9d73c5f`.

adam-grant-hendry:
@Lee-W I made all the changes discussed. Please re-review this PR at your earliest convenience. Thank you!

Lee-W:
@adam-grant-hendry Sure! I'll try to take a look this weekend

Lee-W (approved):
Overall, I'm good with this PR. But we missed a few places missed. Wondering is it designed this way or just missed. https://github.com/commitizen-tools/commitizen/blob/114501de6c9a0a763b291f2ec3b3251522049d94/commitizen/commands/init.py#L38, https://github.com/commitizen-tools/commitizen/blob/114501de6c9a0a763b291f2ec3b3251522049d94/hooks/prepare-commit-msg.py#L63.

But, yep looks like they might be something not affected 🤔

I'm planning on merging this next week. @woile Please let me know if you want to take a deeper looks 🙂

Lee-W:
@adam-grant-hendry ah, just notice a conflict. might need your help to resolve it.

adam-grant-hendry:
> ...But we missed a few places missed. Wondering is it designed this way or just missed.
> 
> https://github.com/commitizen-tools/commitizen/blob/114501de6c9a0a763b291f2ec3b3251522049d94/commitizen/commands/init.py#L38
> 
> ,
> https://github.com/commitizen-tools/commitizen/blob/114501de6c9a0a763b291f2ec3b3251522049d94/hooks/prepare-commit-msg.py#L63
> 
> .
> But, yep looks like they might be something not affected 🤔
> 
I purposefully left these out as I considered them "trivial" cases. By that I mean the `"pyproject.toml"` case is only looking for the `"[tool.poetry]"` table heading, which should generally not be affected by whether the file is opened in `utf-8` mode or not. As for `"/dev/tty"`, there is no equivalent to `"/dev/tty"` on Windows machines, so this would only run properly from a Linux environment, in which case the encoding is already `utf-8` by default.

Lee-W:
Sounds good 👍 Just want to confirm it. I think we're good to merge it after resolving the conflict. Thanks!

woile (approved):
LGTM as well. Thanks!

commitizen/bump.py
```diff
@@ -45,7 +45,12 @@ def find_increment(
 
 
 def update_version_in_files(
-    current_version: str, new_version: str, files: list[str], *, check_consistency=False
+    current_version: str,
+    new_version: str,
+    files: list[str],
+    *,
+    check_consistency=False,
```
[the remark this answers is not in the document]
adam-grant-hendry: Resolved in `69c90`

adam-grant-hendry:
> @adam-grant-hendry ah, just notice a conflict. might need your help to resolve it.

I had to pull in the latest commits to `master` and rebase. It's all good to go now!

Lee-W:
Many thanks! Just merged

Issue #558, opened 2022-08-16
tag message for annotated_tag option


frerksaxen:
### Description
Currently, with `cz bump --annotated-tag`, a tag will be created with the tag message being like the tag itself, just the new version. It's basically `git tag -a -m <version> <version>`.
However, git provides the possibility of adding elaborate messages to the annotated tag: `git tag -a -m <message> <version>`.
It would be great if we could add custom tag messages that could include:
- the changelog
- custom messages
- author and date

### Possible Solution
Maybe adding custom messages:
`cz bump --annotated-tag-msg "This new version will improve your life."`
Maybe adding the changelog:
`cz bump --annotated-tag-changelog`
Maybe adding author and date:
`cz bump --annotated-tag-credentials`

And maybe everything combined:
`cz bump --annotated-tag-msg "This new version will improve your life." --annotated-tag-changelog --annotated-tag-credentials`

might result in the git tag:
**2.32.0**
with the tag message:
```
This new version will improve your life.
Changelog:
2.32.0 (2022-08-18)
Feat:
- annotated-tag-msg <message> adds custom message to annotated git tag
- annotated-tag-changelog adds changelog to annotated git tag
- annotated-tag-credentials adds author and date to annotated git tag
Author/Date:
BestProgrammerEver/2022-08-18
```

### Additional context

This is related to the following pull request: https://github.com/commitizen-tools/commitizen/pull/272

Lee-W:
Hi @frerksaxen , thanks for the feedback. I'm afraid I did not get the point. Could you please add some examples so that we could take a look?

frerksaxen:
@Lee-W Sure. I changed the original Description. Hope it's much clearer now.

LuisHenri:
I agree with the idea of adding a message apart from the tag_name

Lee-W:
Hi, I'm good with the high-level idea. But might need some more discussion on how the detail would be implemented

LuisHenri:
@Lee-W Hi, I forked it and created the PR! :) 
#631

Lee-W:
Sounds great! I'll take a deeper look when I have time. Thanks!

yajo:
FWIW you can look in this page how that'd look in Gitlab: https://gitlab.gnome.org/GNOME/geary/-/tags

Annotated screenshot about annotated tags:

![imagen](https://github.com/commitizen-tools/commitizen/assets/973709/63ebbb12-1032-4dac-bbad-7db4e4e71995)

[...]

PR #631, opened 2022-12-04
Add custom tag message

LuisHenri:
## Description
According to #558, I implemented a custom tag message for annotated tags.
Right now, the annotated tag is given as "tag_name" with its message as "tag_name" as well. It might be nice to have custom tag messages, for example, to add the features added on that tag.
The idea is a new argument to bump with the tag_message,

```bash
cz bump --annotated-tag-message "a message"
cz bump -atm "a message"
```

## Checklist

## Expected behavior
Running either the following commands will create an annotated tag with "a message" as the message of the tag.

```bash
cz bump --annotated-tag-message "a message"
cz bump -atm "a message"
```

<!-- Steps to reproduce the behavior:
1. Run `cz bump -atm "a message"`
2. Run ` git tag -l --format='%(contents:subject)' <the-created-tag-name>`
3. Output should be `'a message'`

## Additional context
#558


Lee-W:
>  Should I name the argument as --annotated-tag-message or as --annotated-tag-msg to keep it short?

I like `--annotated-tag-message` a bit more. 

> For some reason, the Pipeline is not working correctly. I see that this is the case for other PRs as well. Is that really the case @Lee-W ?

This should have been solved. Let me rebase your branch and see how it works.

Lee-W:
Hi @LuisHenri ,`tests/commands/test_bump_command.py::test_bump_changelog_command_commits_untracked_changelog_and_version_files` failed even after rebasing. I might need your help to take a look at this issue. Thanks!

LuisHenri:
> Hi @LuisHenri ,`tests/commands/test_bump_command.py::test_bump_changelog_command_commits_untracked_changelog_and_version_files` failed even after rebasing. I might need your help to take a look at this issue. Thanks!

Hey @Lee-W,
I fixed it. Seems like one of the tests was trying to run `bump --annotated` instead of `bump --annotated-tag`.

commitizen/git.py
```diff
@@ -82,13 +82,16 @@ def from_line(cls, line: str, inner_delimiter: str) -> "GitTag":
...
+    tag: str, annotated: bool = False, signed: bool = False, msg: Optional[str] = None
+) -> cmd.Command:
     _opt = ""
     if annotated:
         _opt = f"-a {tag} -m"
     if signed:
         _opt = f"-s {tag} -m"
-    c = cmd.run(f"git tag {_opt} {tag}")
+
+    c = cmd.run(f'git tag {_opt} "{tag if _opt == "" or msg is None else msg}"')
```
Lee-W, line 94: Does that mean if we provide `msg` then we won't have `tag?`

LuisHenri: No, the parameter `tag` is mandatory. So we will have a tag nonetheless.
If we provide a `msg`, it will all depend on whether we defined the parameters `annotated` or `signed`. Because if `msg` is defined, the `or msg is None` part will be False and it will depend on `_opt`, which depends on `annotated` or `signed`.

Assuming we HAVE defined `msg` and `tag=MY_TAG`:
1. If we define `annotated = False` or `signed = False`, `_opt` will remain an empty string.
   - This means the condition `_opt == ""` will be True, so: `git tag "MY_TAG"`
2. If we define either `annotated = True` or `signed = True`, `_opt` won't be an empty string anymore.
   - This means the condition `_opt == ""` will be False. Since `msg` is also defined, `msg is None` will be False as well. So, the resulting command will be, for example: `git tag -a MY_TAG -m {msg}`

LuisHenri:
@Lee-W is something else needed still?

Lee-W:
Hi @LuisHenri , sorry for the late review; I'm out of bandwidth these days. But I planned to take a look at a few PRs tomorrow. If you can rebase this PR and mention me, I'll take a look tomorrow. Thanks!

LuisHenri:
@Lee-W Sorry for the delay. I'm done.

Lee-W:
Hi @LuisHenri , instead of merging main branch, we prefer rebasing. You can do so through `git rebase -i master`

LuisHenri:
@Lee-W Thanks for the advice! What about now? :)

Lee-W:
@LuisHenri  perfect! I'll try to take a deeper look this weekend.

tests/utils.py
```diff
@@ -55,8 +55,8 @@ def get_current_branch() -> str:
     return c.out
 
 
-def create_tag(tag: str):
-    c = git.tag(tag)
+def create_tag(tag: str, message: str | None = None):
+    c = git.tag(tag, annotated=(message is not None), msg=message)
```
Lee-W, line 59:
```suggestion
def create_tag(tag: str, message: str | None = None) -> None:
    c = git.tag(tag, annotated=(message is not None), msg=message)
```

Lee-W (approved):
Hi @LuisHenri , thanks for your contribution. the change looks good to me!

@woile I'm planning on merging this one these days. Let me know if you want to take a deeper look

woile:
Is it possible to add a message to a light tag? If not, should we throw an error to inform the user?

Lee-W:
@woile According to https://git-scm.com/book/en/v2/Git-Basics-Tagging, it seems we're not able to do that for a light tag

commitizen/git.py
```diff
@@ -82,13 +82,16 @@ def from_line(cls, line: str, inner_delimiter: str) -> GitTag:
...
+    tag: str, annotated: bool = False, signed: bool = False, msg: str | None = None
+) -> cmd.Command:
     _opt = ""
     if annotated:
         _opt = f"-a {tag} -m"
     if signed:
         _opt = f"-s {tag} -m"
-    c = cmd.run(f"git tag {_opt} {tag}")
+
+    c = cmd.run(f'git tag {_opt} "{tag if _opt == "" or msg is None else msg}"')
```
Lee-W, line 94: @woile do you mean we should add a warning here to let the users know if they're using this feature, they're actually creating annotated tag?

>  To create a lightweight tag, don’t supply any of the -a, -s, or -m options, just provide a tag name:

https://git-scm.com/book/en/v2/Git-Basics-Tagging

not sure whether my understanding is correct

woile: I see, maybe a comment there explaining that. I didn't know that if you provide a message it automatically becomes an annotated tag.

Lee-W: @woile I just added the comment and resolved the conflict. Could you please take a look? I think we're pretty close to merge this one :)

cc @noirbizarre

frerksaxen:
@Lee-W `git tag -m` limits the tag annotation message to a one liner, correct? To have multiple lines you would need to allow multiple `-m` messages or provide a `-F` file. Or is it possible to line break with \n within the message string?

Lee-W:
> @Lee-W `git tag -m` limits the tag annotation message to a one liner, correct? To have multiple lines you would need to allow multiple `-m` messages or provide a `-F` file. Or is it possible to line break with \n within the message string?

Something like the following works

```sh
git commit -m "first line
second line
third line"
```

PR #632, opened 2022-12-05
feat(plugins): Switch to an importlib.metadata.EntryPoint-based plugin loading

noirbizarre:
## Description
This PR is proposal implementation for #495

It switches to an `importlib.metadata.EntryPoint`-based plugin loading instead of pattern matching on the module name.
Plugins are now loaded using the `commitizen.plugin` entrypoint while legacy plugins are not loaded anymore but a warning is raised for each legacy plugin seen.

BREAKING CHANGE: Plugins are now exposed as `commitizen.plugin` entrypoints

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes

## Expected behavior

Internal plugins are exposed as `commitizen.plugin` entrypoint. and seen as plugins.
External plugins are seen too as soon as they are exposed as `commitizen.plugin` entrypoint.
Legacy plugins are not loaded and a warning is raised for each legacy plugin detected.

## Steps to Test This Pull Request
 
Try to use any internal plugin without changes in configuration: should work
Try a legacy plugin: KO
Try a plugin migrated to the `commitizen.plugin`: should work

## Additional context

I chose not to have a deprecation warning on this because:
- there was issues with the `cz_` naming
- handling plugins with both `discover_this` and entrypoint expose raise a bit the complexity

But I can switch to this if required:
- register legacy plugins first while raising a deprecation warning
- register entrypoints based plugins, overwritting those registered by legacy pattern
- Add a deprecation warning in the documentation with a target version for support removal

Fixes #495

PR #633, opened 2022-12-05
build(python): Sunset Python 3.6 and support Python 3.11

### LINKED ISSUE #607: [FEAT]: 🌄 Sunset `Python 3.6`
### Description

`Python 3.6` reached its [end of life](https://peps.python.org/pep-0494/) on 2021-12-23 (see also [PEP 602](https://peps.python.org/pep-0602/)).

In attempting to update `pyproject.toml` dependencies, some packages are requiring `Python >= 3.7`, which may warrant retiring `3.6`.

### Possible Solution

Please add a `DeprecationWarning` for `Python 3.6` to give existing users time to update their code bases and issue an agreed-upon sunset date for `Python 3.6` support.

### Additional context

See [PR #605](https://github.com/commitizen-tools/commitizen/pull/605) for discussion of some CI workflows that are failing due to outdated dependencies.

### Additional context

_No response_

noirbizarre:
## Description

**Proposal:** This PR sunsets Python 3.6 and adds supports Python 3.11 given Python 3.6 is been sunsetted 1 year ago and lately buils are not running anymore on Python 3.6 (see https://github.com/commitizen-tools/commitizen/pull/631/checks as example).

**BREAKING CHANGE:** Python 3.6 is not supported anymore 

## Checklist

## Expected behavior

New package don't advertise support for Python 3.6 anymore
CI stop running on Python 3.6
Python 3.11 support is now tested in CI
CI is green again

## Additional context

Fixes #607 

I permitted myself to submit this PR because builds are now failing on Python 3.6 for new PR like mine: #632

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+version: 2
+updates:
+  -
+    # Maintain dependencies for GitHub Actions
+    package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+    labels:
+      - dependencies
+    commit-message:
+      prefix: "ci(actions)"
+      include: "scope"
+  -
+    # Maintain python dependencies
+    package-ecosystem: pip
+    directory: /
+    schedule:
+      interval: daily
+    labels:
+      - dependencies
+    commit-message:
+      prefix: "build(poetry)"
+      include: "scope"
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -6,15 +6,15 @@ jobs:
   python-check:
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         platform: [ubuntu-20.04, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
--- a/.gitignore
+++ b/.gitignore
@@ -107,6 +107,3 @@ venv.bak/
 .idea
 .vscode/
 *.bak
-
-# build
-poetry.lock
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import sys
 from functools import partial
+from types import TracebackType
 from typing import List
 
 import argcomplete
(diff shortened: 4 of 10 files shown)


woile:
Hey thanks a lot for this PR. I'm not sure why is the CI failing tbh.
I'm thinking maybe this could be part of your other PR, so we can have just a single breaking change. What do you think?

> By the way, I was surprised the poetry.lock file is not commited in the repository. Is there a specific reason why ?

My understanding (might be outdated), is that locks are for deployable applications (like an app). Without a lock file, you may get into issues (because semver is not perfect, right? a new feature in another library may break us). This means that contributors of this projects may get into trouble, allowing us to fix the problem. In short, is another way for us to find problems. Does it make sense?

noirbizarre:
Yes, I think another PR is required just to fix the CI (I will try to have a look tonight, seems related to the `exceptiongroup` monkeypatching.

For the lock file, I think it would solve the dev/test env stability part. For the library publication, I think this is depending on how the it done by `poetry`: either it is packaging the dependencies as expressed in `pyproject.toml` andthere shouldn't be any issue, either it is pinning dependencies to their locked version and it would justify not having the lock file.
I will take a look and probably have a PR for that but I think this is the first case.

noirbizarre:
I integrated #634 in this one as multiple fixes were needed to make the CI green again:
- removes the not support anymore Python 3.6 jobs
- ensure build are reproducibles by commiting the poetry.lock
- upgrade some dependencies (basically pytest plugins and `black`
- fix the failing tests (side-effects of pytest upgrades)

I didn't touch the dependencies that where not requiring an upgrade (can be done there or in another PR, as you want).

I added a `dependabot.yml` configuration to receive pull requests updating workflows and poetry.lock ensuring we test and discover new versions issues as soon as they are available.

This should fix the CI ✅  and makes it more resilient to future dependencies update by not staking them.

Lee-W (approved):

woile:
I'm good with it. But I'm just wondering if we want to have this, and then the new plugin system creating multiple Major versions. We'd be at commitizen v4 at that point 😅 I guess I have to break my sentimental attachment to Major versions 🤣

woile:
You can stake them yes, but we don't have any policies in this repo. Normally I would create a new branch for breaking changes, and send everything there, like you suggest. But for this case, it might make sense, v3 to drop 3.6, and v4 for the new plugin system. @Lee-W ?

woile (approved):

woile:
Only thing, I would update this commit:

```diff
- BREAKING CHANGE: Python 3.6 is not supported anymore
+ BREAKING CHANGE: Python 3.6 is not officially supported anymore. Please migrate from 3.6 to 3.7 or greater.
```

noirbizarre:
👆🏼  Updated 👍🏼

noirbizarre:
Note: I have multiple PRs to submit, some to discuss... 😅 
So having a 3.x with updated deps and python support, new extension system + a bunch of other features could be great Major !
(From a lib/cli consumer point of view, I prefer a single release with 3 breaking changes to handle than 3 consecutives releases with breaking changes, it is easier to understand, there is one before and one after, no intermediate states to support)

Lee-W:
I'm actually good with either way. But if we're planning to merge a few breaking changes in a short period, `v3` branch might be a better approach.

Lee-W:
I'm actually good with either way. But if we're planning to merge a few breaking changes in a short period, `v3` branch might be a better approach.

Lee-W:
I just created a `v3` branch. Let me merge it to that branch first 👍

Issue #292, opened 2020-11-02
Add ability to run arbitrary hooks pre/post version bump


janw:
## Description
To use the currently available hook facilities, it is necessary to create a custom cz_ adapter class to use hooks during the version bump / changelog generation. I'd like to be able to run arbitrary hooks (i.e. a shell script or other external executable) from a simple configuration file section. In my specific use-case I'd like for the hooks to be run after the changelog generation and before the version bump in case of `cz bump --changelog` (see also #290), but I can imaging the following hook locations to be useful:

1. before changelog
2. after changelog and before commit and tag
3. after tag

## Possible Solution
Implement hook facilities at the aforementioned locations. They should be configurable via the configuration file for example as follows, allowing for multiple commands to be run at a given location

```toml
[[tool.commitizen.hook_begin]]
command = "./update-something.sh"

[[tool.commitizen.hook_begin]]
command = "./do-another-thing.sh"

[[tool.commitizen.hook_before_commit]]
command = "./build-something-based-on-changelog.sh"

[tool.commitizen.hook_end]
command = "./push-artifacts.py"
```

Given commands should run with useful execution context, such as current and future version as arguments and/or environment variables.

---

I'd be happy to implement this as well. Cheers.

Woile:
This is actually super interesting. 
I wonder if it doesn't introduce any security issues.
Each of those functions would be called with the arguments, right? Like if you execute the hook after the changelog generation, your script should receive the changelog generated.

I actually think that these hooks would be great candidates for plugins, as proposed in #287

janw:
I guess you're right, it does have some security implications when running arbitrary code is allowed. But I'd argue that the commands that are run are inherently documented in the config file (i.e. it's clear for every commit _what_ scripts are going to be run) and that somewhat alleviates these implications.

In a PoC that I'm preparing (just for myself, even if you'd decide not to merge it), I went for providing the arguments via environment variables, as it allows for more context to be passed to the script. Also I reduced the lists to a more simple syntax

```toml
[tool.commitizen]
# ...
hooks_pre_changelog = [
  "./update-something.sh",
  "./do-another-thing.sh"
]
hooks_pre_bump = [
  "./build-something-based-on-changelog.sh"
]
hooks_post_bump = [
  "./push-artifacts.py"
]
```

In case of the "pre-bump" hook I think these variables would be useful in the environment of the executed script/command:

```bash
CZ_BUMP_IS_INITIAL= 
[...]

PR #644, opened 2022-12-21
feat(bump): added support for running arbitrary hooks during bump

skoef:
## Description

To make commitizen integrate even better, new configuration keys hooks_pre_bump and hooks_post_bump were added to allow to run arbitrary commands prior to and right after running bump.

I've backported the work done by @janw and slightly adjusted it:
- updated the patch to apply against the latest version of commitizen
- run `hooks_pre_bump` hooks *after* the versions have been updated but *before* they are committed and tagged, so additional documentation can be generated with the new version or other generators can be run and staged.
- I couldn't think of a usecase why you would want to hook into the process of generating the changelog so I completely removed this part.

## Checklist

1. Add something like this in your commitizen config file:
```toml
[tool.commitizen]
hooks_pre_bump = [
  "touch foo.bar",
  "git add foo.bar"
]
```
2. Run `cz bump` and notice the `Running hook` output
3. Run `git show HEAD` to see `foo.bar` is part of the latest release

## Additional context
Closes: #292
Thanks to @janw to do the heavy lifting here.

### DIFF
--- a/commitizen/hooks.py
+++ b/commitizen/hooks.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+from commitizen import cmd, out
+from commitizen.exceptions import RunHookError
+
+
+def run(hooks, _env_prefix="CZ_", **env):
+    if isinstance(hooks, str):
+        hooks = [hooks]
+
+    for hook in hooks:
+        out.info(f"Running hook '{hook}'")
+
+        c = cmd.run(hook, env=_format_env(_env_prefix, env))
+
+        if c.out:
+            out.write(c.out)
+        if c.err:
+            out.error(c.err)
+
+        if c.return_code != 0:
+            raise RunHookError(f"Running hook '{hook}' failed")
+
+
+def _format_env(prefix: str, env: dict[str, str]) -> dict[str, str]:
+    """_format_env() prefixes all given environment variables with the given
+    prefix so it can be passed directly to cmd.run()."""
+    penv = dict()
+    for name, value in env.items():
+        name = prefix + name.upper()
+        value = str(value) if value is not None else ""
+        penv[name] = value
+
+    return penv
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -19,6 +19,10 @@ style = [
     ["pointer", "reverse"],
     ["question", "underline"]
 ]
+pre_bump_hooks = [
+    "scripts/generate_documentation.sh"
+]
+post_bump_hooks = ["scripts/slack_notification.sh"]
 
 [tool.black]
 line-length = 88
@@ -31,6 +35,8 @@ DICT_CONFIG = {
         "version": "1.0.0",
         "version_files": ["commitizen/__version__.py", "pyproject.toml"],
         "style": [["pointer", "reverse"], ["question", "underline"]],
+        "pre_bump_hooks": ["scripts/generate_documentation.sh"],
+        "post_bump_hooks": ["scripts/slack_notification.sh"],
     }
 }
 
@@ -49,6 +55,8 @@ _settings = {
     "update_changelog_on_bump": False,
     "use_shortcuts": False,
     "major_version_zero": False,
+    "pre_bump_hooks": ["scripts/generate_documentation.sh"],
+    "post_bump_hooks": ["scripts/slack_notification.sh"],
 }
 
 _new_settings = {
@@ -65,6 +73,8 @@ _new_settings = {
     "update_changelog_on_bump": False,
     "use_shortcuts": False,
     "major_version_zero": False,
+    "pre_bump_hooks": ["scripts/generate_documentation.sh"],
+    "post_bump_hooks": ["scripts/slack_notification.sh"],
 }
 
 _read_settings = {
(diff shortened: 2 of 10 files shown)


commitizen/hooks.py
```diff
@@ -0,0 +1,27 @@
...
+
+from commitizen import cmd, out
+from commitizen.exceptions import RunHookError
+
+
+def run(hooks, _env_prefix="CZ_", **env):
+    if isinstance(hooks, str):
+        hooks = [hooks]
+
+    for name, value in env.items():
```
Lee-W, line 11: The hooks part look good to me. But I'm a bit confused here. Why do we need this functionality?

tests/test_conf.py
```diff
@@ -73,6 +83,8 @@
     "version_files": ["commitizen/__version__.py", "pyproject.toml"],
     "style": [["pointer", "reverse"], ["question", "underline"]],
     "changelog_file": "CHANGELOG.md",
+    "hooks_pre_bump": ["scripts/generate_documentation.sh"],
+    "hooks_post_bump": ["scripts/slack_notification.sh"],
```
Lee-W, line 87: I think it would be better for us to add at least one run-able test

Lee-W (commented):
Hi @skoef thanks for your contribution! I left a few questions. Also I like the name `pre_bump_hooks` a bit better

commitizen/hooks.py
skoef: These are environment variables, available during executing of the scripts/commands in the hook. So for instance, the post bump hook passes these kind of environment variables to the script:
```bash
CZ_POST_CURRENT_TAG_VERSION=v3.0.0
CZ_POST_CHANGELOG_FILE_NAME=CHANGELOG.md
CZ_POST_WAS_INITIAL=True
CZ_POST_CURRENT_VERSION=3.0.0
CZ_POST_INCREMENT=MAJOR
CZ_POST_MESSAGE=bump: version 2.38.0 → 3.0.0
CZ_POST_PREVIOUS_VERSION=2.38.0
CZ_POST_PREVIOUS_TAG_VERSION=v2.38.0
```
Depending on what it is you are running during these hooks, these variables can be very useful.

tests/test_conf.py
skoef: @Lee-W I fully agree, but I'm not a real python dev myself and am not familiar with the pytest framework. If it wouldn't be too much work, could you perhaps pitch in here?

commitizen/hooks.py
Lee-W: I'm not sure I like this idea. (nor did i strongly oppose it) @woile what do you think?

tests/test_conf.py
Lee-W: yep, I'll take a deeper look later and see how I can help on this :)

commitizen/hooks.py
skoef: Is it the fact that the environment variables are set "globally" instead of as argument to `subprocess.Popen` or are you just not sure you like environment variables here at all? It is a relatively flexible way to pass metadata to the hook scripts and I think it is important to be able to pass that metadata in _some_ way to the hook scripts.

In the case of the `subprocess.Popen` argument, I could refactor it a bit so the env is not set directly to `os.environ` but rather passed to subprocess as argument.

Lee-W: I think what I'm not sure about is setting environment variables globally. It could potentially pollute or be polluted by other stuffs. But the `subprocess.Popen` solution sounds good!

skoef: I've implemented that and added a testcase for the pure function I introduced for it.

skoef:
@Lee-W I'm not sure why specifically the windows check failed, but do you know if it has something to do with passing env to subprocess?

ysndr (changes requested):

commitizen/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def run(hooks, _env_prefix="CZ_", **env):
+    if isinstance(hooks, str):
+        hooks = [hooks]
+
+    for hook in hooks:
+        out.info(f"Running hook '{hook}'")
+
+        c = cmd.run(hook, env=_format_env(_env_prefix, env))
```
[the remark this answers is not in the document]
skoef: Thanks for pointing this out, this has been resolved now

Lee-W:
> @Lee-W I'm not sure why specifically the windows check failed, but do you know if it has something to do with passing env to subprocess?

It seems to be something related to type annotation. I just fixed it and rebase the latest master to your branch.

ysndr:
@Lee-W
https://github.com/commitizen-tools/commitizen/compare/103def9d6c290afe1b0daca359581cefadce11ce..24f0f27ad7ecf1216d309ff8d7d1690528fbf4e7

this got lost in your rebase, breaking environment support again

skoef:
@ysndr should be fixed now

Lee-W:
@skoef  would it be possible for you to do rebase from the master branch (i.e. using `git rebase -i master`)? Otherwise, we won't be able to merge it. Thanks!

skoef:
@Lee-W done!

tests/test_conf.py
Lee-W: just added a test case to verify it

Lee-W (approved):
Thanks @skoef for your contribution! @woile I've added a few test cases to this PR and I plan to merge it these days. Let me know if you want to check. Thanks~

skoef:
@woile could you please take a look?

woile:
LGTM

Issue #640, opened 2022-12-16
customize changelog through config file


Corfucinas:
### Type

- [X] Content inaccurate
- [X] Content missing
- [ ] Typo

### URL

https://commitizen-tools.github.io/commitizen/customization/#custom-changelog-generator

### Description

This section explains how to customize the behavior of `cz` using `toml`, `.json` and `.yml` file formats. If a project opts for the modern `pyproject.toml` file, customizing becomes very easy by adding a `[tool.commitizen]` section. 

This works great and fast (no additional files, dependencies, or installs). However, if somebody wants to customize the changelog (see #513 ), the only given example us by creating a python file and subclassing it from `BaseCommitizen`. There's also the problem of initiating this class which requires four steps (or just using `cookiecutter gh:commitizen-tools/commitizen_cz_template`).

It would be a great improvement to show how to customize the CHANGELOG by just using the `toml` file.

Particularly I'm interested in referencing the `commit hash` inside it; an example is given on the website

```python
    def changelog_message_builder_hook(self, parsed_message: dict, commit: git.GitCommit) -> dict:
        rev = commit.rev
        m = parsed_message["message"]
        parsed_message["message"] = f"{m} {rev} [{commit.author}]({commit.author_email})"
        return parsed_message
```

What would be the equivalent of this inside a `pyproject.toml` file?

Lee-W:
I don't think we currently support this behavior. So I just added a feature label to this one. Let me know if you think "documentation" is a better label for this issue

Corfucinas:
> I don't think we currently support this behavior. So I just added a feature label to this one. Let me know if you think "documentation" is a better label for this issue

I think the "feature" is more appropriate, thanks!

robertschweizer:
I would love to see this exposed in the `pyproject.toml` config file!

>What would be the equivalent of this inside a pyproject.toml file?

I think the pyproject.toml needs to allow defining an alternative Jinja template (file path or multi-line string in pyproject.toml) instead of the default `keep_a_changelog_template.j2`.

Lee-W:
"file path"  sounds to be a more reasonable solution to me 🤔

PR #645, opened 2022-12-26
feat(template): allow to override the template from cli, configuration and plugins

### LINKED ISSUE #132: [feature - changelog] support custom change log template for changelog command

### LINKED ISSUE #384: Support outputting a changelog in reStructuredText
## Description
As the title states, it would be convenient for use with sphinx documentation to be able to output the changelog in `rst` format. 

## Possible Solution
This seems like it should be a configuration option versus a commandline option.

### LINKED ISSUE #433: Document how to add custom third party templates to commitizen when using pre-commit
## Type

* [ ] Content inaccurate
* [X] Content missing
* [ ] Typo

## URL
https://commitizen-tools.github.io/commitizen/third-party-commitizen/

## Description
As a non-expert in pre-commit, I'm not sure how to add these other dependencies to the pre-commit configuration for commitizen, nor how to add my own custom templates. Please expand on that documentation.

noirbizarre:
## Description

This PR provides full template customization by plugins, configuration and command line.

Plugins can now contribute a template (by providing their own jinja `PackageLoader` as `template_loader` attribute) as well as more template variables with their defaults (using the `template_extras` attribute from `BaseCommitizen). It was my initial use case, I was monkey-patching commitizen [like this](https://github.com/noirbizarre/emotional/blob/main/emotional/changelog.py)

Users can now provide their own template (in the format of their choice, like `.rst` in #384, with extras variables too) by:
- providing a `keep_a_changelog.j2` file at the root of their project
- setting the `template` config to a project root-relative path to their template
- using the `--template/-t` parameter in both `bump` and `changelog` command

They also can provide or override some template variables by:
- setting the `extras` config which is a dict
- using the `--extra/-e` parameter (can be used as many times as there are variables) in both `bump` and `changelog`

This change is backward compatible, fully tested and documented.

## Checklist

- [x] Make template overridable
- [x] Allow extra template parameters
- [x] Allow incremental with overridden templates
- [x] Provides examples for other syntaxes
  - [x] ReStructuredText
  - [x] AsciiDoc
  - [x] Textile

## Additional context

Fixes #132
Fixes #384                                                                                                                                     
Fixes #433
Closes #376 (I didn't keep the `tree` to `releases` variable renaming as it is a breaking change (I can however submit this as another PR because I think the `releases` naming is more explicit)


commitizen/cli.py
```diff
@@ -190,6 +190,20 @@
                         "default": None,
                         "help": "keep major version at zero, even for breaking changes",
                     },
+                    {
```
Lee-W, line 193: Would it be better if we could extract these command value as another dict?

commitizen/commands/changelog.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, config: BaseConfig, args):
         self.tag_format = args.get("tag_format") or self.config.settings.get(
             "tag_format"
         )
+        self.template = args.get("template") or self.config.settings.get("template")
+        self.extras = dict(
```
Lee-W, line 53: is there an example of this `=` usage? I'm a bit confused. I guess it's something like `cz ch -e key=value,key=value`?

Lee-W (commented):
Hi @noirbizarre for all your amazing contribution! I start to review this one but i've not yet be able to finish it. I'll do so when I have time. Thanks!

commitizen/cli.py
noirbizarre: Done 👍🏼

commitizen/commands/changelog.py
noirbizarre: I extracted the parsing in a regular `argparse.Action` and added some more examples.
I opted for `--extra key=value -e key-value` because the parsing is simpler.
It supported single and double quoted string properly, aka. this works: 
```shell
-e key-value -e single='a quoted value' and -e double="another quoted value"
``` 
and should result in:
```python
namespace["extras"] == {"key": "value", "single": "a quoted value", "double": "another quoted value"}
```

docs/customization.md
```diff
@@ -435,3 +435,84 @@ setup(
 ```
 
 Then your plugin will be available under the name `plugin`.
+
+## Customizing the changelog template
+
+Commitizen gives you the possibility to provide your own changelog template, by:
+
+- either by providing one with you customization class
```
woile, line 443: remove the `either by` in this list, it's too wordy

woile, line 443: typo, should be "your" not "you"

commitizen/changelog.py
```diff
@@ -31,13 +31,21 @@
...
+    FileSystemLoader,
+    PackageLoader,
+)
 
 from commitizen import defaults
 from commitizen.bump import normalize_tag
 from commitizen.exceptions import InvalidConfigurationError, NoCommitsFoundError
 from commitizen.git import GitCommit, GitTag
 
+DEFAULT_TEMPLATE = "keep_a_changelog_template.j2"
```
woile, line 47: Where is this file?

woile:
I like the PR 👍🏻 , though I see an issue, commitizen is using markdown to detect which was the last commit made, and from there create an incremental changelog, without touching the rest of the changelog. How would we guarantee this with the templates feature?

https://github.com/commitizen-tools/commitizen/blob/de5c39cf1f6a991fb0a35003bb384c6a3b55239d/commitizen/changelog.py#L150-L156

One solution would be to add parameter to the config:

```toml
[tool.commitizen]
version = "0.3.0"
latest_changelog_version = "0.1.0"
```

Now, it still worries me, how would commitizen know where to inject the incremental changelog?
At the moment we use an approximation system tied to markdown, if it is a **markdown title**, then we try to match to a version format.

If we allow for example templates in sphinx, or any other template system, how would commitizen know where to inject the template?

Even with my provided solution using `latest_changelog_version` a user could potentially add a version as part of a commit (which we have no control). For example, some adds:
```
fix: update dep X to 0.1.0
```

without a way to know if it's a title, commitizen would match that, and it would parse that version, potentially injecting the changelog there.

Does it make sense what I'm saying haha?

Could we have some tests with different templates? For example, using AsciiDoc

```AsciiDoc
= Changelog

== 0.1.0

=== Bug Fixes

-  update dep X to 0.1.0
```

[Textile](http://en.wikipedia.org/wiki/Textile_(markup_language))

```Textile
h1. Changelog

h2. 0.1.0

h3. Bug Fixes

*  update dep X to 0.1.0
```

Thanks!

woile:
We could potentially add a command to `cz changelog --export-template` that prints the template to stdout 🤔

noirbizarre:
Oh, I didn't notice the template parsing. That is definitively something I need to handle in this PR.
I'll fix all comments and try adding something to handle changelog reverse parsing.

The problem I see with the `latest_changelog_version` approach is that it is not backward compatible, and we lose some automation.

What about making `parse_version_from_markdown` a `parse_version_from_changelog` method on `BaseCommitizen` ?

This should be backward compatible, Plugin providing their template could override to provide the parsing. For the template from CLI or conf, I can allow providing a string pointing to a custom method, something like:

```toml
[tool.commitizen]
template = "my_template.whatever"
changelog_version_parser = "path.to:my_parser"
```

I like the `cz changelog --export-template` idea ! Anyone can bootstrap its customized template from the original one. I'll this too 👍🏼

commitizen/commands/changelog.py
Lee-W: got it. it makes sense!

Lee-W:
Hi @noirbizarre I notice you have some discussion with @woile . Wondering whether this pr is ready for review now?

noirbizarre:
I rebased it on latest `v3`, added the `changelog --export-template` but it's lacking the incremental part (from the discussion 👆🏼).
I hope to be able to deliver it by the end of the week.
However, it is already possible to review what's in it. I'll add the incremental part as an extra commit to help review it later.

Lee-W:
Got it. Let me focus on #646 first and will get back to this one afterward 💪

noirbizarre:
Rebased on latest `v3` changes.
I updated the description's checklist to reflect latest review discussions and track what is still remaining to do.

Lee-W:
@noirbizarre Wondering are we going to finish them in this PR? Or will they be sent in another PR ? 

I'm thinking of releasing `v3` after these few merged. @woile Let me know what you think :)

woile:
I like your proposal of adding a method to find the version: `parse_version_from_changelog`.



We could add something to make the line unique, making it easier to find the version. For example, if the template gave the user the `tag + commit-hash`, the person writing the template could include the commit hidden (in markdown):

as a pseudo comment

```md
## {{ version }} ({{ data }}) [comment]: <> {{ hash_id }}
``` 

or html comment

```md
## {{ version }} ({{ date }}) <!--- {{ hash_id }} -->
``` 

or not even hidden and instead, as a link.

Then it becomes much simpler to find the line.

---

On another note, I'm not sure if exposing `--template` and `changelog_version_parser` to the user is wise, ideally they should take the time to implement their own commitizen parser. I say this because, I feel like then, we are gonna have to give support to every user complaining it doesn't work when they make a wrong template or wrong version_parser. What do you think @Lee-W @noirbizarre ?

Lee-W:
We could point them to somewhere in the document or add such doc when encountering such stuffs. I'm still positive to this feature :)

Lee-W:
@noirbizarre hi wondering, have we finished all the things we want to do in this PR? if so, i'll try to find a time for taking a deeper look

noirbizarre:
Sorry, I did not had time to work on this since (except for some rebases), but I will finish in the coming days.

Lee-W:
@noirbizarre not at all 🙂 take your time. I'm also a bit overwhelming these days 😫

noirbizarre:
Hi 👋🏼 

I have some good and bad news.

The good: I have multiple working implementations with both incremental support and support for RestructuredText, Textile and AsciiDoc.

The bad: this raised a lots of issues/questions, including those:
- the `file_name`/`changelog_file` is assumed `CHANGELOG.md` at a lot of places, sometimes hardcoded
- what format do you want for additionnal file format support ?

So I have 2 proposals:
- either introduce an explicit format notion, which default to Markdown to be backward compatible. 
- either make them some `CzConventionnalCommit` child classes

In both cases, I tested and works well, but raise the number of tests and making it builtin also increase the support surface.
In both case, I think `changelog` helpers should become a class because access to config and the current `cz` extension is required.

I have a preference for the first one as it is simpler to test, more flexible and makes `commitizen` more format-agnostic.


About the `v3` strategy: given the amount of stacked changes, given there are still some incoming PRs, I think making v3 the main would be easier. Everything incoming should go to `v3` and this would avoid people working on the v2 stack and then discover it conflict with `v3`.
In particular for #686 (on which I could help given I already started the same).
This would allows for easier refactoring (like the `Changelog` class` but I have a few small others). I you fear to be stuck in this state, we can agree on a deadline. Given I already have multiple working implementations of this PR, time is not the issue anymore. But that would allows be to submit some changes outside this PR while keeping focused on template override in this one.

So, I wait for your decisions. Maybe we need a more "realtime" talk on a chat, a visio or whatever, just to tackle some topics and make quick decisions to go forward.

woile:
Hi noir! `v3` is now the default. My idea is to release v3 this or next week.

Regarding your proposal, could you give an example of what you mean by "explicit format notion"?

You can find me here for a chat: https://hachyderm.io/@woile

docs/customization.md
```diff
@@ -435,3 +435,84 @@ setup(
...
+- providing one from the current working directory and setting it:
+    - as [configuration][template-config]
+    - as `--template` parameter to both `bump` and `changelog` commands
+- either by providing a template with the same name as the default template
+
+By default, the template used is the `CHANGELOG.md.j2` file from the commitizen repository.
+
+### Providing a template with your customization class
+
+There is 3 parameters available to hcange the template rendering from your custom `BaseCommitizen`.
```
Skeen, line 453:
```suggestion
There is 3 parameters available to change the template rendering from your custom `BaseCommitizen`.
```

noirbizarre: Thanks 🙏🏼 
(updated during rebase)

docs/changelog.md
```diff
@@ -21,9 +22,19 @@ optional arguments:
                         set the value for the new version (use the tag value), instead of using unreleased
   --incremental         generates changelog from last created version, useful if the changelog has been manually modified
   --start-rev START_REV
+<<<<<<< HEAD
```
Lee-W, line 25: We seem to accidentally add git conflict markers

docs/customization.md
```diff
@@ -435,3 +435,84 @@ setup(
...
+### Providing a template from the current working directory
+
+Users can provides their own template from their current working directory (your project root) by:
+
+- providing a template with the same name (`CHANGELOG.md.j2` unless overriden by your custom class)
+- setting your template path as `template` configuration
+- giving your template path as `--template` parameter to `bump` and `changelog` commands
+
+!!! Note
+    The path is relative to the current working directory, aka. your project root most of the time,
```
Lee-W, line 489:
```suggestion
    The path is relative to the current working directory, aka. your project root most of the time.
```

docs/changelog.md
Lee-W, line 25:
```suggestion
```

docs/changelog.md
```diff
@@ -21,9 +22,19 @@ optional arguments:
...
                         start rev of the changelog. If not set, it will generate changelog from the start
   --merge-prerelease
                         collect all changes from prereleases into next non-prerelease. If not set, it will include prereleases in the changelog
+||||||| parent of a3c1026 (feat(template): allow to override the template from cli, configuration and plugins)
```
Lee-W, line 29:
```suggestion
```

docs/changelog.md
```diff
@@ -21,9 +22,19 @@ optional arguments:
...
+                        start rev of the changelog.If not set, it will generate changelog from the start
+=======
```
Lee-W, line 31:
```suggestion
```

docs/changelog.md
```diff
@@ -21,9 +22,19 @@ optional arguments:
...
+                        start rev of the changelog.If not set, it will generate changelog from the start
+  --template TEMPLATE, -t TEMPLATE
+                        changelog template file name (relative to the current working directory)
+  --extra EXTRA, -e EXTRA
+                        a changelog extra variable (in the form 'key=value')
+>>>>>>> a3c1026 (feat(template): allow to override the template from cli, configuration and plugins)
```
Lee-W, line 37:
```suggestion
```

docs/customization.md
```diff
@@ -435,3 +435,84 @@ setup(
...
+
+### Template variables
+
+The default template use a single `tree` variable which is a list of entries (a release) with the following format:
+
+| Name | Type | Description |
+| ---- | ---- | ----------- |
+| version | `str` | The release version |
+| date | `datetime` | The release date |
+| changes | `list[tuple[str, list[Change]]]` | The release sorted changes list in the form `(type,changes)` |
```
Lee-W, line 499:
```suggestion
| changes | `list[tuple[str, list[Change]]]` | The release sorted changes list in the form `(type, changes)` |
```

commitizen/changelog.py
```diff
@@ -47,6 +54,8 @@
     # workaround mypy issue for 3.7 python
     VersionProtocol = typing.Any
 
+DEFAULT_TEMPLATE = "CHANGELOG.md.j2"
```
Lee-W, line 57: Should we move it to `defaults.py`?

commitizen/cli.py
```diff
@@ -17,6 +18,51 @@
...
+        will result in
+
+        namespace["opt"] == {
+            "key": "value",
+            "double-quotes": "value",
+            "single-quotes": "value",
+        }
+    """
+
+    def __call__(self, parser, namespace, kwarg, option_string=None):
```
Lee-W, line 42: nitpick: would be great if we could add type annotation here

commitizen/commands/changelog.py
```diff
@@ -106,6 +114,11 @@ def write_changelog(
                 changelog_out = changelog_hook(changelog_out, partial_changelog)
             changelog_file.write(changelog_out)
 
+    def export_template(self):
```
Lee-W, line 117: Just curious. Is this function mainly for users to debug their template? or when do we expect the users to use it?

tests/commands/test_bump_command.py
```diff
@@ -980,3 +984,129 @@ def test_bump_command_prelease_version_type_check_old_tags(
     for version_file in [tmp_version_file, tmp_commitizen_cfg_file]:
         with open(version_file, "r") as f:
             assert "0.2.0" in f.read()
+
+
+@pytest.mark.parametrize(
+    "arg,cfg,expected",
```
Lee-W, line 990:
```suggestion
    "arg, cfg, expected",
```

tests/commands/test_changelog_command.py
```diff
@@ -1210,3 +1216,160 @@ def test_changelog_prerelease_rev_with_use_version_type_semver(
     out, _ = capsys.readouterr()
 
     file_regression.check(out, extension=".second-prerelease.md")
+
+
+@pytest.mark.parametrize(
```
Lee-W, line 1221: The tests here look similar to the ones in `tests/commands/test_bump_command.py`. Should we reduplicate them? This doesn't have to be done in this PR

tests/test_changelog.py
```diff
@@ -1136,6 +1139,91 @@ def test_render_changelog(gitcommits, tags, changelog_content):
     assert result == changelog_content
 
 
+def test_render_changelog_from_default_plugin_values(
+    gitcommits, tags, changelog_content
+):
+    parser = ConventionalCommitsCz.commit_parser
+    changelog_pattern = ConventionalCommitsCz.bump_pattern
```
Lee-W, line 1146:
```suggestion
    changelog_pattern = ConventionalCommitsCz.changelog_pattern
```

Even though they're identical now, I think it might be better for us to use `changelog_pattern` here

tests/test_changelog.py
```diff
@@ -1136,6 +1139,91 @@ def test_render_changelog(gitcommits, tags, changelog_content):
...
+    result = changelog.render_changelog(tree, loader=loader, template=template)
+    assert result == changelog_content
+
+
+def test_render_changelog_override_loader(gitcommits, tags, tmp_path: Path):
+    loader = FileSystemLoader(tmp_path)
+    tpl = "loader overridden"
+    (tmp_path / changelog.DEFAULT_TEMPLATE).write_text(tpl)
+    parser = ConventionalCommitsCz.commit_parser
+    changelog_pattern = ConventionalCommitsCz.bump_pattern
```
Lee-W, line 1161:
```suggestion
    changelog_pattern = ConventionalCommitsCz.changelog_pattern
```

tests/test_changelog.py
```diff
@@ -1136,6 +1139,91 @@ def test_render_changelog(gitcommits, tags, changelog_content):
...
+    )
+    result = changelog.render_changelog(tree, loader=loader)
+    assert result == tpl
+
+
+def test_render_changelog_override_template_from_cwd(gitcommits, tags, chdir: Path):
+    tpl = "overridden from cwd"
+    (chdir / changelog.DEFAULT_TEMPLATE).write_text(tpl)
+    parser = ConventionalCommitsCz.commit_parser
+    changelog_pattern = ConventionalCommitsCz.bump_pattern
```
Lee-W, line 1173:
```suggestion
    changelog_pattern = ConventionalCommitsCz.changelog_pattern
```

tests/test_changelog.py
```diff
@@ -1136,6 +1139,91 @@ def test_render_changelog(gitcommits, tags, changelog_content):
...
+
+
+def test_render_changelog_override_template_from_cwd_with_custom_name(
+    gitcommits, tags, chdir: Path
+):
+    tpl = "template overridden from cwd"
+    tpl_name = "tpl.j2"
+    (chdir / tpl_name).write_text(tpl)
+    parser = ConventionalCommitsCz.commit_parser
+    changelog_pattern = ConventionalCommitsCz.bump_pattern
```
Lee-W, line 1188:
```suggestion
    changelog_pattern = ConventionalCommitsCz.changelog_pattern
```

tests/test_changelog.py
```diff
@@ -1136,6 +1139,91 @@ def test_render_changelog(gitcommits, tags, changelog_content):
...
+
+def test_render_changelog_override_loader_and_template(
+    gitcommits, tags, tmp_path: Path
+):
+    loader = FileSystemLoader(tmp_path)
+    tpl = "loader and template overridden"
+    tpl_name = "tpl.j2"
+    (tmp_path / tpl_name).write_text(tpl)
+    parser = ConventionalCommitsCz.commit_parser
+    changelog_pattern = ConventionalCommitsCz.bump_pattern
```
Lee-W, line 1204:
```suggestion
    changelog_pattern = ConventionalCommitsCz.changelog_pattern
```

tests/test_changelog.py
```diff
@@ -1136,6 +1139,91 @@ def test_render_changelog(gitcommits, tags, changelog_content):
...
+    result = changelog.render_changelog(tree, loader=loader, template=tpl_name)
+    assert result == tpl
+
+
+def test_render_changelog_support_arbitrary_kwargs(gitcommits, tags, tmp_path: Path):
+    loader = FileSystemLoader(tmp_path)
+    tpl_name = "tpl.j2"
+    (tmp_path / tpl_name).write_text("{{ key }}")
+    parser = ConventionalCommitsCz.commit_parser
+    changelog_pattern = ConventionalCommitsCz.bump_pattern
```
Lee-W, line 1217:
```suggestion
    changelog_pattern = ConventionalCommitsCz.changelog_pattern
```

Lee-W (commented):
This PR is awesome 🤩 Sorry for taking so long. I finally have some bandwidth to take a deeper look. Just left some comments. Most of them are just minor changes and suggestions.

Lee-W:
@woile I think we've merged the changes in this PR, right? If so, I'm going to close this one

woile:
I rebased and released v3. The simplest way to fix this PR is to create a new branch and cherrypick your last 4 commits. Sorry for the inconvinience.

Now that v3 has the new plugin system, wouldn't be a good option to expose different changelog as plugins?

Commitizen could have this plugin by default:

```toml
[tool.poetry.plugins."commitizen.changelog_provider"]
commitizen_md = "commitizen.changelog_providers:CommitizenMarkdown"
```

We would have to create a protocol probably, exposing the desired settings that need to be exposed.

noirbizarre:
I updated the pull-request (force pushed on my branch, given I already had pending changes).

I chose to have 3 extra plugin for the format demonstration. I didn't include any further changelog refactorings that I would probably submit separately, given the scope and the size of this PR.

@Lee-W I took you all your changes and suggestions, they are already included in the rebased version

@woile I have withdrawn the format notion, it is in fact not relevant because:

- places where it was required were totally related to the template, not the format
- some format (`restructuredtext` in particular) does not have a one-line based title format and requires extra dependencies (`doctuils` for `restructuredtext`) to handle the general case

Note: build failure is not related to the code but to PyPI being out of service (HTTP 502), relaunching the workflow should make it pass

commitizen/commands/changelog.py
noirbizarre: This addition was following https://github.com/commitizen-tools/commitizen/pull/645#issuecomment-1374407611. The idea is to provide users and plugin developpers with:
- a way to inspect the template
- a way to quickly customize it locally (export to current dir, do the change you want)

woile:
Thanks a lot for the effort, it will take me some time to review, as I'm quite busy, so I apologize in advance

pyproject.toml
```diff
@@ -79,6 +79,9 @@ git-cz = "commitizen.cli:main"
 cz_conventional_commits = "commitizen.cz.conventional_commits:ConventionalCommitsCz"
 cz_jira = "commitizen.cz.jira:JiraSmartCz"
 cz_customize = "commitizen.cz.customize:CustomizeCommitsCz"
+conventional_commits_asciidoc = "commitizen.cz.conventional_commits:ConventionalCommitsAsciiDoc"
+conventional_commits_textile = "commitizen.cz.conventional_commits:ConventionalCommitsTextile"
+conventional_commits_restructuredtext = "commitizen.cz.conventional_commits:ConventionalCommitsRst"
```
woile, line 84: This I was thinking that it could be it's own category, separated from the "rules" provider. The way we have in the new line `86` the "customer version providers". We could have `tool.poetry.plugins."commitizen.changelog"`.

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -213,3 +215,104 @@ def process_commit(self, commit: str) -> str:
         if m is None:
             return ""
         return m.group(3).strip()
+
+
+class ConventionalCommitsTextile(ConventionalCommitsCz):
```
woile, line 220: Would it be possible to make this 'changelog providers' not tied to conventional commits? How would it work for any other plugin?

noirbizarre:
I answer the 2 comments here: it's not possible in the current state.
As of now it only works because the template and its plugin makes a lot of assumptions which are only true for each specific plugin/template couples:
- the version is in the title
- the title format is a one-liner (restructuredtext-style titles are legit in Markdown)
- the title contains the entire version
- the version is a PEP440 compliant version (should be fixed by #686)
- ...

I was thinking it possible until I try to handle restructuredtext: sections titles format are up to the writer, so I made a plugin which is able to detect version/title only for its template. It was a good exercise to take into accounts all customization possibilities but made me realize that handling the specific template associated with a plugin is totally different from handling properly independent file formats. 

As soon as you think it is tied to the format, complexity raises exponentially. To understand, just try to imagine the following:
- what if I use the [alternate title format](https://www.markdownguide.org/basic-syntax#alternate-syntax) (seen in some changelog) ? It would still be valid markdown, but this won't work anymore
- what if just want a nested bulleted list of versions ? Still valid markdown, but the algorithm doesn't work anymore
- what if I just add an HTML anchor before the title because I need this for my platform ? Still valid markdown, still not working anymore

The `parse_version_from_markdown` is entirely tied to the template and, in fact, not that much to Markdown. Given the PR goal is to allow to entirely customizing template, I had no choice but to make the version/title detection part of the plugin.

Splitting the template rendering from the commit convention, while being a good long term goal, is far from easy. Rendering and reparsing for incremental rendering are as of now made taking the template into consideration and not the file format. Commitizen does not manipulate markdown, only commits and templates.

TBH, I tried, and it resulted into one of these 2 possibilities:
- format is not enough and there is for each format a list of constraint to define, which are most of the time defeating the purpose of allowing customization (you can customize the template, but you should keep this, that and that or else it won't be working anymore)
- format is enough, but each format implementation will rely on a dedicated lib to be able to handle all cases supported by the format but not taken into account by the current template/plugin couples). There will still be some constraints because the current algorithm still assume a lot on the template format

Both cases feel like putting too much constraint on the customization (and as a consequence will raise a lot of support to answer questions like the above "what if").

The chosen approach has been taken because:
- it does not rely on format parsing
- it works whatever the end-user want to make as customization
- it is backward-compatible

The same apply for rules and templates: whatever rules you use to group commits, your template will have to take them into consideration. Ex: grouping by scope in the template is only possible if there is a scope in the convention and so in the parsed commits.

In my humble opinion: file format is entirely tied to the template and should stay this way. What would be great in some future would be to split plugins into commit convention and template rendering, but this was not the goal of this PR and this is not a small task. I think it requires a bit of conception and at least a bit of feedback on what is currently being delivered (because if this is the theoretical ideal, I am not sure the user needs this kind of complexity and almost sure you don't want to maintain this kind of complexity).
I would provide the template customization and wait for feedback on what is actually missing or too complex to handle or not fit.

Remember that a few months ago, there was no customization possible and this PR was initially just providing a simple way to slightly override templates (my case was working since the first commit, everything added after was to support cases which are not mine, I had to dig textile, asciidoc, restructuredtext... while I only wanted to override a markdown template).
I would love to be able to merge this one (it is more than 4 months old, rebases are costing more and more given the current size of the PR) and then iterate to improve based on real feedback on limitations.

What do you think ?

robertschweizer:
I believe this would also close https://github.com/commitizen-tools/commitizen/issues/640

noirbizarre:
Forget about my above comment: I should be able to split a `commitizen.format` from the current `commitizen.plugin` rules. In fact it was my first attempt and I dropped it because I was focusing on having the smallest pull-request possible. Given how the scope widen, I think it is now an acceptable solution. Even more if #773 is merged because it will allow to have something quite nice (full separation between the version source, the version scheme, the changelog format and the rules).

@robertschweizer I will add this to the commit 👍🏼

woile:
Impressive work 💪🏻 this will take me some time to properly test and review.

At first hindsight things look great. 
How is commitizen loading the changelog generator? is it based on the extension? If my setting says `CHANGELOG.rst` will it then load the `restructured text` generator?

I think we should also add some documentation over the supported formats in the `changelog` section.

woile:
Should we think about possible multiple plugins for the same format? Like 2 plugins for markdown? What do you think?

woile:
I have not forgotten this PR, apologies for the delay, I've been very busy with other matters. Do you think you could rebase? After that I think it's ready to merge

Skeen:
I would very much like to see this merged. Can I assist somehow?

Skeen:
@woile I did a rebase of this PR here: https://github.com/commitizen-tools/commitizen/pull/808

noirbizarre:
Sorry, I wasn't available lately. 
I started a rebase I didn't submitted yet.
@Skeen Thanks for the rebase ! Is it working ? If so maybe for tracebility (history, comments...) I should take it and push it back there ? Would you be OK with this ?

@Lee-W @woile What would you prefer as path of action ?

Skeen:
> Sorry, I wasn't available lately. I started a rebase I didn't submitted yet. @Skeen Thanks for the rebase ! Is it working ? If so maybe for tracebility (history, comments...) I should take it and push it back there ? Would you be OK with this ?
> 
> @Lee-W @woile What would you prefer as path of action ?

Hi @noirbizarre,

I believe the rebase is working, but I haven't done a throughout manual testing.

Feel free to take it if you wish, that's super okay with me.
My motivation for doing the rebase was purely to get it merged by any means possible.

woile:
No preference on my side. If you update this one I'll review it

noirbizarre:
OK, PR has been rebased against the current `master` (Thanks @Skeen for the help) and is now up-to-date and ready to review.

pyproject.toml
```diff
@@ -83,6 +83,12 @@ cz_conventional_commits = "commitizen.cz.conventional_commits:ConventionalCommit
 cz_jira = "commitizen.cz.jira:JiraSmartCz"
 cz_customize = "commitizen.cz.customize:CustomizeCommitsCz"
 
+[tool.poetry.plugins."commitizen.format"]
```
Lee-W, line 86: Should it be changelog_format or something similar to make it more readable?

docs/config.md
```diff
@@ -192,6 +192,22 @@ Default: `utf-8`
...
+
+Provide custom changelog jinja template path relative to the current working directory. [Read more][template-customization]
+
+### `extras`
+
+Type: `dict[str, Any]`
+
+Default: `{}`
+
+Provide extra variables to the changelog template. [Read more][template-customization] |
```
Lee-W, line 209: Is the `|` at the end expected? or is it a typo?

Lee-W (commented):
As this is a huge PR, I'm planning to review it in a few rounds

pyproject.toml
noirbizarre: It's a free form string so yes, it can `changelog_provider` but I chose `format` only for the follozing reason:
- I try as much as possible to stick with single word namings
- It was consistent with other endpoints: `commitizen.provider`, `commitizen.scheme` and `commitizen.plugin` which should be more like `commitizen.version_provider`, `commitizen.version_scheme`, `commitizen.core_plugin` based on the same reasonning
- It was consistent with the package name too `commitizen.format`
- it has been a very long iteration and the name comes from a past discussion in this PR (can't remind which one)

But except those points, I don't have a strong opinion on this, so I can rename it if you prefer

docs/config.md
noirbizarre: No, I think this is a rebase artifact (it was a table by the time I started this PR). There was another one this file which is on `main` too. I added a fix commit for those

pyproject.toml
Lee-W: If there were already discussions on it, I'm good with it. @woile what do you think?

woile: I like `changelog_provider`, and I think the others should be changed 🤔 of course no need now. Just `format` can be too open to interpretation

noirbizarre: Renamed from `commitizen.format` to `commitizen.changelog_format` in the last commit 👍

woile (approved):

commitizen/formats/errors.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+
+class FormatError(Exception):
```
Lee-W, line 4: It would be better if we could inherit from https://github.com/commitizen-tools/commitizen/blob/master/commitizen/exceptions.py#L38 and assign an error code to it

commitizen/defaults.py
```diff
@@ -57,6 +57,9 @@ class Settings(TypedDict, total=False):
     prerelease_offset: int
     encoding: str
     always_signoff: bool
+    format: str | None
```
Lee-W, line 60: I think it might be better for us to change it to `changelog_format` as well

commitizen/defaults.py
```diff
@@ -98,6 +101,9 @@ class Settings(TypedDict, total=False):
     "prerelease_offset": 0,
     "encoding": encoding,
     "always_signoff": False,
+    "format": None,  # default guessed from changelog_file
```
Lee-W, line 104: I think it might be better for us to change it to `changelog_format` as well

commitizen/defaults.py
```diff
@@ -129,3 +135,4 @@ class Settings(TypedDict, total=False):
 bump_message = "bump: version $current_version → $new_version"
 
 commit_parser = r"^((?P<change_type>feat|fix|refactor|perf|BREAKING CHANGE)(?:\((?P<scope>[^()\r\n]*)\)|\()?(?P<breaking>!)?|\w+!):\s(?P<message>.*)?"  # noqa
+version_parser = r"(?P<version>([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+)?(\w+)?)"
```
Lee-W, line 138: Not quite sure. but should we use version_scheme or  reuse https://github.com/commitizen-tools/commitizen/blob/master/commitizen/version_schemes.py#L36C1-L36C1?

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
+from __future__ import annotations
+
+import sys
+
+from typing import ClassVar, Optional
```
Lee-W, line 5:
```suggestion
from typing import ClassVar
```

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+        Extract the changelog metadata.
+        """
+
+
+KNOWN_FORMATS: dict[str, type[Format]] = {
+    ep.name: ep.load() for ep in metadata.entry_points(group=FORMAT_ENTRYPOINT)
+}
+
+
+def get_format(config: BaseConfig, filename: Optional[str] = None) -> Format:
```
Lee-W, line 61:
```suggestion
def get_format(config: BaseConfig, filename: str | None = None) -> Format:
```

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Get a format from its name
+
+    :raises FormatUnknown: if a non-empty name is provided but cannot be found in the known formats
+    """
+    name: Optional[str] = config.settings.get("format")
```
Lee-W, line 67:
```suggestion
    name: str | None = config.settings.get("format")
```

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    format: Optional[type[Format]] = guess_format(filename)
```
Lee-W, line 68:
```suggestion
    format: type[Format] | None = guess_format(filename)
```

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    if name and name in KNOWN_FORMATS:
+        format = KNOWN_FORMATS[name]
+
+    if not format:
+        raise FormatUnknown(f"unknown format '{name}'")
+
+    return format(config)
+
+
+def guess_format(filename: Optional[str]) -> Optional[type[Format]]:
```
Lee-W, line 79:
```suggestion
def guess_format(filename:str | None) ->type[Format] | None:
```

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+FORMAT_ENTRYPOINT = "commitizen.changelog_format"
+DEFAULT_FORMAT = "markdown"
+DEFAULT_TEMPLATE_EXTENSION = "j2"
+
+
+class Format(Protocol):
```
Lee-W, line 24: I guess we'd better change it to `ChangelogFormat` as well ?

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+FORMAT_ENTRYPOINT = "commitizen.changelog_format"
+DEFAULT_FORMAT = "markdown"
+DEFAULT_TEMPLATE_EXTENSION = "j2"
+
```
Lee-W, line 22: Not sure whether we should put it here or in https://github.com/commitizen-tools/commitizen/blob/master/commitizen/defaults.py

commitizen/formats/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def template(self) -> str:
+        """Expected template name for this format"""
+        return f"CHANGELOG.{self.extension}.{DEFAULT_TEMPLATE_EXTENSION}"
+
+    @property
+    def default_changelog_file(self) -> str:
+        return f"CHANGELOG.{self.extension}"
+
+    def get_metadata(self, filepath: str) -> Metadata:
```
Lee-W, line 50: Do we need to raise `Not Implemented Error`?

commitizen/formats/asciidoc.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+import re
+
+from typing import Optional
```
Lee-W, line 5:
```suggestion
```

commitizen/formats/asciidoc.py
```diff
@@ -0,0 +1,31 @@
...
+
+from .base import BaseFormat
+
+
+class AsciiDoc(BaseFormat):
+    extension = "adoc"
+
+    RE_TITLE = re.compile(r"^(?P<level>=+) (?P<title>.*)$")
+
+    def parse_version_from_title(self, line: str) -> Optional[str]:
```
Lee-W, line 17:
```suggestion
    def parse_version_from_title(self, line: str) -> str | None:
```

commitizen/formats/base.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import annotations
+from abc import ABCMeta
+
+import os
+
+from typing import IO, Any, ClassVar, Optional
```
Lee-W, line 6:
```suggestion
from typing import IO, Any, ClassVar
```

commitizen/formats/base.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import annotations
+from abc import ABCMeta
+
+import os
```
Lee-W, line 4:
```suggestion


import os
from abc import ABCMeta
```

commitizen/formats/base.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_metadata(self, filepath: str) -> Metadata:
+        if not os.path.isfile(filepath):
+            return Metadata()
+
+        with open(filepath, "r") as changelog_file:
+            return self.get_metadata_from_file(changelog_file)
+
+    def get_metadata_from_file(self, file: IO[Any]) -> Metadata:
+        meta = Metadata()
+        unreleased_level: Optional[int] = None
```
Lee-W, line 36:
```suggestion
        unreleased_level: int | None = None
```

commitizen/formats/base.py
```diff
@@ -0,0 +1,76 @@
...
+        for index, line in enumerate(file):
+            line = line.strip().lower()
+
+            unreleased: Optional[int] = None
```
Lee-W, line 40:
```suggestion
            unreleased: int | None = None
```

commitizen/formats/textile.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+import re
+
+from typing import Optional
```
Lee-W, line 5:
```suggestion
```

commitizen/formats/textile.py
```diff
@@ -0,0 +1,29 @@
...
+
+from .base import BaseFormat
+
+
+class Textile(BaseFormat):
+    extension = "textile"
+
+    RE_TITLE = re.compile(r"^h(?P<level>\d)\. (?P<title>.*)$")
+
+    def parse_version_from_title(self, line: str) -> Optional[str]:
```
Lee-W, line 17:
```suggestion
    def parse_version_from_title(self, line: str) -> str | None:
```

commitizen/formats/textile.py
```diff
@@ -0,0 +1,29 @@
...
+        if not self.RE_TITLE.match(line):
+            return None
+        m = re.search(defaults.version_parser, line)
+        if not m:
+            return None
+        return m.group("version")
+
+    def parse_title_level(self, line: str) -> Optional[int]:
```
Lee-W, line 25:
```suggestion
    def parse_title_level(self, line: str) -> int | None:
```

commitizen/formats/markdown.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+import re
```
Lee-W, line 2:
```suggestion

import re
```

commitizen/formats/markdown.py
```diff
@@ -0,0 +1,29 @@
...
+from typing import Optional
```
Lee-W, line 3:
```suggestion

```

commitizen/formats/markdown.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Markdown(BaseFormat):
+    extension = "md"
+
+    alternative_extensions = {"markdown", "mkd"}
+
+    RE_TITLE = re.compile(r"^(?P<level>#+) (?P<title>.*)$")
+
+    def parse_version_from_title(self, line: str) -> Optional[str]:
```
Lee-W, line 16:
```suggestion
    def parse_version_from_title(self, line: str) -> str | None:
```

commitizen/formats/markdown.py
```diff
@@ -0,0 +1,29 @@
...
+        m = self.RE_TITLE.match(line)
+        if not m:
+            return None
+        m = re.search(defaults.version_parser, m.group("title"))
+        if not m:
+            return None
+        return m.group("version")
+
+    def parse_title_level(self, line: str) -> Optional[int]:
```
Lee-W, line 25:
```suggestion
    def parse_title_level(self, line: str) -> int | None:
```

commitizen/formats/restructuredtext.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import annotations
+
+import re
+import sys
+from itertools import zip_longest
+from typing import IO, TYPE_CHECKING, Any, Optional, Union, Tuple
```
Lee-W, line 6:
```suggestion
from typing import IO, TYPE_CHECKING, Any
```

commitizen/formats/restructuredtext.py
```diff
@@ -0,0 +1,95 @@
...
+        unreleased_title_kind: Optional[TitleKind] = None
+        in_overlined_title = False
+        lines = file.readlines()
+        for index, (first, second, third) in enumerate(
+            zip_longest(lines, lines[1:], lines[2:], fillvalue="")
+        ):
+            first = first.strip().lower()
+            second = second.strip().lower()
+            third = third.strip().lower()
+            title: Optional[str] = None
```
Lee-W, line 47:
```suggestion
            title: str | None = None
```

commitizen/formats/restructuredtext.py
```diff
@@ -0,0 +1,95 @@
...
+            kind: Optional[TitleKind] = None
```
Lee-W, line 48:
```suggestion
            kind: TitleKind | None = None
```

commitizen/commands/changelog.py
```diff
@@ -115,18 +127,26 @@ def write_changelog(
...
```
Lee-W, line 130: nitpick

```suggestion
    def export_template(self) -> None:
```

tests/test_changelog_formats.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import annotations
+from typing import Optional
```
Lee-W, line 2:
```suggestion
```

tests/test_changelog_formats.py
```diff
@@ -0,0 +1,55 @@
...
+    "name, expected",
+    [pytest.param(name, format, id=name) for name, format in KNOWN_FORMATS.items()],
+)
+def test_get_format(config: BaseConfig, name: str, expected: type[Format]):
+    config.settings["format"] = name
+    assert isinstance(get_format(config), expected)
+
+
+@pytest.mark.parametrize("filename", (None, ""))
+def test_get_format_empty_filename(config: BaseConfig, filename: Optional[str]):
```
Lee-W, line 38:
```suggestion
def test_get_format_empty_filename(config: BaseConfig, filename: str | None):
```

tests/test_changelog_formats.py
```diff
@@ -0,0 +1,55 @@
...
+def test_get_format_empty_filename_no_setting(
+    config: BaseConfig, filename: Optional[str]
+):
+    config.settings["format"] = None
+    with pytest.raises(FormatUnknown):
+        get_format(config, filename)
+
+
+@pytest.mark.parametrize("filename", ("extensionless", "file.unknown"))
+def test_get_format_unknown(config: BaseConfig, filename: Optional[str]):
```
Lee-W, line 53:
```suggestion
def test_get_format_unknown(config: BaseConfig, filename: str|None):
```

Lee-W (commented):
Finally, finish reviewing it 🙌 I think we're really close to merge

commitizen/formats/restructuredtext.py
```diff
@@ -0,0 +1,95 @@
...
+
+if TYPE_CHECKING:
+    # TypeAlias is Python 3.10+ but backported in typing-extensions
+    if sys.version_info >= (3, 10):
+        from typing import TypeAlias
+    else:
+        from typing_extensions import TypeAlias
+
+
+TitleKind: TypeAlias = Union[str, Tuple[str, str]]
```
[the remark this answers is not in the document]
noirbizarre: This one is broken before Python 3.10, cf. https://bugs.python.org/issue42233

noirbizarre:
I update the PR including:
- rebased on latest `master` (on #889 exactly to benefit from Python 3.8+ syntax update)
- applied all suggestion except the `-> None` because other methods don't have this syntax (I think this should go in another PR where linter is configured accordingly)
- all `format/Format` have been replace by `changelog_format/ChangelogFormat` everywhere
- history squashed to have consistent commits (and so changelog)
- `version_parser` handling has been updated to the `version_scheme` one

I'll close all Python 3.8_ syntax update suggestion for easier review

commitizen/commands/changelog.py
noirbizarre: Other methods don't use this syntax. I think this one should be handled in another PR setting `ruff` and `mypy` accordingly.

commitizen/formats/__init__.py
noirbizarre: Done:
- `DEFAULT_FORMAT` moved in `commitizen.defaults.CHANGELOG_FORMAT`
- `DEFAULT_TEMPLATE_EXTENSION` just renamed into `TEMPLATE_EXTENSION` as this is not a setting and so not a default value, just here to avoid a magic value or an hardcoded extension

commitizen/formats/__init__.py
noirbizarre: Done 👍🏼

commitizen/defaults.py
noirbizarre: Renamed into `changelog_format` 👍🏼

commitizen/defaults.py
noirbizarre: Renamed into `changelog_format` 👍🏼

commitizen/formats/errors.py
noirbizarre: Done 👍🏼 
The `commitizen.changelog_format.errors` has been removed and `ChangelogFormatUnknown` moved into `commitizen.exceptions` (only exception really used) and `exit_code` has been added (doc was not up to date on those, so I also updated it)

Lee-W:
@noirbizarre I think we're good to merge this one 🎉  but after merging https://github.com/commitizen-tools/commitizen/pull/889, it seems to have some conflicts 😱  Please merge it once you resolve the conflict. Thanks so much for the hard work! this is a really exciting feature!

noirbizarre:
Done 👍🏼 
🤞🏼

noirbizarre:
This is driving me crazy !!! 😱 
Windows build are failing on pre-commit (not touched in the PR and not related to the PR).
I think the last `pre-commit` update is broken on Windows runners. Excluding this version seems to work.
I think I'll do another separate PR to fix the pre-commit test case by using a check which is not relying on NodeJS (will be faster to execute and safer). Most probably a check based on ruff to have both speed (install and exec) and no dependency management to handle in pre-commit.

woile:
Awesome news 🎉 🎉 🚀 
Thanks everyone!

Issue #641, opened 2022-12-16
setuptools-scm compatibility


jaustinpage:
### Description

I have had a lot of success in the past with using setuptools-scm. One nice feature is that it can be used in a way where you have a single source of truth as to what version you are on (via a tag). However, some features that setuptools-scm does not help with are having a good way to walk users through commit message writing, and having logic for how to bump from version to version based on those commit messages. When trying to run projects for large user bases and varied programming experience, the extra help goes a long way to easing pain for developers.

### Possible Solution

In a perfect world, I would be able to use both tools. One possibility might work like this:
  If commitizen is unable to find a current version number using its config mechanisms, and it detects that setuptools-scm is in use (maybe by checking the pyproject.toml['build-system']['requires'] for setuptools-scm?), then it falls back to using setuptools-scm's get_version() for the current version number. I would also be happy with an explicit flag or config line to specify to run in "setuptools-scm compatibility mode".

  It would also be required that bump not try to save the version number to the config file when it is in this compatibility mode, but instead recompute after tagging with an additional call to setuptools-scm get_version() function to update the current version stored in the config.
    
I would also be happy with a way to extend the behavior of bump, perhaps something like https://github.com/commitizen-tools/commitizen/issues/292 or some sort of other hooks for plugins might be useful for allowing people to extend this kind of functionality.

### Additional context

potential issues:

It would be pretty easy for someone to try to write versions from both commitizen and setuptools_scm to the same files. both tools try to do this automatically in many cases, so the potential for conflict is definitely there.

### Additional context

Thanks so much for taking the time to read this, and, after the discussion, i am happy to work on some pr's. I also recognize that the ethos of commitizen and setuptools_scm is a bit different from each other, but, I hope that maybe both together makes for an even better development environment.

jaustinpage:
I think the "discussion" label might be good for this one...

Lee-W:
Hi @jaustinpage sorry for late reply. I'm not that familiar with setuptools-scm. So my understanding of this feature is to use setuptools-scm to maintain the project version and use other commitzien. Is my understanding correct?

jaustinpage:
@Lee-W : your understanding is fundamentally correct.

In this mode:

Setuptools-scm would be the source of truth for what the "current version"  is. 

[...]

PR #646, opened 2022-12-28
Feature: version providers

noirbizarre:
## Description

This PR is a proposal because I know it has been stated than there won't be an official PEP621.

This PR doesn't focus on `PEP621` but on opening the version source of trust by introducing a version provider.

If has been splitted in multiple commits that can be removed from the PR depending on your acceptance.

The first one is simply introducing the entrypoint `commitizen.provider` and the `commitizen` provider which is the current scheme (version taken from commitizen configuration) and backward compatible.

The second one introduce some common TOML-based formats (PEP621, poetry, Cargo) as `tomlkit` is style preserving. Those implementations does not suffer from the regexp side-effect aka. changing any field matching the criteria (happen as soon as you have another version matching in the file) because they are using the format structured access instead of search and replace.

The third one introduce the JSON-based providers for some formats (npm/package.json, composer), mostly to showcase the possibility (and to echo the [FAQ entry](https://commitizen-tools.github.io/commitizen/faq/#support-for-pep621) for package.json/npm projects).
Given the builtin `json` is not style preserving (I assume an indentation for each one), I would totally understand their removal.

The fourth one is simply a documentation update to be able to have proper snippets with title in the documentation.

The last one is exactly #647 but was required to make this PR CI green. See [the previous build](https://github.com/commitizen-tools/commitizen/actions/runs/3796117778/jobs/6455902375) failing on some totally unrelated (and untouched) cases.

This PR is backward compatible (aka. it is possible not to change the version provider as well as using the `pep621` provider while handling both `Cargo.toml` and `package.json` with `version_files`), tested and documented.

Note: this PR use the same extension mechanism as #632 and it is totally possible to extract all implementations to only keep the `commitizen` one given they can be provided as an external package.

Edit: Added an extra commit to add an `scm` provider for #641. It handles `tag_format` reverse parsing. I discovered an issue (I will submit) on the [`tag_format` handling](https://commitizen-tools.github.io/commitizen/bump/#tag_format): while `devrelease` is documented as being a variable placeholder in `tag_format`, it is not handled as a template variable but appended to the tag (`.devX` when submitted as `--devrelease X`) whatever it's position is in the template, even if it is not present in the template.

## Checklist

### DIFF
--- a/commitizen/providers.py
+++ b/commitizen/providers.py
@@ -1,15 +1,18 @@
 from __future__ import annotations
 
 import json
+import re
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, ClassVar, cast
+from typing import Any, Callable, ClassVar, Optional, cast
 
 import importlib_metadata as metadata
 import tomlkit
+from packaging.version import VERSION_PATTERN, Version
 
 from commitizen.config.base_config import BaseConfig
 from commitizen.exceptions import VersionProviderUnknown
+from commitizen.git import get_tags
 
 PROVIDER_ENTRYPOINT = "commitizen.provider"
 DEFAULT_PROVIDER = "commitizen"
@@ -161,6 +164,65 @@ class ComposerProvider(JsonProvider):
 
     filename = "composer.json"
     indent = 4
+
+
+class ScmProvider(VersionProvider):
+    """
+    A provider fetching the current/last version from the repository history
+
+    The version is fetched using `git describe` and is never set.
+
+    It is meant for `setuptools-scm` or any package manager `*-scm` provider.
+    """
+
+    TAG_FORMAT_REGEXS = {
+        "$version": r"(?P<version>.+)",
+        "$major": r"(?P<major>\d+)",
+        "$minor": r"(?P<minor>\d+)",
+        "$patch": r"(?P<patch>\d+)",
+        "$prerelease": r"(?P<prerelease>\w+\d+)?",
+        "$devrelease": r"(?P<devrelease>\.dev\d+)?",
+    }
+
+    def _tag_format_matcher(self) -> Callable[[str], Optional[str]]:
+        pattern = self.config.settings.get("tag_format") or VERSION_PATTERN
+        for var, tag_pattern in self.TAG_FORMAT_REGEXS.items():
+            pattern = pattern.replace(var, tag_pattern)
+
+        regex = re.compile(f"^{pattern}$", re.VERBOSE)
+
+        def matcher(tag: str) -> Optional[str]:
+            match = regex.match(tag)
+            if not match:
+                return None
+            groups = match.groupdict()
+            if "version" in groups:
+                return groups["version"]
+            elif "major" in groups:
+                return "".join(
+                    (
+                        groups["major"],
+                        f".{groups['minor']}" if groups.get("minor") else "",
+                        f".{groups['patch']}" if groups.get("patch") else "",
+                        groups["prerelease"] if groups.get("prerelease") else "",
+                        groups["devrelease"] if groups.get("devrelease") else "",
+                    )
+                )
+            elif pattern == VERSION_PATTERN:
+                return str(Version(tag))
+            return None
+
+        return matcher
+
+    def get_version(self) -> str:
+        matcher = self._tag_format_matcher()
+        return next(
+            (cast(str, matcher(t.name)) for t in get_tags() if matcher(t.name)), "0.0.0"
+        )
+
+    def set_version(self, version: str):
+        # Not necessary
+        pass
 
 
 def get_provider(config: BaseConfig) -> VersionProvider:
--- a/commitizen/providers.py
+++ b/commitizen/providers.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
+import json
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import ClassVar, cast
+from typing import Any, ClassVar, cast
 
 import importlib_metadata as metadata
 import tomlkit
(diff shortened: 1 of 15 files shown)


woile:
Wow! Thanks, it's an interesting idea, @Lee-W what do you think? I think it could definitely work.

docs/faq.md
```diff
@@ -8,23 +8,16 @@ name = "spam"
...
-
-Instead of supporting all the different project files, you can use `version_files`
-inside `[tool.commitizen]`, and it will cheaply keep any of these project files in sync
+Commitizen provides a [`pep621` version provider](config.md#version-providers) to get and set version from this field.
+You just need to set the proper `version_provider` setting:
 
 ```toml
+[project]
+name = "spam"
+version = "2020.0.0"
```
woile, line 17: do we support this kind of version?

woile:
Do you think we could do something like this for this https://github.com/commitizen-tools/commitizen/issues/150 ?
It would be interesting if it works as pluggins, and by default we support semver + pep440

docs/faq.md
noirbizarre: I should change the example. `2020` is an integer, so `commitizen` can work with this. The problem is with `calver` version management, not really specified (each project having its own rules) and I don't want people to think this is properly supported just with this example.
Note: but adding `calver` support by using the same pattern for a `VersionSchemProvider` could be a good idea for later

noirbizarre:
Yes, I think this pattern can also be applied to customize the versioning scheme. We can definitively imagine another feature splitting the versioning schemes into a `commitizen.scheme` entrypoint plugins too (I don't want to tease, I need deeper investigation but if this is OK for you, I can definitively try working on this later)

docs/faq.md
noirbizarre: Version changed to something not looking like CalVer 👍🏼

Lee-W:
> Wow! Thanks, it's an interesting idea, @Lee-W what do you think? I think it could definitely work.

I'm really interested in this one! But kinda out of bandwidth to take a deeper look 🤦‍♂️ Will do so when I'm not that busy

davidlday:
I came across this PR while looking for a way to ensure cz doesn't clobber dependency versions in `package-lock.json`. Using a JSON query solves it, as you'd implemented in the JsonProvider. Would love to see this PR accepted!

Lee-W (commented):
Hi @noirbizarre , it's way more interesting that I previously thought! I've browsed through it and found no big issue. Wondering if there's any review order you want us to do between 645 and 646?

Lee-W:
I've also rebased the latest master back to v3 which leads to a few conflicts 🤦‍♂️

noirbizarre:
There is no order or dependency between #645 and this one, but I would start with this one because it is ready to review.
Rebased on updated `v3` branch, conflict resolved 👍🏼

Lee-W:
Thanks for the prompt reply! Let me start with this one 👍

commitizen/providers.py
```diff
@@ -0,0 +1,234 @@
...
+
+import importlib_metadata as metadata
+import tomlkit
+from packaging.version import VERSION_PATTERN, Version
+
+from commitizen.config.base_config import BaseConfig
+from commitizen.exceptions import VersionProviderUnknown
+from commitizen.git import get_tags
+
+ENTRYPOINT = "commitizen.provider"
```
Lee-W, line 17: Should we name it as `PROVIDER_ENTRYPOINT` for explicitness?

commitizen/providers.py
```diff
@@ -0,0 +1,234 @@
...
+DEFAULT = "commitizen"
```
Lee-W, line 18: Should we name it as `DEFAULT_PROVIDER` for explicitness?

tests/commands/test_version_command.py
```diff
@@ -70,3 +74,30 @@ def test_version_for_showing_commitizen_system_info(config, capsys):
...
+    version = "0.0.0"
+    mock = mocker.MagicMock(name="provider")
+    mock.get_version.return_value = version
+    get_provider = mocker.patch(
+        "commitizen.commands.version.get_provider", return_value=mock
+    )
+
+    commands.Version(
+        config,
+        {"report": False, "project": project, "commitizen": False, "verbose": True},
```
Lee-W, line 96: I'd suggest we changing the test case here. These 4 are in the same [exclusive_group](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/cli.py#L315) and thus only one of them can be `True`. Although it does not affect the functionality, it'd still be better if we could keep the test cases closer to our use cases. Something like `@pytest.mark.parametrize("project, verbose", ((True, False), (False, True)))` would work. Thanks!

tests/test_version_providers.py
```diff
@@ -0,0 +1,248 @@
...
+    assert isinstance(provider, CommitizenProvider)
+
+
+def test_raise_for_unknown_provider(config: BaseConfig):
+    config.settings["version_provider"] = "unknown"
+    with pytest.raises(VersionProviderUnknown):
+        get_provider(config)
+
+
+def test_commitizen_provider(config: BaseConfig, mocker: MockerFixture):
```
Lee-W, line 48: nitpick: I notice we have some duplicate code for providers other than scm-provider. Should we use `parametrize` to de duplicate? I'm good with the current implementation. But it would be even better if we could do that :)

Lee-W (commented):
@woile I think we're almost ready to merge this one except for a few minor changes. I'm planning on merging it this week. Let me know if you want to take a deeper look

commitizen/providers.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class VersionProvider(ABC):
```
woile, line 21: What do you think of using protocols?

tests/commands/test_version_command.py
noirbizarre: Solved with `verbose = not project`

tests/test_version_providers.py
noirbizarre: Done, I parametrized the file providers tests

commitizen/providers.py
noirbizarre: In this precise case, I prefer `ABC` over `Protocol` because:
- I expect both a constructor and methods (I could move the `config` in the method signatures, but this makes more verbose code without benefits in my opinion)
- it is intended for user/community to extend this, using `ABC` force to implement methods and raise an explicit error if not, which I think is easier to understand (so more discoverable for them, less support for you)

But if you prefer `Protocol` over `ABC` I can switch 👍🏼

Lee-W (approved):
Thanks @noirbizarre for the prompt update! I'm good with both `ABC` and `Protocol`. Once that discussion is resolved, I think we're good to merge this one

commitizen/providers.py
Lee-W: @woile What do you think?

woile: I prefer `Protocol` 🤣 but using `ABC` is fine as well. Point 2 is the key, with `Protocols` people would have to use `mypy` to catch the problems, I like the explicit error.

noirbizarre:
Rebased, conflicts solved. ✅

Can we merge this one please 🙏🏼 ? (because, without blaming anyone, it's starting to represent a lot of extra work for me to maintain those branches up to date for the past 2 months, aka. almost weekly rebasing and solving conflicts 😅 and for the moment I wait for this one to be merged to finish the templating part in #645 as well as the versioning scheme to coming after to avoid increasing the amount of conflicts)

Lee-W (approved):
Thanks @noirbizarre for your prompt update! 

@woile I'm planning on merging it today. Let me know if you have any thought :)

woile:
Let's go! thanks for the awesome work!

PR #638, opened 2022-12-13
test(git): ensures that global git user settings are not interfering with the tests

noirbizarre:
## Description

This PR create a git sanbox to isolated from the host configuration.
As example, in my case given I have a default signin key, without this fix, pre-commit fail trying to prompt for gpg credentials.
This ensure that test are run on a default configuration with `user.name` and `user.email` already set and nothing else.

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes (N/A)

## Expected behavior

Custom git user configuration does not interefere with the test suite anymore

## Steps to Test This Pull Request

Run the tests with any change in you git config

## Additional context

Submitted on `v3` because of the working CI (and it is my current base for the next PRs) but the issue already exists on `master`

PR #647, opened 2022-12-30
test: improve git isolation and fixes test missing a repository

noirbizarre:
## Description
This PR follows #638 and extend the isolation to any `GIT_` prefixed variables. It also ensure that every test requiring a `tmp_git_project` or `tmp_commitizen_project`  fixture has it, even for test expecting a failure (to ensure the proper failure).

This fixes some false negatives in `pre-commit` as well as ordering side-effects for tests that were missing the fixture like [this build](https://github.com/commitizen-tools/commitizen/actions/runs/3796117778/jobs/6455902375) failing on some totally unrelated (and untouched) cases.

There is few misc fixes and redundancy removal (lots of unnecessary nested `.as_cwd()` calls as well as some fixtures extracted into `@pytest.mark.usefixtures()` when not used in the test but still requiring the fixture setup))

## Checklist

- [x] Add test cases to all the changes you introduce (N/A)
- [x] Update the documentation for the changes (N/A)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,12 @@ SIGNER_MAIL = "action@github.com"
 @pytest.fixture(autouse=True)
 def git_sandbox(monkeypatch: pytest.MonkeyPatch, tmp_path: Path):
     """Ensure git commands are executed without the current user settings"""
+    # Clear any GIT_ prefixed environment variable
+    for var in os.environ:
+        if var.startswith("GIT_"):
+            monkeypatch.delenv(var)
+
+    # Define a dedicated temporary git config
     monkeypatch.setenv("GIT_CONFIG_GLOBAL", str(tmp_path / "gitconfig"))
     cmd.run(f"git config --global user.name {SIGNER}")
     cmd.run(f"git config --global user.email {SIGNER_MAIL}")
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -15,6 +15,12 @@
 @pytest.fixture(autouse=True)
 def git_sandbox(monkeypatch: pytest.MonkeyPatch, tmp_path: Path):
     """Ensure git commands are executed without the current user settings"""
+    # Clear any GIT_ prefixed environment variable
+    for var in os.environ:
+        if var.startswith("GIT_"):
+            monkeypatch.delenv(var)
```
Lee-W, line 21: Will it affect user's environment?

Lee-W (commented):
Just one minor question. I think this one is almost ready to merge

tests/conftest.py
noirbizarre: No, it is only for the tests using the fixture (all given the fixture has `autouse=True`. 
It isolate from user-defined `GIT_` variables which are interfering with test, especially in `pre-commit` where some of those are set by `pre-commit` .
The `monkeypatch` fixture ensures that:
- this is contained to the current test and does not leak on other tests nor to the user environement
- this is properly cleaned at the end of the test

noirbizarre:
Yes, #638 was using a temporary `GIT_CONFIG_GLOBAL` ensuring the clean configuration.
This PR extends the isolation to all `GIT_` environment variables because some settings can be defined by both the configuration file and the environment.
`pre-commit` is interfering with some of those, the `git` command is also exporting some depending on the version you use (see: https://github.com/pre-commit/pre-commit/blob/1f59f4cba8241ed7e572a1b5818bba74c6f31181/pre_commit/git.py#L27-L48), your shell can export some, your custom promt too... 
That's why without isolation you may have different behavior and some false negative depending on wether you are running in `pre-commit` or not, depending on your `pre-commit` version, depending on your `git` version, or depending of your own custom shell...
With this change, you ensure consistent and reproducible behavior on all environments, they always run with the minimal configuration (`author` and `email`) and no setting defined outside of the test case

Lee-W:
@noirbizarre Cool! many thanks for the detailed explanation!

Issue #593, opened 2022-09-21
cz check range sometimes failed in GitLab CI


SsuperL:
### Description

Run`cz check --range origin/master..HEAD` to check commits in CI occasionally fails.
And it worked after I retried the job.But sometimes it didn't work after retrying multiple times.
I couldn't find out the reason ,

### Steps to reproduce

lint:
  retry: 1
  script:
    - cz version
    - cz check --rev-range origin/master..HEAD

### Current behavior

`cz check --rev-range origin/master..HEAD` failed. Outputs attached below:

      fatal: ambiguous argument 'master..HEAD': unknown revision or path not in the working tree.
      Use '--' to separate paths from revisions, like this:
      'git <command> [<revision>...] -- [<file>...]'

### Desired behavior

`cz check --rev-range` can run successfully in gitlab CI.

### Screenshots

![image](https://user-images.githubusercontent.com/40085989/191435749-b4f7304e-f78c-4c86-b18e-bcaaedacd090.png)

### Environment

commitizen version:  2.34.0
python version: 3.7.5
operating system: Linux

Lee-W:
Is it possible that the git repo doesn't setup remote 🤔

SsuperL:
I run command `git remote -v` in the CI , and it shows that the remote has been setup.

SsuperL:
![image](https://user-images.githubusercontent.com/40085989/191940776-60422858-4243-47a6-980c-6a22df76097c.png)

Lee-W:
Hmmm... As I'm not able to reproduce it, I don't have any clue on how we could fix it. But we can keep it open till you or someone encounter  this issue again with more detail

SsuperL:
Actually,this issue has come up again.

SsuperL:
![image](https://user-images.githubusercontent.com/40085989/192184140-09642a6e-0278-409f-b3f1-4a7824a53616.png)
And when I merged the branch into master,it said that there is no commit found range `origin/master..HEAD`.

Lee-W:
Do you mean after merging back to the master branch? If that's the case, your `HEAD` will on on `master` so the result seem to be expected.

SsuperL:
> 



> Do you mean after merging back to the master branch? If that's the case, your `HEAD` will on on `master` so the result seem to be expected.

yes. Or can it skip check when there's no commit rather than make the job failed?

[...]

PR #649, opened 2023-01-02
feat: Make `cz bump` work on shallow clones

robertschweizer:
## Description

Try to `git fetch` missing tags if working on a shallow clone.

This allows reliably setting up your CI to only fetch the latest changes. E.g. Gitlab CI uses shallow clones with a depth of 20 by default: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/77576

Only `tag_exists()` had to be modified to make this work for `cz bump`. Other functions are not ready yet, more functions would have to include a `git fetch` for that. There are also other problems still with Gitlab CI's shallow clones, e.g. https://github.com/commitizen-tools/commitizen/issues/593. But for my use-case of just running `cz bump` and `cz check` in Gitlab CI this works nicely.

## Checklist

- [ ] Update the documentation for the changes  -> The printed warning makes this unnecessary.

## Additional context

I was worried about slowing down error reporting in non-CI workflows ("git fetch" calls can sometimes take a while if your connection is bad). To minimize the risk for this, I'm now only fetching the tag for shallow clones and always printing a warning when this happens.

I was hit by this issue while trying out Commitizen in Gitlab CI on a monorepo, where some packages haven't seen a release for >50 commits. It's also large enough for the shallow clone to really make sense in the CI.

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -168,7 +168,22 @@ def get_tags(dateformat: str = "%Y-%m-%d") -> List[GitTag]:
 
 def tag_exist(tag: str) -> bool:
     c = cmd.run(f"git tag --list {tag}")
-    return tag in c.out
+    if tag in c.out:
+        return True
+
+    # In shallow clones (e.g. set up by Gitlab CI), the previous release tag might not
+    # be available locally, so try and fetch it.
+    if cmd.run("git rev-parse --is-shallow-repository").out.strip() == "true":
+        out.warn(
+            f"Could not find tag {tag} locally. Since this is a shallow clone, "
+            "trying to fetch it from all available remotes now."
+        )
+        for remote in cmd.run("git remote").out.strip().splitlines():
+            res = cmd.run(f"git fetch --depth=1 {remote} tag {tag}")
+            if res.return_code == 0:
+                return True
+
+    return False
 
 
 def is_signed_tag(tag: str) -> bool:
--- a/tests/commands/test_bump_command.py
+++ b/tests/commands/test_bump_command.py
@@ -1,5 +1,8 @@
 import inspect
+import os
+import re
 import sys
+from pathlib import Path
 from typing import Tuple
 from unittest.mock import MagicMock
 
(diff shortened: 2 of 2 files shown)


robertschweizer:
It doesn't look like the CI tests failed because of my change. Please let me know if that's the case in your opinion.

Lee-W:
@robertschweizer I just reran the CI and it passed. Will take a deeper look. Thanks!

tests/commands/test_bump_command.py
```diff
@@ -377,6 +380,32 @@ def test_bump_major_version_zero_when_major_is_not_zero(mocker, tmp_commitizen_p
...
+    cmd.run(f"git clone --depth=1 file://{tmp_commitizen_project} {cloned_repo}")
+    os.chdir(cloned_repo)
+
+    # Bumping should work, but with a warning that we are pulling the tag from remote.
+    testargs = ["cz", "bump"]
+    mocker.patch.object(sys, "argv", testargs)
+    capsys.readouterr()  # Reset capsys fixture
+    cli.main()
+    stdout, _stderr = capsys.readouterr()
+    assert re.search("Could not find tag.*trying to fetch", stdout)
```
Lee-W, line 406: Should we also check whether the function returns correctly after fetching?

commitizen/git.py
```diff
@@ -168,7 +168,22 @@ def get_tags(dateformat: str = "%Y-%m-%d") -> List[GitTag]:
 
 def tag_exist(tag: str) -> bool:
     c = cmd.run(f"git tag --list {tag}")
-    return tag in c.out
+    if tag in c.out:
+        return True
+
+    # In shallow clones (e.g. set up by Gitlab CI), the previous release tag might not
```
Lee-W, line 174: I'm good with this functionality. But one of the concerns I have is whether this is an unexpected action from the user point of view. I'm thinking of using a flag.

woile: Agreed, this is more CI configuration, is up to the user, I'm fine with a warning and an explanation with the next steps to the user. But the user should ensure the environment (git) is providing commitizen what's needed, not the other way around.

robertschweizer: Thanks for the reviews!

I agree with your concerns that auto-fetching missing tags is unexpected from the user point of view. Actually it would be cleanest to me if `git tag --list {tag}` had a flag to do this fetching for shallow clones automatically. But without auto-fetching missing tags in shallow clones, the only option would be to do a full clone, which can be quite slow for large/old repositories.

I believe both Gitlab CI and Github Actions use shallow clones by default. Because of that, Commitizen should ideally handle shallow clones out-of-the-box without setting an extra flag. Do you have examples which workflows could break if we fetch in shallow clones by default? I'm mostly worried about connection issues/the CI not setting up the Git remote for later fetching.

If you keep not being convinced, should we call the flag `--auto-fetch` and only add it for the `cz bump` command for now?

Lee-W: Hi @robertschweizer Sorry for late reply. I like the idea of the flag `--auto-fetch`. @woile What do you think?

woile: Yes, a flag and a setting. We can prompt the user in the init command

robertschweizer:
Thanks a lot for your review and feedback! I'm sorry about this, but I actually don't need this change anymore.

In our fork we actually restrict `git log` output to the CWD (`.`) to run commitizen on monorepo folders. It turns out filtering logs by directory does not work with shallow clones unless you unshallow the complete history up to the reference tag. Combine that with Gitlab's unstable behavior when doing a lot of unshallowing made us go back to using full clones for this CI job.

Feel free to close this MR unless someone else finds it useful.

PR #657, opened 2023-01-20
docs: add commit cli

### LINKED ISSUE #417: remove coloring from pre-commit prints
## Description
I use the pre-commit framework to apply some sanity check before committing to my repositories. I recently discover this lib and wanted to give it a try. Fantastic work ! 

I use te following command to create a new commit: 
```
cz commit
```
 And I get the following results (sorry for the image but everything is about coloring):  
Capture d’écran 2021-09-07 à 16 41 00

after `#258`which is the commit message, the pre-commit checks start and normally at the end of their lines `skipped`should be blue, `passed`: green and `failed` red. when creating a commit with `cz`I loose these colors. I tested back using simply `git commit`and it worked. 

Question then: why is the coloring of `pre-commit` disabled by `cz` ?

12rambau:
## Description
Create a dedicated page for the `commit` command and document the drop of hooks coloring.
Fix #417  

## Checklist

check the new doc :-)


docs/commit.md
```diff
@@ -0,0 +1,11 @@
...
+
+## About
+
+In your terminal run `cz commit` or the shortcut `cz c` to generate a guided git commit.
+
+A commit can be signed off using `cz commit --signoff` or the shortcut `cz commit -s`.
+
+!!! note
+    To maintain platform compatibility, the `commit` command disable ANSI escaping in its output.
+    In particular pre-commit hooks coloring will be deactivated.
```
Lee-W, line 11: I think we can add a link to the original issue.

Lee-W (changes requested):
Thanks @12rambau . most of the part looks good to me , but could you please remove `.DS_Store`? Thanks!

.gitignore
```diff
@@ -110,3 +110,6 @@ venv.bak/
 
 # build
 poetry.lock
+
+# macOSX
```
Lee-W, line 114: I'd suggest we don't add it to project `.gitignore`. It should be added to somewhere like `~/.gitignore` based on platform's need

Lee-W (changes requested):

.gitignore
12rambau: Before removing it let me fight a bit for this modification: 

`.DS_Store` files are generated by mac computer all the time. Lucky me I didn’t execute the lib itself so there is just one but they can appear in any folder. All Mac IDE are hiding them from us meaning that before `git add` I simply don’t know where they are. If you refuse to add this line to the .gitignore you simply deny to all mac users the possibility to use `git add .` which can be a real pain when modifying many files. 

Let’s have a look at the `.gitignore` itself. it’s not a « commitizen » specific file but the github template for Python language with small customization. It includes: 
- IDE specific folders (.vscode)
- sphinx folders (docs/_build/)
- all the django related files
- translation files (.mo, .po) 
- scrappy stuff 
- All the distribution files including things unrelated to the project
- unitesting from the most known tools (coverage, nosetest and tox) even though tox is not part of your lib

meaning that the `.gitignore` already includes things that no developer will ever need as it’s unrelated to the project. I don’t understand why 1 extra line that will be useful for all mac users seems too much. it’s present in all the open-source projects I contribute in like: [opensartoolkit](https://github.com/ESA-PhiLab/OpenSarToolkit/blob/main/.gitignore), [pydata-sphinx-theme](https://github.com/12rambau/pydata-sphinx-theme/blob/main/.gitignore), [sepal_ui](https://github.com/12rambau/sepal_ui/blob/main/.gitignore), [pygadm](https://github.com/12rambau/pygadm/blob/main/.gitignore)

let me know if even after my little speech you still want me to remove it and I will. 😢

Lee-W: Hi @12rambau thanks for sharing the thought in such detail! **tl;dr, you convinced me to accept it in this project.**

I just found out we also add `.idea` and `.vscode` into the `.gitignore`. Thus, adding `.DS_Store` could make sense to this project.

Not adding these stuff is kinda my personal preference which was suggested by others back to the time I contributed to git-extras. They suggested me adding these config to my local development environment. Otherwise, we'll need to add lots and lots different kinds of temp file and config into `.gitignore`. Just share the reason why I asked you to remove it at the first place 🙂

Lee-W (approved):

PR #663, opened 2023-02-08
use shield.io badge for github workflow

12rambau:
The github badge was the only one not in flat-square style/not the sae green, not with the same logo color


12rambau:
superceeded by #664

PR #675, opened 2023-02-23
fix(bump): fixed environment variables in bump hooks

skoef:
During development of #644, merging the hook environment variables with existing OS variables broke. This commit fixes that. I think it got missing during an additional rebase of that MR, but it is fixed again.


skoef:
@Lee-W perhaps you could take a look at it?

Lee-W (approved):

PR #589, opened 2022-09-17
Relax specification for argcomplete dependency

jakob-keller:
## Description

Closes #588 

## Checklist

- [ ] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [ ] Test the changes on the local machine manually
- [ ] Update the documentation for the changes

## Expected behavior

Allow for installation of argcomplete 2.0.x in commitizen enabled projects

PR #680, opened 2023-03-08
Relax specification for argcomplete dependency

jakob-keller:
## Description

Similar to #589

## Checklist

## Expected behavior

Allow for installation of argcomplete 2.1.x in commitizen enabled projects

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ packaging = ">=19"
 tomlkit = ">=0.5.3,<1.0.0"
 jinja2 = ">=2.10.3"
 pyyaml = ">=3.08"
-argcomplete = ">=1.12.1,<2.1"
+argcomplete = ">=1.12.1,<2.2"
 typing-extensions = "^4.0.1"
 charset-normalizer = ">=2.1.0,<3.1"
 


nejch:
Thanks for this @jakob-keller, @woile. :bow: I guess this will be in the next release within a few weeks judging from the release cadence here.

In general, argcomplete seems to follow semver, would it make sense to relax this to v2 entirely? Not sure if it was intentional in https://github.com/commitizen-tools/commitizen/commit/92b1e8900686affa97fe92cb11a1e8f8b183e7ba to restrict minor versions.

This sometimes gets installed in environments along with other dependencies so just to reduce the number of times we get conflicts.

jakob-keller:
> In general, argcomplete seems to follow semver, would it make sense to relax this to v2 entirely? Not sure if it was intentional in [92b1e89](https://github.com/commitizen-tools/commitizen/commit/92b1e8900686affa97fe92cb11a1e8f8b183e7ba) to restrict minor versions.
> 
> This sometimes gets installed in environments along with other dependencies so just to reduce the number of times we get conflicts.

I agree. My intention here was to relax the restriction without questioning the generally restrictive way dependencies are specified in this repo. But this would be a great request, i.e. a separate issue IMO.

nejch:
Thanks @jakob-keller. I've opened https://github.com/commitizen-tools/commitizen/issues/683, feel free to chime in :)

Issue #665, opened 2023-02-09
Does not work with python 3.6


gabohc:
### Type

- [X] Content inaccurate
- [ ] Content missing
- [ ] Typo

### URL

https://commitizen-tools.github.io/commitizen/

### Description

According to the docs, it requires python 3.6 at minimum.

However, `hooks.py` uses `from __future__ import annotations`, which is [not supported by python 3.6](https://docs.python.org/3.6/library/__future__.html)

Lee-W:
@woile It seems I accidentally merged a branch that would break 3.6. I'm thinking of merging `v3` and explicitly dropping 3.6 support

woile:
Agreed, py36 support was dropped officially, we should drop it and maybe next year also 3.7. I remember a discussion or PR to drop 3.6 somewhere, right?

Lee-W:
Yep, if my memory serves me right, we decide to do so in `v3`. As this sounds like a breaking change to me. `v3` might be a reasonable option

woile:
We no longer support python 3.6

PR #684, opened 2023-03-11
Relax runtime dependency specifications

### LINKED ISSUE #683: Relax version specification for dependencies following semver
### Description

Currently quite a few prod dependencies are defined in very strict ranges, even though commitizen is likely to be installed more as a library along with other dependencies rather than a single application. See:

https://github.com/commitizen-tools/commitizen/blob/fb0d1ebacb29b3ca0ea33a349dbb8064ae6c72fe/pyproject.toml#L60-L73

Partially I would blame poetry's default behavior for adding dependencies, which is more restrictive than you'd really want for libraries IMO. So I'd personally use PEP-440 specifiers here rather than poetry's custom caret versioning.

For dependencies that reliably follow semver, I think it's not an issue to at least allow the current major version, or even open it up completely and restrict versions only when there are issues, which is more common in python IMO (this is not npm so we cannot have multiple versions of transitive dependencies at the same time.. so we kind of have to be a bit more broad with ranges to make life easier).

/cc @jakob-keller 

### Possible Solution

Check the current production dependencies and see if they follow semantic versioning. If they do, assume good intentions and lock only major versions for those ;)

### Additional context

_No response_

### Additional context

https://github.com/commitizen-tools/commitizen/pull/680

jakob-keller:
## Description

Relaxes overly strict runtime dependency specifications for semver packages.

## Checklist

## Expected behavior

Allows for `commitizen` to be installed alongside current versions of runtime dependencies.

## Additional context

Closes #683

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,10 +51,7 @@ python = "^3.6.2"
 questionary = "^1.4.0"
 decli = "^0.5.2"
 colorama = "^0.4.1"
-termcolor = [
-    { "version" = "^1.1", python = "< 3.7" },
-    { "version" = ">= 1.1, < 3", python = ">= 3.7" },
-]
+termcolor = ">=1.1,<3"
 packaging = ">=19"
 tomlkit = ">=0.5.3,<1.0.0"
 jinja2 = ">=2.10.3"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ tomlkit = ">=0.5.3,<1.0.0"
 jinja2 = ">=2.10.3"
 pyyaml = ">=3.08"
 argcomplete = ">=1.12.1,<2.2"
-typing-extensions = "^4.0.1"
+typing-extensions = { version = "^4.0.1", python = "<3.8" }
 charset-normalizer = ">=2.1.0,<3.1"
 
 [tool.poetry.dev-dependencies]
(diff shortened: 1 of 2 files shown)


jakob-keller:
`argcomplete` does not appear to claim semver compliance. I suggest we leave that dependency as it is. What do you think?

jakob-keller:
Support for Python 3.6 appears to be broken according to #665.

Do you want me to bump the required Python version to 3.7 as part of this PR? I think we should.

jakob-keller:
I think it is time to look at the dev dependencies, too. Want be to have a look at that as well?

pyproject.toml
```diff
@@ -42,35 +42,23 @@ classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Console",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: MIT License",
     "Natural Language :: English",
     "Operating System :: OS Independent",
-    "Programming Language :: Python",
```
woile, line 48: Are these tags not required?

jakob-keller: poetry attaches relevant Python 3 specific tags automatically. `Programming Language :: Python` is not attached automatically, nor is `Programming Language :: Python :: Implementation :: CPython`. I assume they are not required for anything, but can add them back in, if you like.

woile: No, no need, thanks for sharing, I was not aware of this.

Issue #519, opened 2022-05-24
generate changelog only from tags matching specific patterns


bhelgs:
### Description

It seems right now the changelog generator is capturing all tags, rather than just the tags matching the version pattern.
It was mentioned in [this `TODO` in the code](https://github.com/commitizen-tools/commitizen/blob/afa0d93544405bb0a871b5072e0a814720d933be/commitizen/changelog.py#L106).

Without filtering tags:
* A user might create a pre-release tag `v2.27.0-alpha` and then later come to find it in their changelog.
    * it seems they would have to delete the tag 
       otherwise they see `v2.26.0...v2.27.0-alpha...v2.27.0` rather than `v2.26.0...v2.27.0`
    * (see "Additional context" section for more info if necessary)
* It prevents users from having some sort of tagging practices unrelated to "release versions"
    * (see "Additional context" section for more info if necessary)

### Possible Solution

1.  Allow users to specify a custom tag format via configuration regex.
    ``` toml
    [tool.commitizen]
    # tag_format = "v$major.$minor.$patch$prerelease" # already used by "bump" (nothing new proposed)
    tag_parser = "v[0-9]*\.[0-9]\.[0-9]*" # proposed: filter for "released" versions only.
                                                             # tag_parser mirrors tag_format used by bump.
    ```
2. Mimic current functionality by default (base commitizen):
   * match every tag by setting a default `.*` regex.
   
I believe this is an alternate solution to what was discussed on #364 and may be more extensible, for example:
* The user could just edit `tag_parser` if they want multiple changelogs with different parse criteria
    * avoids the creating custom CLI that was proposed here: #454
* Capture each field within the version for use in other parts of `changelog` module:
    * `v(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.(?P<patch>[0-9]+)(?P<pre>[0-9A-Za-z-]?`
* The user may have changed their versioning pattern at some point and the parser would allow for that:
   * `(v[0-9.]*|release-[0-9.]*)`  (pattern matches `v1.2.0` or `release-1.2.0`)

### Additional context

A user may have created a pre-release tag not knowing it would eventually show up in their changelog:
```
## v2.27.0 (2022-05-16)
### Feat
- added another thing
## v2.27.0-alpha (2022-05-16)
### Feat
-  let it respect pre-commit reformats when bumping
## v2.26.0 (2022-05-14)
### Feat
- Add --allow-abort option
````

When they really wanted:
 ```
[...]

PR #537, opened 2022-07-19
adds a changelog tag parser to filter tags (--tag-parser; tag_parser)

bhelgs:
## Description
Some repositories have tags that are not matching the `bump` version pattern (e.g. `alpha`/`beta` release) #519.
It makes sense to be able to filter those out. See this "Expected Behavior" section for an example.

The first commit:
1. feat: adds the filtering.

Subsequent commits (probably best reviewing each commit individually):
1. test: adds a test for when a "feat:" commit is tagged. Previously only "bump:" commits were tagged.
1. refactor: the linter was complaining. Mostly a copy/paste to reduce function complexity.
1. refactor: further refactoring to reduce the number of `if` (by avoiding an optional).

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes

## Expected behavior
originally:
```
## v1.1.0(2022-02-13)
### Feat
- add more

## v1.1.0-beta (2022-02-13)
### Feat
- add new

## v1.0.0 (2022-02-13)
### Feat
- initial
```

once filtered with `cz changelog --dry-run --tag-parser 'v[0-9]*\.[0-9]*\.[0-9]*'` will be:
```
## v1.1.0 (2022-02-13)
### Feat
- add more
- add new

## v1.0.0 (2022-02-13)
### Feat
- initial
```
Similar can be done with the `toml` option: `tag_parser: "v[0-9]*\\.[0-9]*\\.[0-9]*"`. Note `toml` required an extra `\`.

## Steps to Test This Pull Request
1. `scripts/test`

In the case below there is a filterable tag as the first commit. 
```
## v1.2.0-beta (2022-02-13)
### Feat
- another thing

## v1.1.0 (2022-02-13)
### Feat
- add more
- add new

## v1.0.0 (2022-02-13)
### Feat
- initial
```

Question:
 * Should the `v1.2.0-beta` tag:
    * stay as `## v1.2.0-beta`?
    * be replaced with `## Unreleased`? 
    * Maybe `## Prerelease tag: v1.2.0-beta`?
    * be customizable?
 * No test in place yet. 

## Additional context

#519

PR #692, opened 2023-03-23
feat: Introduce tag_regex option with smart default 

robertschweizer:
## Description

Closes https://github.com/commitizen-tools/commitizen/issues/519

CLI flag name: --tag-regex

Heavily inspired by https://github.com/commitizen-tools/commitizen/pull/537, but extends it with a smart default value to exclude non-release tags. This was suggested in https://github.com/commitizen-tools/commitizen/issues/519#issuecomment-1163923719

I separated commits to make review easier.

## Checklist


robertschweizer:
It looks like these changes need some reworking because they partially duplicate what was done in https://github.com/commitizen-tools/commitizen/commit/3a1af6a1b3299841d450a4f07b89fb918abc5cad.

Would you merge this if I rework it? It didn't get a review before.

Lee-W:
Hi @robertschweizer sorry for the late reply. I briefly walk through it. I like the idea and will definitely take a look after reworked 🙂

robertschweizer:
Sorry, I don't have much time at the moment. I started rebasing the first commit and opened https://github.com/commitizen-tools/commitizen/pull/775 to get that merged.

To continue here (not sure if I will find the time myself), it's important to consider:
* My use-case is to ignore tags in the changelog that we don't care about when running `cz bump --changelog`. I probably don't need a `tag_regex`.
* `cz bump --changelog` sets the `--incremental` flag. This finds the previous version with a different logic than `cz bump`. The first line matching `version_schema`'s `parser` in the existing changelog is used as previous version.
  This version is then formatted with `tag_format` and searched for in existing tags with a similarity threshold of 0.89.
* Suspected bugs in `cz changelog`: 
  * Existing tags (`get_tags()` function) and `--rev-range` values are validated by constructing the configured `version_scheme` on them. 
    This means it matches `packaging.Version._regex` instead of `BaseVersion.parser`. 
    Both `SemVer` and `Pep440` version providers thus use `packaging`'s PEP440 version matcher here.
    * The two standards require different pattern matching
  * `--incremental` uses `version_schema` but ignores `tag_format` to search changelog. 
    * Probably `cz bump --changelog` should pass the current (previous) release version explicitly, to make searching of the existing changelog unnecessary.
      * This would probably fix my use-case already.
    * It's probably rather restrictive to require the changelog to contain the used Git tag names.
* `ScmProvider` should really use some smarter default regex matching when a non-standard `tag_format` is set. Maybe using placeholders other than `$version` in `tag_format` should be deprecated to make this easier. The `$version` string can now be configured using `version_scheme`.

@Lee-W @noirbizarre

woile:
Looks good, hopefully we can merge it soon

commitizen/commands/changelog.py
```diff
@@ -55,8 +58,8 @@ def __init__(self, config: BaseConfig, args):
             or defaults.change_type_order
         )
         self.rev_range = args.get("rev_range")
-        self.tag_format = args.get("tag_format") or self.config.settings.get(
-            "tag_format"
+        self.tag_format: str = args.get("tag_format") or self.config.settings.get(
+            "tag_format", DEFAULT_SETTINGS["tag_format"]
```
noirbizarre, line 62: You don't need to inject default there, default are already loaded and overwritten by actual settings

docs/config.md
```diff
@@ -38,10 +38,18 @@ Version provider used to read and write version [Read more](#version-providers)
...
-Default: `None`
+Default: `$version`
 
 Format for the git tag, useful for old projects, that use a convention like `"v1.2.1"`. [Read more][tag_format]
 
+### `tag_regex`
+
+Type: `str`
+
+Default: Based on `tag_format`
```
noirbizarre, line 49:
```suggestion
Default: Computed from `tag_format`
```
or
```suggestion
Default: Extrapolated from `tag_format`
```

poetry.toml
noirbizarre, line 1: Not related to this PR, can you remove this ?

commitizen/git.py
```diff
@@ -140,7 +141,7 @@ def get_filenames_in_commit(git_reference: str = ""):
         raise GitCommandError(c.err)
 
 
-def get_tags(dateformat: str = "%Y-%m-%d") -> List[GitTag]:
+def get_tags(dateformat: str = "%Y-%m-%d", *, pattern: re.Pattern) -> List[GitTag]:
```
noirbizarre, line 144: Make the pattern filtering optional:
```suggestion
def get_tags(dateformat: str = "%Y-%m-%d", *, pattern: re.Pattern | None = None) -> List[GitTag]:
```

tests/test_tags.py
```diff
@@ -0,0 +1,76 @@
+import re
+from typing import Dict
+
+import pytest
+from packaging.version import Version
+
+from commitizen.tags import make_tag_pattern, tag_from_version
+
+TAG_FORMATS: Dict[str, Dict[str, list]] = {
```
noirbizarre, line 9: As a failsafe, add a case for the minimal `$version` that should always work:
```suggestion
TAG_FORMATS: Dict[str, Dict[str, list]] = {
	"$version": {
		"tags_per_version": [
			("1.2.3", "1.2.3"),
            ("1.2.3a2", "1.2.3a2"),
            ("1.2.3b2", "1.2.3b2"),
            ("1.2.3+1.0.0", "1.2.3+1.0.0"),
		],
		"invalid_tags": ["v1.2.3", "unknown-tag", "1-2-3", "ver1.0.0"]
	},
```

noirbizarre (commented):
Seems to go the right way.
I added a few suggestions.

AlexeySanko:
Looks as great feature for monorepo. Any update on it?

noirbizarre:
I am wondering if https://github.com/commitizen-tools/commitizen/pull/1297 is not providing somehow the same feature, but I might be wrong.

Lee-W:
As this has no update for a few months and there are lots of conflicts, we're not able to work on it on our own, I'm going to close this one. Feel free to reopen it if we still want this improvement in 🙂

Issue #687, opened 2023-03-13
Current release changelog with merged pre-release changelog


Apkawa:
### Description

in my case, in gh-actions manually called `cz ch` like this, as example:

```
cz changelog --dry-run 0.2.0a2 > CHANGELOG.txt
```
and use as body for gh-release .

Example:
```
$ git tag
v0.1.0  
v0.2.0-a0
v0.2.0-a1
v0.2.0-a2
```

for the prerelease case, I use the current tag, like "0.2.0a2", that's fine.

but for release after all pre-releases i need to collect all commits between releases message like `0.2.0a0..0.2.0`

at the moment it looks like this

```shell
$ cz ch 0.2.0a0..0.2.0 --dry-run
## v0.2.0 (2023-03-13)

### Fix

- foo fix 3

## v0.2.0b0 (2023-03-13)

## v0.2.0a1 (2023-03-13)

### Fix

- some fix

## v0.2.0a0 (2023-03-13)

### Feat

- super feature

## v0.1.0 (2023-03-13)
```

### Possible Solution

I'm thinking about adding a flag for the merged prerelease commit in one release.
For example, `--merge-prerelease`

I want something like this:
```
cz ch 0.2.0a0..0.2.0 --dry-run  --merge-prerelease
## v0.2.0 (2023-03-13)

### Fix

- foo fix 3
- some fix

### Feat

- super feature
```

In the case of a multi-release range `0.1.0..0.3.0`, the output would be:
```
$ cz ch 0.1.0..0.3.0 --dry-run  --merge-prerelease

## v0.3.0 (2023-03-13)

### Fix

- foo fix 3
- some fix

### Feat

- super feature

## v0.2.0 (2023-03-13)

### Feat

- feature

## v0.1.0 (2023-03-13)

### Feat

- feature foo
```

### Additional context

This feature useful for gh-actions for me

```
      - name: Check version
        uses: apkawa/gh-action-semver-check@master
        id: version
      - name: Generate Changelog
        run: |
          cz changelog --dry-run ${{ steps.version.outputs.full_with_prefix }}a0..${{ steps.version.outputs.full_with_prefix }}  > ${{ github.workspace }}-CHANGELOG.txt

      - name: Publish archives and packages
        uses: softprops/action-gh-release@v1
        if: ${{ !github.event.schedule && startsWith(github.ref, 'refs/tags') }}
        with:
          draft: true
          prerelease: ${{ needs.version.outputs.prerelease }}
[...]

PR #703, opened 2023-04-13
feat(changelog): add merge-prerelease flag

### LINKED ISSUE #685: `codecov` dev dependency is deprecated
### Description

`codecov` has been deprecated a while ago and should be removed or replaced.

### Possible Solution

Follow [migration guide](https://docs.codecov.com/docs/deprecated-uploader-migration-guide#python-uploader)

### Additional context

https://github.com/codecov/codecov-python

### Additional context

_No response_

MoritzBoehme:
## Description

Added a `--merge-prerelease` flag and config option as described in #687.

## Checklist

## Additional context

The dependency `codecov` seems to be no longer available. ~~I can try to fix it in this pull request, but could also create a new one, depending on what is preferred.~~
I have now to my own frustration fixed the `codecov` action in the second commit. This also closes #685


commitizen/cli.py
```diff
@@ -247,10 +247,19 @@
                         "name": "--start-rev",
                         "default": None,
                         "help": (
-                            "start rev of the changelog."
+                            "start rev of the changelog. "
```
MoritzBoehme, line 250: Hope it's fine to sneak this small fix in here. Just something I noticed.

MoritzBoehme:
It seems the pre-commit hook `prettier` is failing for windows only...
No idea, why this would be the case :/

MoritzBoehme:
@woile could you perhaps have a look at the pr? Do you maybe have some idea on how to keep `prettier` from failing?
Thanks for this awesome project btw =)

tests/test_changelog.py
```diff
@@ -800,6 +1080,16 @@ def test_generate_tree_from_commits(gitcommits, tags):
...
 
 
+def test_generate_tree_from_commits_merge_prerelease(gitcommits, tags):
+    parser = defaults.commit_parser
+    changelog_pattern = defaults.bump_pattern
+    tree = changelog.generate_tree_from_commits(
+        gitcommits, tags, parser, changelog_pattern, merge_prerelease=True
+    )
+
+    assert tuple(tree) == COMMITS_TREE_MERGED_PRERELEASES
```
woile, line 1090: Should `COMMITS_TREE_MERGED_PRERELEASES` be renamed to `COMMITS_TREE_AFTER_MERGED_PRERELEASES`? 

How is this test different than just `test_generate_tree_from_commits`?

woile:
How does this feature affect existing changelogs?
Will it update old pre-releases as well? could we add a test for it?

woile:
Can you point this PR to `v3`? thanks!

tests/test_changelog.py
MoritzBoehme: Yeah, that name is better. It's different in the sense, that `COMMITS_TREE_AFTER_MERGED_PRERELEASES` does not contain pre-releases (there were 2) and the changes of them are now part of the next non-prerelease.

MoritzBoehme:
> How does this feature affect existing changelogs? Will it update old pre-releases as well? could we add a test for it?

Existing changelogs will be changed when run with this flag. This is unless you also add the `--incremental`, set the `changelog_incremental` config option to `true` or use `cz bump --changelog` (as this implies incremental). I am adding a test right now to ensure this leaves prior pre-releases untouched this way.

> Can you point this PR to v3? thanks!

Will do!

PR #712, opened 2023-04-19
Release/v3

woile:
## Description

Preparing commitizen for v3.


commitizen/defaults.py
```diff
@@ -50,8 +50,8 @@ class Settings(TypedDict, total=False):
 
 name: str = "cz_conventional_commits"
 config_files: List[str] = [
-    "pyproject.toml",
     ".cz.toml",
+    "pyproject.toml",
```
Lee-W, line 54: Wondering why we change the loading order

commitizen/commands/init.py
```diff
@@ -13,59 +13,150 @@
...
+
+    def tags(self) -> Optional[List]:
+        """Not a property, only use if necessary"""
+        if self.latest_tag is None:
+            return None
+        return get_tag_names()
+
+    @property
+    def is_pre_commit_installed(self) -> bool:
+        return shutil.which("pre-commit") is not None
```
Lee-W, line 69: nitpick: I might prefer `not shutil.which("pre-commit")` a bit more

commitizen/commands/init.py
```diff
@@ -13,59 +13,150 @@
...
             return
 
-        # No config for commitizen exist
-        config_path = self._ask_config_path()
+        out.info("Welcome to commitizen!\n")
+        out.line(
+            "The prompts will ask you different questions " "to configure your project."
+        )
+        out.line("For further configuration visit:")
+        out.line("\n\thttps://commitizen-tools.github.io/commitizen/config/\n")
```
Lee-W, line 88: If we're using `out.line`, do we still need the `\n` here?

commitizen/commands/init.py
```diff
@@ -13,59 +13,150 @@
...
         hook_types = questionary.checkbox(
             "What types of pre-commit hook you want to install? (Leave blank if you don't want to install)",
             choices=[
-                questionary.Choice("commit-msg", checked=True),
-                questionary.Choice("pre-push", checked=True),
+                questionary.Choice("commit-msg", checked=False),
+                questionary.Choice("pre-push", checked=False),
             ],
-        ).ask()
+        ).unsafe_ask()
```
Lee-W, line 134: May I know why do we choose to use `unsafe_ask` here?

commitizen/commands/init.py
```diff
@@ -13,59 +13,150 @@
...
+        if major_version_zero:
+            values_to_add["major_version_zero"] = major_version_zero
+
+        # Collect hook data
         hook_types = questionary.checkbox(
             "What types of pre-commit hook you want to install? (Leave blank if you don't want to install)",
             choices=[
-                questionary.Choice("commit-msg", checked=True),
-                questionary.Choice("pre-push", checked=True),
+                questionary.Choice("commit-msg", checked=False),
```
Lee-W, line 131: May I know why do we want to set both of them to `false` instead of `true`

commitizen/commands/init.py
```diff
@@ -108,21 +199,90 @@ def _ask_tag_format(self, latest_tag) -> str:
...
                 tag_format = "$version"
         return tag_format
 
-    def _search_pre_commit(self) -> bool:
-        """Check whether pre-commit is installed"""
-        return shutil.which("pre-commit") is not None
+    def _ask_version_provider(self) -> str:
+        """Ask for setting: version_provider"""
+
+        OPTS = {
```
Lee-W, line 217: Should we make it a variable in `version_provider.py`?

pyproject.toml
```diff
@@ -71,28 +50,27 @@ charset-normalizer = ">=2.1.0,<3.1"
 # Use the Python 3.11 compatible API: https://github.com/python/importlib_metadata#compatibility
 importlib_metadata = { version = ">=4.13,<5"}
 
-[tool.poetry.dev-dependencies]
-ipython = "^7.2"
+[tool.poetry.group.dev.dependencies]
```
Lee-W, line 53: If we're using this feature, should we set the minimum poetry version?

pyproject.toml
```diff
@@ -7,27 +7,6 @@ version_files = [
   ".pre-commit-config.yaml:rev:.+Commitizen"
 ]
 
-[tool.black]
```
Lee-W, line 10: Please correct me if I'm wrong, we're removing these lines because they're default settings right?

pyproject.toml
```diff
@@ -71,28 +50,27 @@ charset-normalizer = ">=2.1.0,<3.1"
...
 pytest-mock = "^3.10"
 pytest-regressions = "^2.4.0"
 pytest-freezer = "^0.4.6"
+pytest-xdist = "^3.1.0"
 # code formatter
 black = "^22.10"
-isort = "^5.10.0"
 # linter
-flake8 = "^3.6"
+ruff = "^0.0.262"
```
Lee-W, line 65: super 🙌

scripts/format
```diff
@@ -5,5 +5,6 @@ export PREFIX="poetry run python -m "
 
 set -x
 
-${PREFIX}isort commitizen tests
+# ${PREFIX}isort commitizen tests
```
Lee-W, line 8: Should we just remove it instead of commenting it out?

pyproject.toml
```diff
@@ -146,3 +121,30 @@ known_first_party = ["commitizen", "tests"]
...
+
+[tool.mypy]
+files = "commitizen"
+ignore_missing_imports = true
+disallow_untyped_decorators = true
+disallow_subclassing_any = true
+warn_return_any = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_unused_configs = true
```
Lee-W, line 150: I'm not sure but won't it be too strict for our codebase now? but I'd be really glad if commitizen can pass these checks 🙌

commitizen/defaults.py
woile: You are right, it's not needed anymore

commitizen/commands/init.py
woile: No problemo

commitizen/commands/init.py
woile: Yes, for formatting, we want the link to be surrounded by new lines. I think it makes it more visible and standing.

commitizen/commands/init.py
woile: `unsafe_ask` doesn't handled `KeyboardInterrupt`. What happens if you use `ask`? Well, if you are running `cz init` and you type `ctrl+c` it won't terminate the program, instead it will terminate the current question and go to the next one. Making the user call `ctrl+c` many times, and that won't even terminate the program and you'll end up with the `.cz.toml` file anyways.

With unsafe, we handle the keyboard interrupt instead.

https://github.com/commitizen-tools/commitizen/pull/712/files#diff-43bdf96da59e2edd79c3797ccaf86dda6495266c02768cb045502851ebb047e8R101

commitizen/commands/init.py
woile: I think `.pre-commit` with `cz check` is overkill for many projects, I actually don't like enforcing conventional commits because it allows skipping those commits which do not match. Having it set to true feels "enforced" by commitizen. I'd like to run `cz init` press enter many times and only have `commitizen` configured fast. If a set of users make use of `.pre-commit`, then they have to actively set it to true.

pyproject.toml
woile: I didn't know it was possible, I'll give it a try

commitizen/commands/init.py
woile: I thought about it. But the thing is that the values are actually defined in pyproject.toml as part of `[tool.poetry.plugins."commitizen.provider"]`. Setting them in `version_provider` may create confusion, as they are only used by init. I think in the future we can try to read them from the plugins system (which I don't know yet how to do)

pyproject.toml
woile: Yes, none of the lines removed affected anything. In many other projects I haven't configured anything for black and it just works. I thought it's nice to do some clean up, and only add if we have to

scripts/format
woile: yep

pyproject.toml
woile: I think this is how it was configured (and they are passing). But mypy settings where in the `setup.cfg` before.

pyproject.toml
woile: I found this: https://github.com/python-poetry/poetry/issues/3316

commitizen/commands/init.py
Lee-W: Got it. If that's the case, what do you think of the following lines? It's much easier for me to image the output of it.

```python
        out.line(
            "The prompts will ask you different questions to configure your project.\n"
            "For further configuration visit:\n"
            "\n"
            "https://commitizen-tools.github.io/commitizen/config/"
        )
```

commitizen/commands/init.py
Lee-W: Wow, that's awesome. 🤩 Thanks for explaining! I should have try it out myself 🤦‍♂️ Out of time yesterday 😢

commitizen/commands/init.py
Lee-W: Got it. I think I'm the so-called "a set of users" haha 😆 But ya, I'm good with it as well 🙂

pyproject.toml
Lee-W: ah. I thought we could do that through build section. Then maybe we can add a description in the `README.md`?

commitizen/commands/init.py
```diff
@@ -13,59 +13,150 @@
...
             return
 
-        # No config for commitizen exist
-        config_path = self._ask_config_path()
+        out.info("Welcome to commitizen!\n")
+        out.line(
+            "The prompts will ask you different questions " "to configure your project."
+        )
+        out.line("For further configuration visit:")
+        out.line("\n\thttps://commitizen-tools.github.io/commitizen/config/\n")
```
Lee-W, line 88: out.line(
            "The prompts will ask you different questions to configure your project.\n"
            "For further configuration visit:\n"
            "\n"
            "https://commitizen-tools.github.io/commitizen/config/"
        )

docs/config.md
```diff
@@ -2,30 +2,185 @@
...
+
+Disallow empty commit messages, useful in ci. [Read more][allow_abort]
+
+### `changelog_file`
+
+Type: `str`
+
+Default: `CHANGELOG.md`
+
+filename of exported changelog
```
Lee-W, line 91:
```suggestion
Filename of exported changelog
```

maybe we should unify the capitalization?

docs/config.md
```diff
@@ -2,30 +2,185 @@
...
+
+### `changelog_incremental`
+
+Type: `bool`
+
+Default: `false`
+
+Update changelog with the missing versions. This is good if you don't want to replace previous versions in the file. Note: when doing `cz bump --changelog` this is automatically set to `true`
```
Lee-W, line 99: Will it be a bit confusing as the default behavior for `cz bump --changelog` and `cz changelog` are different?

docs/init.md
```diff
@@ -1,15 +1,18 @@
...
 
-We hope to detect the files where the version is also repeated eventually,
-like `package.json` or `__version__`.
+1. Choose a convention rules (`name`)
+1. Choosing a version provider (`commitizen` or for example `Cargo.toml`)
+1. Detecting your project's version
+1. Detecting the tag format used
+1. Choosing a version type (`semver` or `pep440`)
+1. Whether to create the changelog automatically or not during bump
+1. Whether you want to keep the major as zero while building alpha software.
```
Lee-W, line 18:
```suggestion
2. Choosing a version provider (`commitizen` or for example `Cargo.toml`)
3. Detecting your project's version
4. Detecting the tag format used
5. Choosing a version type (`semver` or `pep440`)
6. Whether to create the changelog automatically or not during bump
7. Whether you want to keep the major as zero while building alpha software.
```

nitpick: even though I know markdown will handle that for us, I still think it would be more readable haha

Lee-W (approved):
Awesome! I left a few minor comments and suggests, but I'm good even if they're not addressed.

docs/config.md
woile: This has always been the case. `cz bump --changelog` avoid messing with your changelog (this allows users to modify further the file). With `cz changelog` the whole file will be recreated. An alternative would be to set `changelog_incremental` as default = True. This way `cz bump --changelog` and `cz changelog` would behave the same. But it's another breaking change

Lee-W: I like `changelog_incremental=True` as default a bit more, but would be ok if we keep it as it is

PR #696, opened 2023-04-05
feat: draft of the --empty parameter

12rambau:
## Description

First draft of a new feature. The idea is to grant the user the capability to run `cz bump` without prior commits. The reasons are described in the associated issue. 

here we add an `--empty` parameter to the bump method
when this parameter is set and if no increment is find, a automatic "PATCH" is released. 

I'll wait for your feedback before working on tests and documentation.

## Checklist

- [ ] Add test cases to all the changes you introduce
- [ ] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [ ] Test the changes on the local machine manually
- [ ] Update the documentation for the changes

## Additional context
Fix #662

PR #723, opened 2023-04-24
feat: add --allow-no-commit option

### LINKED ISSUE #662: create a dev bump straight after a normal bump
### Description

In projects I'm working in (the ones that are not yet using `commitizen`) maintainers are bumping a dev version straight after the bump of a new one. That ensures that the version number of the current master/main branch is not the same as the latest release. 

I wanted to use the new `--devrelease` option of the `bump` command for that purpose but as there are no changes to report since the last version, it's not working and the bump fail.

Do you think it would be doable or is it a workflow that you don't want to cover? 

### Steps to reproduce

```
cz bump
cz bump --devrelease 0
```

### Current behavior

bump fails

### Desired behavior

bump succeed (bonus point if nothing is written to the changelog file)

### Screenshots

_No response_

### Environment

cz version: master (as I still have an issue with my name convention and didn't find the time to work on #463)

12rambau:
superceed #696 

## Description

First draft of a new feature. The idea is to grant the user the capability to run `cz bump` without prior commits. The reasons are described in the associated issue. 

here we add an `--empty` parameter to the bump method
when this parameter is set and if no increment is find, a automatic "PATCH" is released. 

I'll wait for your feedback before working on tests and documentation.

## Checklist

## Additional context
Fix #662

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -371,6 +371,12 @@ data = {
                         "help": "Determine the next version and write to stdout",
                         "default": False,
                     },
+                    {
+                        "name": ["--empty"],
+                        "default": False,
+                        "help": "bump tags without new commits",
+                        "action": "store_true",
+                    },
                 ],
             },
             {
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -372,9 +372,9 @@ data = {
                         "default": False,
                     },
                     {
-                        "name": ["--empty"],
+                        "name": ["--allow-no-commit"],
                         "default": False,
-                        "help": "bump tags without new commits",
+                        "help": "bump version without eligible commits",
                         "action": "store_true",
                     },
                 ],
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -160,6 +160,7 @@ class Bump:
         build_metadata = self.arguments["build_metadata"]
         increment_mode: str = self.arguments["increment_mode"]
         get_next: bool = self.arguments["get_next"]
+        is_empty: bool | None = self.arguments["empty"]
 
         if manual_version:
             if increment:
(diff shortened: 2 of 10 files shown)


woile:
LGTM 💪🏻 , needs tests though

12rambau:
🍾 Nice to see that you approve the idea. 

> needs tests though

absolutely, I'll work on them this week (I hope)

Lee-W (commented):
LGTM. I recently happen to work on a project need such feature lol 🎉

Lee-W:
@woile @noirbizarre, I picked this up and refactored it a bit. I changed the argument to `--allow-no-commit`. One weird thing that might happen is that one commit might have multiple versions. I'm not sure whether it would cause any issue

12rambau:
thanks @Lee-W

Lee-W:
Hey @woile @noirbizarre , I just rebased it from the latest master branch. Would be nice if you could take a look. Thanks!

Lee-W:
> Should this be exposed as a config parameter as well? besides that it looks good

Yep, could be. let me update it as well. Just notice the CI also failed 🤦‍♂️

Lee-W:
> Should this be exposed as a config parameter as well? besides that it looks good

@woile After a second thought, I think we probably shouldn't add it. I feel this is somewhat like an exception, and if we're to do it, we probably need to introduce a better config/arugment handling to avoid https://github.com/commitizen-tools/commitizen/pull/996

woile:
And there's still an issue with `vsc`  with no commits

Lee-W:
> And there's still an issue with `vsc`  with no commits

I'm afraid I don't quite get this one 🤔

noirbizarre:
I'll take a look tomorrow 👍🏽

noirbizarre (approved):
LGTM ! (before it reaches 2 years old, sorry for that)
This feature is needed. I think we can iterate on smaller pull request to add later settings or any other adjustments

Issue #698, opened 2023-04-05
user does not need to run format and test manually anymore


12rambau:
### Type

- [X] Content inaccurate

### URL

https://commitizen-tools.github.io/commitizen/contributing/

### Description

I just edited the pre-commit hook in a PR and I realized the test and format script are run internally. would it make sense to remove these steps from both the contributing documentation and the PR checklist ? 

On the other hand I faced many dificulties running them as my nevironment was missing a lot of the test/formatting dependencies. wouldn't it be better if these tests were run using isolated environments using either [nox](https://nox.thea.codes/en/stable/) or [tox](https://tox.wiki/en/latest/) ? 

happy to make a PR

woile:
I personally prefer to use the scripts and let the CI handle different python versions.
What kind of difficulties did you have with the test/formatting? If you do `poetry install` you should have all the deps, including the development deps. There's a chance the problem is actually `poetry`, which is deprecating the section `tool.poetry.dev-dependencies` and instead, we have to create a new group `tool.poetry.groups.dev` for example

12rambau:
nox is not forcing you to test multiple version of python. If you don't sepcify anything in the session parameter, it will use the python installed on your computer (in my case 3.10).

Note that it is not recommended to run the tests in pre-commits [(quoting the author of pre-commits)](https://github.com/pre-commit/pre-commit-hooks/issues/291#issuecomment-394167917):

> - tests are often slow, committing should be fast  
> - It would have to be a system / script hook since you'd need repository specific dependencies installed (as you've noticed)

The format script could be drop all together by using richer pre-commit hooks (namely the holy trinity: ruff, black, prettier)

bearomorphism:
our pre-commit hooks does not contain tests anymore. closing this issue as completed.

PR #724, opened 2023-04-24
refactor: move ruff and black from the format file to the pre-commit hooks

12rambau:
## Description
I ask myself, why using a format script to lint the lib instead of a pre-commit hook ? Looking at the format file, it's only using ruff and black so as they have very nicely design hooks I used them instead.

## Checklist

If you agree with this modification, I'll update the PR/documentation instructions

related to #698

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,18 @@ repos:
           - post-commit
           - push
 
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.262"
+    hooks:
+      - id: ruff
+        stages: [commit]
+
+  - repo: "https://github.com/psf/black"
+    rev: "23.3.0"
+    hooks:
+      - id: black
+        stages: [commit]
+
   - repo: local
     hooks:
       - id: format
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,12 +42,6 @@ repos:
 
   - repo: local
     hooks:
-      - id: format
-        name: format
-        language: system
-        pass_filenames: false
-        entry: ./scripts/format
-        types: [python]
 
       - id: linter and test
         name: linter and test
--- a/commitizen/bump.py
+++ b/commitizen/bump.py
@@ -23,7 +23,6 @@ else:
 def find_increment(
     commits: List[GitCommit], regex: str, increments_map: Union[dict, OrderedDict]
 ) -> Optional[str]:
-
     if isinstance(increments_map, dict):
         increments_map = OrderedDict(increments_map)
 
@@ -103,7 +102,6 @@ def semver_generator(current_version: str, increment: str = None) -> str:
     # so it doesn't matter the increment.
     # Example: 1.0.0a0 with PATCH/MINOR -> 1.0.0
     if not version.is_prerelease:
-
         if increment == MAJOR:
             increments_version[MAJOR] += 1
             increments_version[MINOR] = 0
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -74,7 +74,6 @@ class GitTag(GitObject):
 
     @classmethod
     def from_line(cls, line: str, inner_delimiter: str) -> "GitTag":
-
         name, objectname, date, obj = line.split(inner_delimiter)
         if not obj:
             obj = objectname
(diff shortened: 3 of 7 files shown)


woile:
I'm a heavy user of the `format` script, so please leave it. And also leave them as dev dependencies on the `pyproject.toml`, they are used by the CI and the IDE's

12rambau:
thanks @woile for the feedback, let me advocate a bit for this proposed change before I remove them: 

> I'm a heavy user of the format script, so please leave it.

if they are part of the pre-commits what is preventing you from executing them all at once in 5s with `pre-commit run -a` while developping ? This command is widelly known so no need for specific explaination to contributors. Just say "run the pre-commits".

> also leave them as dev dependencies on the pyproject.toml, they are used by the CI and the IDE's.

If they are used in actions, why not using the pre-commit actions https://github.com/12rambau/sepal_ui/blob/cbca0af9a09e347261af068a40be06ca93b62c5c/.github/workflows/unit.yml#L15

The advantage that I see are: 
- only 1 version for everything (local, CI, hooks)
- no need for linting deps in the lib
- can be used in any CI by running `pre-commit run -a` or the nicer github action

the issue for me with the current formatting: 
- multiplication of piloting files: format, pre-commit-hook, github actions 
- duplication of dependencies e.g. black/ruff is tested in both format and pre-commit hooks which use 2 different environments
- On my end I deactivated pre-commits because of installation issues for format and test which never occurs to me with github based hooks. (I assume it was due to my poetry version)

If you are really against it, there is no point to my PR as the main objective is to use one and only config for linting/format operation and make it centralized in the pre-commits-config file. Let me know and feel free to close it (along with the associated issue).

noirbizarre:
To be honest, I hesitated to submit the same kind of change, use `ruff` and `black` pre-commit plugins like this PR, but also:
- use `commitizen` pre-commit plugin (dogfooding, if it doesn't work for `commitizen` itself, this is an issue)
- use `mypy` pre-commit plugin
- target a single Python version for linting (targeting all version with MyPy result in a false positive and some ugly work-around to support both 3.7 and 3.11)
- remove the tests from `pre-commit` (time lost when you already spent some time on your PR running tests again and again)
- remove pre-commit on push (redundant because you push commits that already pass the test)
- remove the scripts/*
- use a script runner based on `pyproject.toml` for dev scripts (like https://poethepoet.natn.io/)
- add a `tox.ini` to test against all supported versions

Note that in the current state, sometimes you need to perform a `poetry install` before committing or pushing because entrypoints needs to be up to date.

In fact I already have a branch with some of those changes (pre-commit changes) and I have a local `tox.ini` file for testing.
Adding Poe The Poet wouldn't cost much to add to the PR.

So I am really interested in this discussion and I would love to contribute my changes too.

12rambau:
@noirbizarre I agree with your analysis but packaging modification have lots of impact on existing branches and forks so in my experience it's better to work with baby steps like this one instead of 1 single PR that will lead to huge conflicts for everyone else. If you look at the associated issue, it's more or less in the pipe. Maybe we can discuss there with the maintainers and keep things simple in this PR ?

Lee-W:
Hi @12rambau @noirbizarre , sorry for the late reply. I just read through this discussion. I like @noirbizarre 's detailed idea and agree with @12rambau that we should take baby steps. Here are some of my thoughts regarding each point.


> target a single Python version for linting (targeting all version with MyPy result in a false positive and some ugly work-around to support both 3.7 and 3.11)

Python 3.7 is about to reach its EOL and I think it would still be better if we could lint on each version. It still finds some unexpected errors earlier. This strategy is also applied to many large projects as well.

> remove the tests from pre-commit (time lost when you already spent some time on your PR running tests again and again)

I'm not sure. I still wish everyone actually ran the test once before they pushed. I'm ok with moving it to push only

> remove pre-commit on push (redundant because you push commits that already pass the test)

The reason of using pre-commit on push is to run the long-running tasks that don't need to be run everytime.

> use a script runner based on pyproject.toml for dev scripts (like https://poethepoet.natn.io/)

This is awesome 🤩

> add a tox.ini to test against all supported versions

Just a random idea, should we try nox? But tox is good for me as well

12rambau:
>Just a random idea, should we try nox? But tox is good for me as well

I am a nox person so I would prefer to use it instead of tox but that 100% personnal though functionalities are more or less the same. 

As mentioned earlier in the PR, I think baby steps will be safer so maybe @noirbizarre comments could be moved to an issue and we could iterate in other PRs.

12rambau:
I now face issues in most of the files (nothing severe though), anything I should do to move this PR forward ?

Lee-W:
I think we'll need @woile 's feedback on this one

PR #725, opened 2023-04-24
chore: add chore to types list

tal-lederman-dy:
## Description
add chore to types list
Following description here:
https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits

## Checklist

- [x] Add test cases to all the changes you introduce - nothing to test

## Expected behavior
another option is added to the list of types

just run `cz commit`

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -107,6 +107,14 @@ class ConventionalCommitsCz(BaseCommitizen):
                         ),
                         "key": "c",
                     },
+                    {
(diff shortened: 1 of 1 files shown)


Lee-W:
Hi @tal-lederman-dy , thanks for opening this issue, but we've decided not to add it. Please check [Why are revert and chore valid types in the check pattern of cz conventional_commits but not types we can select?¶](https://commitizen-tools.github.io/commitizen/faq/#why-are-revert-and-chore-valid-types-in-the-check-pattern-of-cz-conventional_commits-but-not-types-we-can-select) Thanks!

PR #726, opened 2023-04-25
Fix/major version zero bump map

woile:
## Description

It's currently not possible to modify major version zero, meaning that it only works with conventional commits.

If you have a custom rule, and you set `major_version_zero` to true, it won't find any commits, because it will be actually using the one in `defaults.bump_major_version_zero` which matches with conventional commit conventions

## Checklist

## Expected behavior

Users should be able to express the behavior when major_version_zero is true, and they have a different set of rules.

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -30,6 +30,7 @@ def parse_subject(text):
 class ConventionalCommitsCz(BaseCommitizen):
     bump_pattern = defaults.bump_pattern
     bump_map = defaults.bump_map
+    bump_map_major_version_zero = defaults.bump_map_major_version_zero
     commit_parser = defaults.commit_parser
     version_parser = defaults.version_parser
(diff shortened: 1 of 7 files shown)


commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -30,6 +30,7 @@ def parse_subject(text):
 class ConventionalCommitsCz(BaseCommitizen):
     bump_pattern = defaults.bump_pattern
     bump_map = defaults.bump_map
+    bump_map_major_version_zero = defaults.bump_map_major_version_zero
```
Lee-W, line 33: If this is not actually a default for every cz, should we move it out of `defaults`? I've been thinking for moving these kind of stuff from defaults for a long time but not yet have a chance to do so 🤦‍♂️

Lee-W (approved):
Overall, I'm good with the PR. Just have one minor discussion. But even if we agree on that, we can open an issue and work on that later 🙂

Issue #249, opened 2020-08-14
cz commit with prepare-commit-msg hook


saygox:
## Description

*cz commit* is very good tool.
If prepare-commit-msg hook calls *cz commit*, become easy to commit by many user.

## Possible Solution

like *cz check*, writing pre-commit config

```yaml
---
repos:
- hooks:
  - id: commitizen-prepare-commit-msg
    stages: [prepare-commit-msg]
```

and user type *git comit*, then prepare-commit-msg hook works and invokes *cz commit*.

Is this feature accepted as this product policy?

## Additional context

I did a basic research. And I found that there are two issues.

- The current *cz commit* calls git. But  in this case, git calls *cz commit*
- Git hook is started without standard input, so we can't create commit message interactively

I made a test implementation to solve the problem by the following method

- When called from git, write the commit message to the file  which is set by "--commit-msg-file" argument
- Open tty manualy and use as stdio. 

but this change has sideffect.

tty is manually opened as io.FileIO.
But prompt_toolkit (which used by questionary) assumes tty is opened as io.TextIOWrapper.
As a workaround  I added wrap class 

```python
class WrapStdin:
    def __init__(self):
        fd = os.open("/dev/tty", os.O_RDWR | os.O_NOCTTY)
        tty = open(fd, "wb+", buffering=0)
        self.tty = tty

    def __getattr__(self, key):
        if key == "encoding":
            return "UTF-8"
        return getattr(self.tty, key)

    def __del__(self):
        self.tty.close()
```

All patch code is here.
https://github.com/saygox/commitizen/commit/033159d08e240c493b5fa36d67e870e3b443d4d6

Lee-W:
I love this new feature! I never thought we could use git hook like this. One problem I can think of is the compatibility for windows users. We might make some more tests before we can accept it. @Woile What do you think?

Woile:
Quite interesting indeed, please open a PR so we can discuss it there!

Lee-W:
Although we have a PR https://github.com/commitizen-tools/commitizen/pull/250, I'm not able to reproduce the result

mrlubos:
Hi @Lee-W, is there any update on this? Would love to automatically replace `git commit` with `cz commit` the same way I am able to in npm projects!

Lee-W:
[...]

PR #731, opened 2023-04-28
feat(commit): add --write-message-to-file option

### LINKED ISSUE #689: Is there any way to use this in a prepare-commit-msg hook?
### Description

The Node version of commitizen provides a `--hook` flag that allows users to generate a commit message, without actually performing a commit, is there a way to replicate this behavior in the Python version?

The example `prepare-commit-hook` in the cz-cli documentation is something like this:
```bash
#!/bin/bash
exec < /dev/tty && node_modules/.bin/cz --hook || true
```

### Possible Solution

A new command or flag such as `cz message` or `cz commit --message-only` that could be embedded into a git hook that would generate a commit message without performing a full commit.

### Additional context

Since the commit message in `cz --dry-run` is output to stdout (along with the questions) there's no way to just pipe the output of that directly into `.git/COMMIT_MSG`

### Additional context

_No response_

crai0:
## Description

This pull request resolves #689 by implementing my proposal for it.

## Checklist

## Expected behavior

See https://github.com/commitizen-tools/commitizen/issues/689#issuecomment-1526221321.

New option:
1. Run `cz commit --write-message-to-file message.txt` in a Git repository with staged changes and complete interactive dialog
1. The content of `message.txt` should be the generated commit message
1. The other behavior of `cz commit` is unchanged

Git hook:
1. Follow the tutorial in `docs/tutorials/auto_prepare_commit_message.md` in a Git repository with staged changes
1. When running `git commit` or `git commit -m "..."`, the usual `cz commit` interactive dialog runs
1. After completing the dialog (and exiting the GIT_EDITOR if it opened), a new commit with the generate commit message is created

## Additional context

These changes supersede #249 and #250.

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -63,9 +63,14 @@ data = {
                         "help": "show output to stdout, no commit, no modified files",
                     },
                     {
+                        "name": "--write-message-to-file",
+                        "metavar": "FILE_PATH",
+                        "help": "write message to file before commiting (can be combined with --dry-run)",
+                    },
+                    {
                         "name": ["-s", "--signoff"],
                         "action": "store_true",
-                        "help": "Sign off the commit",
+                        "help": "sign off the commit",
                     },
                 ],
             },
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -1,6 +1,7 @@
 import argparse
 import logging
 import sys
+from pathlib import Path
 from functools import partial
 from types import TracebackType
 from typing import List
@@ -64,6 +65,7 @@ data = {
                     },
                     {
                         "name": "--write-message-to-file",
+                        "type": Path,
                         "metavar": "FILE_PATH",
                         "help": "write message to file before commiting (can be combined with --dry-run)",
                     },
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -69,12 +69,8 @@ class Commit:
         if git.is_staging_clean() and not dry_run:
             raise NothingToCommitError("No files added to staging!")
 
-        if write_message_to_file is not None:
-            if not isinstance(write_message_to_file, str):
-                raise NotAllowed(
-                    "Commit message file name is broken.\n"
-                    "Check the flag `--write-message-to-file` in the terminal"
-                )
+        if write_message_to_file is not None and write_message_to_file.is_dir():
+            raise NotAllowed(f"{write_message_to_file} is a directory")
 
         retry: bool = self.arguments.get("retry")
 
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -14,6 +14,7 @@ from commitizen.exceptions import (
     NoAnswersError,
     NoCommitBackupError,
     NotAGitProjectError,
+    NotAllowed,
     NothingToCommitError,
 )
 from commitizen.git import smart_open
@@ -63,9 +64,17 @@ class Commit:
 
     def __call__(self):
         dry_run: bool = self.arguments.get("dry_run")
+        write_message_to_file = self.arguments.get("write_message_to_file")
 
         if git.is_staging_clean() and not dry_run:
             raise NothingToCommitError("No files added to staging!")
+
+        if write_message_to_file is not None:
+            if not isinstance(write_message_to_file, str):
+                raise NotAllowed(
+                    "Commit message file name is broken.\n"
+                    "Check the flag `--write-message-to-file` in the terminal"
+                )
 
         retry: bool = self.arguments.get("retry")
 
@@ -75,6 +84,10 @@ class Commit:
             m = self.prompt_commit_questions()
 
         out.info(f"\n{m}\n")
+
+        if write_message_to_file:
+            with smart_open(write_message_to_file, "w") as file:
+                file.write(m)
 
         if dry_run:
             raise DryRunExit()
--- a/docs/commit.md
+++ b/docs/commit.md
@@ -8,7 +8,8 @@ A commit can be signed off using `cz commit --signoff` or the shortcut `cz commi
 
 You can run `cz commit --write-message-to-file COMMIT_MSG_FILE` to additionally save the
 generated message to a file. This can be combined with the `--dry-run` flag to only
-write the message to a file and not modify files and create a commit.
+write the message to a file and not modify files and create a commit. A possible use
+case for this is to [automatically prepare a commit message](./tutorials/auto_prepare_commit_message.md).
 
 !!! note
     To maintain platform compatibility, the `commit` command disable ANSI escaping in its output.
--- a/docs/commit.md
+++ b/docs/commit.md
@@ -6,6 +6,10 @@ In your terminal run `cz commit` or the shortcut `cz c` to generate a guided git
 
 A commit can be signed off using `cz commit --signoff` or the shortcut `cz commit -s`.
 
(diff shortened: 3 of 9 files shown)


commitizen/cli.py
```diff
@@ -62,10 +62,15 @@
...
+                    {
+                        "name": "--write-message-to-file",
+                        "metavar": "FILE_PATH",
+                        "help": "write message to file before commiting (can be combined with --dry-run)",
+                    },
                     {
                         "name": ["-s", "--signoff"],
                         "action": "store_true",
-                        "help": "Sign off the commit",
+                        "help": "sign off the commit",
```
crai0, line 73: not sure if this was intended to be capitalized, but i figured it should be lower case for consistency

tests/commands/test_commit_command.py
```diff
@@ -109,6 +109,32 @@ def test_commit_command_with_dry_run_option(config, mocker: MockFixture):
         commit_cmd()
 
 
+@pytest.mark.usefixtures("staging_is_clean")
+def test_commit_command_with_write_message_to_file_option(
```
crai0, line 113: let me know if i should also create a negative test cases where the option is getting passed no value or an invalid value

commitizen/cli.py
woile: good

docs/tutorials/auto_prepare_commit_message.md
```diff
@@ -0,0 +1,27 @@
+# Automatically prepare message before commit
+
+## About
+
+To automatically prepare a commit message prior to committing, you can use a [Git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).
```
woile, line 5: I'm not familiar with this hook, could you add a bit more about the motivation, and explaining the workflow for this?

docs/commit.md
```diff
@@ -6,6 +6,10 @@ In your terminal run `cz commit` or the shortcut `cz c` to generate a guided git
 
 A commit can be signed off using `cz commit --signoff` or the shortcut `cz commit -s`.
 
+You can run `cz commit --write-message-to-file COMMIT_MSG_FILE` to additionally save the
+generated message to a file. This can be combined with the `--dry-run` flag to only
+write the message to a file and not modify files and create a commit.
+
```
woile, line 12: We should add a link to the tutorial here.

```
Learn how to [Automatically prepare message before commit](./tutorials/auto_prepare_commit_message.md)
```

commitizen/commands/commit.py
```diff
@@ -63,10 +64,18 @@ def prompt_commit_questions(self) -> str:
 
     def __call__(self):
         dry_run: bool = self.arguments.get("dry_run")
+        write_message_to_file = self.arguments.get("write_message_to_file")
 
         if git.is_staging_clean() and not dry_run:
             raise NothingToCommitError("No files added to staging!")
 
+        if write_message_to_file is not None:
```
woile, line 72: Can we merge this 2 lines?

```python
if write_message_to_file is not None and not isinstance(write_message_to_file, str)
```

I'm wondering if we should also check if it's a file 🤔 I guess the `smart_open` will raise an error

woile:
Great work! 💪🏻 thanks!

commitizen/commands/commit.py
crai0: `smart_open` in line 89 does raise an error if the path points to an existing directory but, considering that the user had to go through answering the questions at that point, it makes sense to already raise an error here. does the cli library support `pathlib.Path` as a type?

if not, i'd add a check like this:

```python
if write_message_to_file is not None:
    if not isinstance(write_message_to_file, str):
        raise NotAllowed(...)
    
    if os.path.isdir(write_message_to_file):
        raise NotAllowed(...)
```

crai0:
i noticed that it's not possible to use the `--retry` option with this workflow. if another hook aborts the commit after running `git commit` for example, the message generated with commitizen is lost. i would like to propose a follow-up issue in which the retry behavior can be configured as a default and a new `--no-retry` flag is added to override the configuration.

commitizen/commands/commit.py
```diff
@@ -63,10 +64,18 @@ def prompt_commit_questions(self) -> str:
...
+            if not isinstance(write_message_to_file, str):
+                raise NotAllowed(
+                    "Commit message file name is broken.\n"
+                    "Check the flag `--write-message-to-file` in the terminal"
+                )
```
crai0, line 77: `smart_open` in line 89 does raise an error if the path points to an existing directory but, considering that the user had to go through answering the questions at that point, it makes sense to already raise an error here. does the cli library support `pathlib.Path` as a type?

if not, i'd add a check like this:

```suggestion
        if write_message_to_file is not None:
            if not isinstance(write_message_to_file, str):
                raise NotAllowed(
                    "Commit message file name is broken.\n"
                    "Check the flag `--write-message-to-file` in the terminal"
                )
    
            if os.path.isdir(write_message_to_file):
                raise NotAllowed(f"{write_message_to_file} is a directory")
```

crai0: it seems like you're using `argparse` for [woile/decli](https://github.com/woile/decli) so we could also change the type of `write_message_to_file` to `pathlib.Path` and let `argparse` do the argument validation. then this would simplify to:

```suggestion
        if write_message_to_file is not None and write_message_to_file.is_dir():
            raise NotAllowed(f"{write_message_to_file} is a directory")
```

commitizen/cli.py
```diff
@@ -62,10 +62,15 @@
                         "action": "store_true",
                         "help": "show output to stdout, no commit, no modified files",
                     },
+                    {
+                        "name": "--write-message-to-file",
```
crai0, line 66: in case we're going with the `pathlib.Path` approach:

```suggestion
                        "name": "--write-message-to-file",
                        "type": Path,
```

commitizen/commands/commit.py
crai0: https://github.com/commitizen-tools/commitizen/pull/731#discussion_r1181007836

commitizen/commands/commit.py
woile: I like the idea of using `Path` 👍🏻

commitizen/cli.py
woile: Go for it

woile:
The `--no-retry` would be a breaking change, I'd rather keep the API. The `--retry` has been requested a lot, so I imagine a lot of people are already using it, and I'd rather avoid changing it.

Is it not possible to run other checks and then the commit? 

Is there a way to refactor the code in a non-breaking change?

crai0:
this would not be a breaking change if it can be configured and isn't be enabled by default. i'll open an issue for it and describe the feature in a backward compatible manner.

crai0:
i had to remove the hook i was using because i added it on accident but otherwise i think this can be merged, right?

crai0:
never mind, i couldn't stand the way the hook worked before so i made the following changes:

- have two separate hooks (`prepare-commit-msg` and `post-commit`) that can be downloaded from this repo
- the hooks are written in python (which is actually more cross-compatible than bash)
- the hooks check if commitizen is even installed
- it's possible to make commits using commitizen using both `git commit` and `cz commit`
- the hooks have a similar retry feature to the one that commitizen has and they're interoperable so that `cz commit --retry` can be used to retry the last commit made via `git commit` and vice versa.
- the tutorial now contains an installation guide and a feature overview

woile:
That's nice! Thanks for taking the time to build this. I'll take a look at it tomorrow

PR #686, opened 2023-03-13
Add `version_type` option to make version compatibly with semver

Apkawa:
## Description

I decided to use it in a rust project but cargo crashed when trying to use the prerelease versions.

I have added an optional setting that solves my problem for now.

### `version_type`

Choose version type

* `pep440` - default version provider.
    - prerelease - `1.0.1a0`
    - devrelease - `1.0.1dev0`
    - dev and pre - `1.0.1a0.dev0`
* `semver` - semver compatibly provider. Added "-" delimiter
    - prerelease - `1.0.1-a0`
    - devrelease - `1.0.1-dev0`
    - dev and pre - `1.0.1-a0-dev0`

TODO: renamed pre to long name, like alpha, beta, ..etc via additional option `semver_long`

```toml
[tool.commitizen]
version_type = "semver"
```

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes

## Steps to Test This Pull Request
```shell
$ cz bump --check-consistency --no-verify --prerelease alpha --version-type=semver --dry-run
bump: version 0.1.0-a2 → 0.1.0-a3
tag to create: v0.1.0-a3

```

## Additional context

#150

PR #733, opened 2023-05-01
Feature: version schemes endpoint

noirbizarre:
## Description

This PR is a continuation of #686 following [this comment](https://github.com/commitizen-tools/commitizen/pull/645#issuecomment-1500945092).

It does the following:
- rename `version_type` into `scheme` to follow specifications semantic ([PEP 440](https://peps.python.org/pep-0440/), [SemVer](https://semver.org/) but also [CalVer](https://calver.org/) which are all using the `scheme` term)
- expose a `commitizen.scheme` endpoint for version scheme plugins
- move some methods from `bump` to the plugin allowing support for all commitizen feature (especially incremental versioning which is relying on the ability to parse version as well as bumping which should be able to generate the next version according to the scheme)

This PR is also related to #645 as the incremental feature requires both the template and the version knowledge to be complete (aka. one of the 2 PR should be rebased after the merge of the other to properly handle this part)

## Checklist


commitizen/schemes.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # Self is Python 3.11+ but backported in typing-extensions
+    if sys.version_info < (3, 11):
+        from typing_extensions import Self
+
+
+DEFAULT_VERSION_PARSER = r"(?P<version>([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+)?(\w+)?)"
+
+
+class Version(BaseVersion):
```
woile, line 27: Why was the protocol removed? I think it was a great way to let anyone identify a way to implement their own version schema. Now it seems that we are forcing the use of BaseVersion

commitizen/cli.py
```diff
@@ -4,12 +4,12 @@
...
 from functools import partial
 from types import TracebackType
-from typing import List
+from typing import List, Optional
 
 import argcomplete
 from decli import cli
 
-from commitizen import commands, config, out, version_types
+from commitizen import commands, config, out, schemes
```
woile, line 12: I prefer `version_schemes`, rather than the lose `schemes`, it's more scoped

woile:
This looks really cool. I really like using the proper name: "scheme".

My main concern is that this a big breaking change, right? I already migrated a lot of projects to `version_type` at work and recently released a lot of `v3` related things for commitizen. I would rather do it in steps to give me some buffer:

1. Keep in the config `version_type` and `--version-type` with a deprecation warning:
> version-type will be deprecated in the next major release, please use "version-scheme" instead.

2. On `v4` we remove `version-type`

---

I also think using `version-scheme` in file names and as config and cli args, is more clear than "scheme". I don't think using just "scheme" gives enough context to users.

noirbizarre:
OK, I rebased and squash the PR, taking every remark into account (this includes the deprecation warnings).

Note: I wonder if exposing this as a cli parameter (`--version-scheme`, was `--version-type` before) makes sense: I can't see a use case where you have commitizen properly configured for your project and you need to provide another version scheme on the command line.

commitizen/schemes.py
noirbizarre: I removed it for multiple reason:
- we actually rely on almost all `Version` properties
- it felt like duplicating the upstream `Version` class which is already wrapping another one
- `Protocol` only provide type hinting, and while I get the "bring your own" point, when contributing I fill less helped by a protocol than a base class
- it was not documented so I felt it was less useful than the base class (I added some docstrings from the upstream `packaging.version`)
- it was not covering the entire scope (I completed it in the last push)
- there is no syntax to explicitely says a class implement a `Protocol`, unless you inherit it (which doesn't bring much value over inheriting directly the `packaging.version.Version`) but you can't inherit both `Version` (which we requires) and `VersionProtocol`
- `Protocol` are not enforced, they only provide type hinting on duck typing leaving developers with only one choice: have a test suite covering all the expected case (which I doubt given none of the existing plugins is covered) because MyPy won't be able to check that it should and implements the protocol (for the same reason, there is no syntax to express it)
- they don't exists in Python 3.7 while being supported by Commitizen

Given I get the "bring your own" point, I restored it in the last push and completed it (missing properties, docstrings...) and found a way to have proper type semantic with it.

noirbizarre:
@Lee-W @woile this one is ready as well as #645.
Whatever order you choose to merge, the second one will need a rebase to resolve conflict and integrate the other.

woile:
Looks really good, sorry for the delay. I will take a deeper look when i'm back from holidays (June). Thanks a lot 🚀

Lee-W:
This is fantastic! I am interested in delving into it further in the upcoming weeks. I have been extremely busy this year and feeling a bit overwhelmed but will try my best to help!

commitizen/version_schemes.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # Self is Python 3.11+ but backported in typing-extensions
+    if sys.version_info < (3, 11):
+        from typing_extensions import Self
+
+
+DEFAULT_VERSION_PARSER = r"v?(?P<version>([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+)?(\w+)?)"
+
+
+class VersionProtocol(Protocol):
```
woile, line 37: I was thinking the `VersionProvider(ABC)` is an `ABC` instead of `Protocol`, should we stick to ABC for consistency?

woile: btw how does `ABC` work with type inference? I'd like to have something that can type hint (this helps during development *if* you use mypy or similar), and that can also fail at runtime if not implemented 🤔 (this for those who don't use any kind of tpye hinting)

woile:
This looks awesome, I think it's ready to be merged (besides my comment). Thanks @noirbizarre

commitizen/version_schemes.py
noirbizarre: We need a compromise here:
- `ABC` forces you into inheritance, which is defeating protocol purpose (especially https://github.com/commitizen-tools/commitizen/pull/733#discussion_r1182835871) and it would force `Pep440` and `SemVer` into multiple inheritance
- `Protocol` is the opposite, it forces into nothing because by default it is ignored at runtime

However, there is one possibility which is interesting: we can make the protocol [runtime checkable](https://docs.python.org/3/library/typing.html#typing.runtime_checkable) and raise a warning or fail when calling `get_version_scheme`, something like:

```python
scheme = ep.load()
try:
    isinstance(scheme, VersionProtocol)
except TypeError:
    warnings.warn(f"{name} is not implementing the Version protocol")
return scheme
```

This way we would have best of both world:
- clean type hinting
- freedom of implementation
- runtime error or warning for devs not using MyPy

What do you thing of this solution ?

woile: I think it's great. I would also like to choose one between protocol and ABC. Do we need then to keep `versionProvider` as an `ABC`? or can we migrate (at some point in the future) to Protocol with runtime check?

noirbizarre: OK, it has been added in the latest push (with some tests)

I think there is nothing preventing from switching between `ABC` and `Protocol`. In both cases, this is a design choice, the same as above 👆🏼

noirbizarre:
PR updated with the protocol runtime check and the associated tests 👍🏼 
It should be ready to merge 🚀 🎉

commitizen/version_schemes.py
woile: Totally, I just think we should stick to one for plugins, to reduce cognitive load on those who will implement them.

woile:
Sorry for the delay, I'm a bit slow, this looks good to me, shall we merge?

woile (approved):

PR #740, opened 2023-05-10
fix: use only version tags in changelog

### LINKED ISSUE #741: Changelog can be generated with non-version tags as a header
### Description

Changelogs should be generated only with version tags as headers. Tags are indeed being checked that they are version tags when generating the changelog [here](https://github.com/commitizen-tools/commitizen/blob/v3.2.1/commitizen/changelog.py#L58), but this logic is missing for the _latest_ commit's tag.

This means that changelogs can be generated with an incorrect tag for the latest version

I've already opened a PR to fix this issue [here](https://github.com/commitizen-tools/commitizen/pull/740), I hope that's okay and is useful! 🙂 
I'm open to suggestions if you see anything to improve there of course

### Steps to reproduce

On a project with commitizen setup

1. touch some-change
2. `git commit -m "feat: some change"`
3. `git tag a-tag-that-has-nothing-to-do-with-versions`
4. `cz changelog --dry-run`

See that the output is
```sh
## a-tag-that-has-nothing-to-do-with-versions (2023-05-10)

### Feat

- some change
```

instead of 
```sh
## Unreleased
...
```
or
```sh
## 0.1.0 (2023-05-10)
...
```

### Current behavior

Non-version tags can appear as the `latest` version header of the changelog

### Desired behavior

Only version tags should ever appear in the changelog, for the latest change & subsequent changes

### Screenshots

_No response_

### Environment

Commitizen Version: 3.2.1
Python Version: 3.10.8 (main, Dec 27 2022, 10:58:47) [Clang 14.0.0 (clang-1400.0.29.202)]
Operating System: Darwin

sacha-c:
All tags used to generate the changelog must be version tags. 
This PR fixes two issues:
- The 'latest' commit was not being checked for it being a version tag, which can result in some non-version tag written as a header for the changelog
- The `rev-range` created when running `cz changelog 3.2.0` could use a non-version tag as a lower-bound, resulting in an incomplete changelog if the non-version tag was in the middle of two versions.

## Description
Changelogs should be generated only with version tags as headers. Tags are indeed being checked that they are version tags when generating the changelog [here](https://github.com/commitizen-tools/commitizen/blob/v3.2.1/commitizen/changelog.py#L58), but this logic is missing for the _latest_ commit's tag.

This means that changelogs can be generated with an incorrect tag for the latest version

For example, if I have a project with a commit "feat: some change" tagged with a tag "i_am_not_a_version", running `cz changelog` will generate something like:\
See that the output is
```sh
## i_am_not_a_version (2023-05-10)

### Feat

- some change
```

instead of 
```sh
## Unreleased
...
```
or
```sh
## 0.1.0 (2023-05-10)
...
```

## Checklist

- [x] Update the documentation for the changes (not relevant)

## Expected behavior
The changelog should only contain version tags

On a project with commitizen setup

1. touch some-change
2. `git commit -m "feat: some change"`
3. `git tag a-tag-that-has-nothing-to-do-with-versions`
4. `cz changelog --dry-run`

closes #741

### DIFF
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -30,14 +30,20 @@ import os
 import re
 from collections import OrderedDict, defaultdict
 from datetime import date
-from typing import TYPE_CHECKING, Callable, Iterable, cast
+from typing import TYPE_CHECKING, Callable, Iterable, Type, cast
 
 from jinja2 import Environment, PackageLoader
 
+from commitizen import out
 from commitizen.bump import normalize_tag
 from commitizen.exceptions import InvalidConfigurationError, NoCommitsFoundError
 from commitizen.git import GitCommit, GitTag
-from commitizen.version_schemes import DEFAULT_SCHEME, Pep440, InvalidVersion
+from commitizen.version_schemes import (
+    DEFAULT_SCHEME,
+    BaseVersion,
+    InvalidVersion,
+    Pep440,
+)
 
 if TYPE_CHECKING:
     from commitizen.version_schemes import VersionScheme
@@ -65,6 +71,19 @@ def tag_included_in_changelog(
         return False
 
     return True
+
+
+def get_version_tags(scheme: Type[BaseVersion], tags: list[GitTag]) -> list[GitTag]:
+    valid_tags: list[GitTag] = []
+    for tag in tags:
+        try:
+            scheme(tag.name)
+        except InvalidVersion:
+            out.warn(f"InvalidVersion {tag}")
+        else:
+            valid_tags.append(tag)
+
+    return valid_tags
 
 
 def generate_tree_from_commits(
--- a/commitizen/commands/changelog.py
+++ b/commitizen/commands/changelog.py
@@ -137,7 +137,7 @@ class Changelog:
         # Don't continue if no `file_name` specified.
         assert self.file_name
 
-        tags = git.get_tags()
+        tags = changelog.get_version_tags(self.scheme, git.get_tags())
         if not tags:
             tags = []
 
--- a/commitizen/commands/changelog.py
+++ b/commitizen/commands/changelog.py
@@ -137,12 +137,9 @@ class Changelog:
         # Don't continue if no `file_name` specified.
         assert self.file_name
 
-        tags = changelog.get_version_tags(self.scheme, git.get_tags())
-        if not tags:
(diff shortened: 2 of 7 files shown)


sacha-c:
My PR is not passing `coverage/patch` because my changes include code that is run conditionally based on the OS (`if os.name == 'nt`)

My code is actually covered through tests, but it doesn't end up in `coverage.xml` because codecov only uses coverage.xml generated from linux runs: https://github.com/commitizen-tools/commitizen/blob/5cd12cbd42c0fdb67d2bad40af1f5841e83377c3/.github/workflows/pythonpackage.yml#L32

Any suggestion on how to proceed?

Lee-W:
Hi @sacha-c sorry for the late reply. Both maintainers are a bit overwhelmed these days. I've verified this issue and am now going to take a deeper look at this PR. Thanks for your contribution!

Lee-W:
Hi @sacha-c , I tried to resolve the conflict due to the latest merged PRs, but it seems to fail on the windows instance. Could you please take a look at the windows issue? I think we're almost ready to merge this one 💪

sacha-c:
> Hi @sacha-c , I tried to resolve the conflict due to the latest merged PRs, but it seems to fail on the windows instance. Could you please take a look at the windows issue? I think we're almost ready to merge this one 💪

Thanks for checking this issue! 🙂 

I had a look at the error and at first glance it seems unrelated to the changes of this PR. The test failing is [test_bump_pre_commit_changelog](https://github.com/commitizen-tools/commitizen/actions/runs/5355721704/jobs/9714345813?pr=740#step:5:73), and the exception is thrown [when installing pre-commit](https://github.com/commitizen-tools/commitizen/actions/runs/5355721704/jobs/9714345813?pr=740#step:5:307)
`* Install prebuilt node (20.3.1) .Failed to download from https://nodejs.org/download/release/v20.3.1/node-v20.3.1-win-x64.zip`
Is there a way to re-run the test suite and see if that fixes it?

Lee-W:
Yep, you're right. After I reran it, it succeeded. 🙂

Lee-W (approved):
This fix looks good to me. 

@woile I'm planning to merge it this weekend. Let me know if you want to take a deeper look 🙂

woile (approved):

PR #754, opened 2023-06-03
feat: add support for cargo workspaces

woile:
## Description

This PR adds support for `Cargo.toml` workspaces.

When the root `Cargo.toml` of a project is:

```
[workspace.package]
version = "0.1.0"
```
In your apps you can have a `Cargo.toml` with:

```
[package]
version = { workspace = true }
```

## Checklist

### DIFF
--- a/commitizen/providers.py
+++ b/commitizen/providers.py
@@ -115,14 +115,25 @@ class PoetryProvider(TomlProvider):
 class CargoProvider(TomlProvider):
     """
     Cargo version management
+
+    With support for `workspaces`
     """
 
     filename = "Cargo.toml"
 
     def get(self, document: tomlkit.TOMLDocument) -> str:
-        return document["package"]["version"]  # type: ignore
+        try:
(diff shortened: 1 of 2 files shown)


commitizen/providers.py
```diff
@@ -115,14 +115,25 @@ def set(self, pyproject: tomlkit.TOMLDocument, version: str):
...
+
+    With support for `workspaces`
     """
 
     filename = "Cargo.toml"
 
     def get(self, document: tomlkit.TOMLDocument) -> str:
-        return document["package"]["version"]  # type: ignore
+        try:
+            return document["package"]["version"]  # type: ignore
```
Lee-W, line 126: would it be possible for us to try `document["package"].get("version")`? IMO, it's more readible

woile: The idea is to make it fail if any of the keys are not present

Lee-W: got it. Make sense 👍

Lee-W (approved):

Issue #757, opened 2023-06-06
`cz bump` with a breaking change after an RC does not automatically increase the major version


Alexander-Serov:
### Description

Trying to bump in the following repo structure:

- cz bump new
- realize we instead want a major version release, make some commits with a breaking change
- tag v0.5.1rc0
- some commits
- tag v0.5.0

```
> cz bump --dry-run
bump: version 0.5.1rc0 → 0.5.1
tag to create: v0.5.1
increment detected: MAJOR
```

`cz bump` correctly detects that there is a major version increase but doesn't actually do it. This looks like a bug.

### Steps to reproduce

1. Try to `cz bump` with a breaking change on top of the last bugfix rc tag

### Current behavior

Says it's a major increment, but just releases the rc: v0.5.1rc0 -> v0.5.1

### Desired behavior

Release the next major version: v0.5.1rc0 -> v1.0.0

### Environment

```
Commitizen Version: 3.2.2
Python Version: 3.8.16 | packaged by conda-forge | (default, Feb  1 2023, 16:01:13) 
[Clang 14.0.6 ]
Operating System: Darwin
```

CZ configuration:
```
$ cat ./.cz.toml
[tool]
[tool.commitizen]
annotated_tag = true
changelog_incremental = true
changelog_merge_prerelease = true
name = "cz_conventional_commits"
tag_format = "v$version"
update_changelog_on_bump = true
version_provider = "scm"
version_type = "pep440"
```

Alexander-Serov:
> Can you share your toml configuration?

Updated above.

woile:
I think it relates to #688 what do you think?

Lee-W:
I just tested it without encountering the same issue with the latest version (3.26.0). Note that we'll need to set `major_version_zero = false`. I'm closing this one. But feel free to reopen if you deem otherwise 🙂

Issue #688, opened 2023-03-16
Bump increments to new pre-release version from pre-release version when it wouldn't from final version


rpatterson:
### Description

Running `$ cz bump --prerelease beta` when the current version is a previous pre-release bumps the version even when there are no commits that would require it, e.g. only `build`/`ci` commits.

### Steps to reproduce

1. Start with a final version, e.g. `v0.0.1`
2. `$ git commit -m "ci: CI-specific change not requiring release"`
3. `$ cz bump --prerelease beta` -> Refuses to bump the version as expected 
4. `$ git commit -m "fix: Prospective fix requiring beta testing"`
5. `$ cz bump --prerelease beta` -> Bumps to `v0.0.2b0` as expected 
6. `$ git commit -m "build: Local devel change not requiring release"`
7. `$ cz bump --prerelease beta` -> Bumps unexpectedly to `v0.0.2b1`

### Current behavior

Maintainers can't use `$ cz bump` as a tool to interpret conventional commit messages to determine if a release is required when the current version is a pre-release.

### Desired behavior

Maintainers should be able to use `$ cz bump`, or `$ cz ...` in some other form, to determine if the conventional commit messages since the last pre-release require a new pre-release.

### Environment

```bash
(build) rpatterson@rpatterson:~/src/devel/python-project-structure$ cz version --report
Commitizen Version: 2.42.1
Python Version: 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
Operating System: Linux
```

jenstroeger:
I just noticed the same behavior: with the `--prerelease` switch commitizen only ever increments the version once and then continues to bump the pre-release counter.

There’s currently no official specification on how conventional commits should increment under a pre-release. However, comment https://github.com/conventional-commits/conventionalcommits.org/issues/398#issuecomment-969387687 makes a sensible suggestion for incrementing pre-release version numbers.

Any plan for `cz` to implement that soon?

woile:
I'm really not that convinced. I think we need a more formal definition of what a pre-release is and a longer explanation. I find it strange the possibility to go from `0.1.1a1 ` to `0.2.0a0 ` to `1.0.0a0` to `2.0.0a0` without ever doing a non-pre-release, feels like a strange workflow. Under which circumstance someone would develop a 3.0 without ever releasing 1.0?

I'm not against doing something about it per-se, but I think we need further thought on the topic.

[...]

Issue #751, opened 2023-05-23
Allow releasing after a release candidate without commits


Alexander-Serov:
### Description

In our company, we make a release candidate and if it works, perform the full release. Currently, this logic is not compatible with `cz bump` because it always requires to have commits. Here is an example of the log:

The release candidate bump:
```
$ cz bump -pr rc 
bump: version 0.6.0 → 0.6.1rc0
tag to create: v0.6.1rc0
increment detected: PATCH
[detached HEAD de92ff6] bump: version 0.6.0 → 0.6.1rc0
 1 file changed, 11 insertions(+)
```

Attempting to do a real release on the commit tagged as 0.6.1rc0:
```
$ cz bump
bump: version 0.6.1rc0 → 0.6.1
tag to create: v0.6.1
No commits found
```
and exits with the exit code 3.

### Possible Solution

Sometimes a release candidate is good. Let us just allow updating the changelog and tag that commit.

### Additional context

We are using `changelog_merge_prerelease = True` so I would expect a changelog entry with the same info as in 0.6.1rc0 to be created, committed and tagged.

We have also tried to ignore the error and force proceed with
```
cz bump -nr 3
```
as indicated in the Documentation, but that gave the same result.

aferreirap:
Our company would also be impacted by this feature!

Lee-W:
I just tested it with the latest version (3.26.0), and it seems to work fine. I'm closing this one. Please feel free to reopen if the feature does not work. Thanks! 🙂

PR #799, opened 2023-07-27
fix: properly bump versions between prereleases

eduardocardoso:
## Description

Properly bump version numbers between prereleases respecting the commit messages bump levels.

## Checklist

## Expected behavior

Bump command should bump the version number according to the commit messages since the last final version when bumping between prereleases.

start at version 0.1.0

1. git commit --allow-empty --message 'fix: a fix'
2. cz bump --prerelease alpha → bumps to `0.1.1a0`
3. git commit --allow-empty --message 'feat: a feature'
4. cz bump --prerelease alpha → bumps to `0.2.0a0` instead of `0.1.1a1`

## Additional context
#688

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -230,29 +230,12 @@ class Bump:
                     "To avoid this error, manually specify the type of increment with `--increment`"
                 )
 
-            # Increment is removed when current and next version
-            # are expected to be prereleases.
-            force_bump = False
-            if current_version.is_prerelease:
-                last_final = self.find_previous_final_version(current_version)
-                if last_final is not None:
-                    commits = git.get_commits(last_final)
-                    increment = self.find_increment(commits)
-                    base = last_final.increment_base(
-                        increment=increment, force_bump=True
-                    )
-                    if base != current_version.base_version:
-                        force_bump = True
-                elif prerelease:
-                    increment = None
-
             new_version = current_version.bump(
                 increment,
                 prerelease=prerelease,
                 prerelease_offset=prerelease_offset,
                 devrelease=devrelease,
                 is_local_version=is_local_version,
-                force_bump=force_bump,
             )
 
         new_tag_version = bump.normalize_tag(
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -24,7 +24,11 @@ from commitizen.exceptions import (
 )
 from commitizen.changelog_formats import get_changelog_format
 from commitizen.providers import get_provider
-from commitizen.version_schemes import InvalidVersion, get_version_scheme
+from commitizen.version_schemes import (
+    get_version_scheme,
+    InvalidVersion,
+    VersionProtocol,
+)
 
 logger = getLogger("commitizen")
 
@@ -228,8 +232,19 @@ class Bump:
 
             # Increment is removed when current and next version
             # are expected to be prereleases.
-            if prerelease and current_version.is_prerelease:
-                increment = None
+            force_bump = False
+            if current_version.is_prerelease:
+                last_final = self.find_previous_final_version(current_version)
+                if last_final is not None:
+                    commits = git.get_commits(last_final)
+                    increment = self.find_increment(commits)
+                    semver = last_final.increment_base(
+                        increment=increment, force_bump=True
+                    )
+                    if semver != current_version.base_version:
+                        force_bump = True
+                elif prerelease:
+                    increment = None
 
             new_version = current_version.bump(
                 increment,
@@ -237,6 +252,7 @@ class Bump:
                 prerelease_offset=prerelease_offset,
                 devrelease=devrelease,
                 is_local_version=is_local_version,
+                force_bump=force_bump,
             )
 
         new_tag_version = bump.normalize_tag(
@@ -398,3 +414,33 @@ class Bump:
         if self.no_verify:
             commit_args.append("--no-verify")
         return " ".join(commit_args)
+
+    def find_previous_final_version(
+        self, current_version: VersionProtocol
+    ) -> VersionProtocol | None:
+        tag_format: str = self.bump_settings["tag_format"]
+        current = bump.normalize_tag(
+            current_version,
+            tag_format=tag_format,
+            scheme=self.scheme,
+        )
+
+        final_versions = []
+        for tag in git.get_tag_names():
+            assert tag
+            try:
+                version = self.scheme(tag)
+                if not version.is_prerelease or tag == current:
+                    final_versions.append(version)
+            except InvalidVersion:
+                continue
+
+        if not final_versions:
+            return None
+
+        final_versions = sorted(final_versions)  # type: ignore [type-var]
+        current_index = final_versions.index(current_version)
+        previous_index = current_index - 1
+        if previous_index < 0:
+            return None
+        return final_versions[previous_index]
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -238,10 +238,10 @@ class Bump:
                 if last_final is not None:
                     commits = git.get_commits(last_final)
                     increment = self.find_increment(commits)
-                    semver = last_final.increment_base(
+                    base = last_final.increment_base(
                         increment=increment, force_bump=True
                     )
-                    if semver != current_version.base_version:
+                    if base != current_version.base_version:
                         force_bump = True
                 elif prerelease:
                     increment = None
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -100,6 +100,7 @@ class VersionProtocol(Protocol):
         prerelease_offset: int = 0,
         devrelease: int | None = None,
         is_local_version: bool = False,
+        force_bump: bool = False,
     ) -> Self:
         """
         Based on the given increment, generate the next bumped version according to the version scheme
@@ -166,7 +167,9 @@ class BaseVersion(_BaseVersion):
 
         return f"dev{devrelease}"
 
-    def increment_base(self, increment: str | None = None) -> str:
+    def increment_base(
+        self, increment: str | None = None, force_bump: bool = False
+    ) -> str:
         prev_release = list(self.release)
         increments = [MAJOR, MINOR, PATCH]
         base = dict(zip_longest(increments, prev_release, fillvalue=0))
@@ -175,7 +178,7 @@ class BaseVersion(_BaseVersion):
         # must remove its prerelease tag,
         # so it doesn't matter the increment.
         # Example: 1.0.0a0 with PATCH/MINOR -> 1.0.0
-        if not self.is_prerelease:
+        if not self.is_prerelease or force_bump:
             if increment == MAJOR:
                 base[MAJOR] += 1
                 base[MINOR] = 0
@@ -195,6 +198,7 @@ class BaseVersion(_BaseVersion):
         prerelease_offset: int = 0,
         devrelease: int | None = None,
         is_local_version: bool = False,
+        force_bump: bool = False,
     ) -> Self:
         """Based on the given increment a proper semver will be generated.
 
@@ -212,9 +216,21 @@ class BaseVersion(_BaseVersion):
             local_version = self.scheme(self.local).bump(increment)
             return self.scheme(f"{self.public}+{local_version}")  # type: ignore
         else:
-            base = self.increment_base(increment)
+            base = self.increment_base(increment, force_bump)
             dev_version = self.generate_devrelease(devrelease)
-            pre_version = self.generate_prerelease(prerelease, offset=prerelease_offset)
+            release = list(self.release)
+            if len(release) < 3:
+                release += [0] * (3 - len(release))
+            current_semver = ".".join(str(part) for part in release)
+            if base == current_semver:
+                pre_version = self.generate_prerelease(
+                    prerelease, offset=prerelease_offset
+                )
+            else:
+                base_version = cast(BaseVersion, self.scheme(base))
+                pre_version = base_version.generate_prerelease(
+                    prerelease, offset=prerelease_offset
+                )
             # TODO: post version
             return self.scheme(f"{base}{pre_version}{dev_version}")  # type: ignore
 
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -221,8 +221,8 @@ class BaseVersion(_BaseVersion):
             release = list(self.release)
             if len(release) < 3:
                 release += [0] * (3 - len(release))
-            current_semver = ".".join(str(part) for part in release)
-            if base == current_semver:
+            current_base = ".".join(str(part) for part in release)
+            if base == current_base:
                 pre_version = self.generate_prerelease(
                     prerelease, offset=prerelease_offset
                 )
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -167,27 +167,20 @@ class BaseVersion(_BaseVersion):
 
         return f"dev{devrelease}"
 
-    def increment_base(
-        self, increment: str | None = None, force_bump: bool = False
-    ) -> str:
+    def increment_base(self, increment: str | None = None) -> str:
         prev_release = list(self.release)
         increments = [MAJOR, MINOR, PATCH]
         base = dict(zip_longest(increments, prev_release, fillvalue=0))
 
-        # This flag means that current version
-        # must remove its prerelease tag,
-        # so it doesn't matter the increment.
-        # Example: 1.0.0a0 with PATCH/MINOR -> 1.0.0
-        if not self.is_prerelease or force_bump:
-            if increment == MAJOR:
-                base[MAJOR] += 1
-                base[MINOR] = 0
-                base[PATCH] = 0
-            elif increment == MINOR:
-                base[MINOR] += 1
-                base[PATCH] = 0
-            elif increment == PATCH:
-                base[PATCH] += 1
+        if increment == MAJOR:
+            base[MAJOR] += 1
+            base[MINOR] = 0
+            base[PATCH] = 0
+        elif increment == MINOR:
+            base[MINOR] += 1
+            base[PATCH] = 0
+        elif increment == PATCH:
+            base[PATCH] += 1
 
         return f"{base[MAJOR]}.{base[MINOR]}.{base[PATCH]}"
 
@@ -216,7 +209,20 @@ class BaseVersion(_BaseVersion):
             local_version = self.scheme(self.local).bump(increment)
             return self.scheme(f"{self.public}+{local_version}")  # type: ignore
         else:
-            base = self.increment_base(increment, force_bump)
+            if not self.is_prerelease:
+                base = self.increment_base(increment)
+            elif force_bump:
+                base = self.increment_base(increment)
+            else:
+                base = f"{self.major}.{self.minor}.{self.micro}"
+                if increment == PATCH:
+                    pass
+                elif increment == MINOR:
+                    if self.micro != 0:
+                        base = self.increment_base(increment)
+                elif increment == MAJOR:
+                    if self.minor != 0 or self.micro != 0:
+                        base = self.increment_base(increment)
             dev_version = self.generate_devrelease(devrelease)
             release = list(self.release)
             if len(release) < 3:
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -147,6 +147,12 @@ class BaseVersion(_BaseVersion):
         if not prerelease:
             return ""
 
+        # prevent down-bumping the pre-release phase, e.g. from 'b1' to 'a2'
+        # https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases
+        # https://semver.org/#spec-item-11
+        if self.is_prerelease and self.pre:
+            prerelease = max(prerelease, self.pre[0])
+
         # version.pre is needed for mypy check
         if self.is_prerelease and self.pre and prerelease.startswith(self.pre[0]):
             prev_prerelease: int = self.pre[1]
--- a/docs/bump.md
+++ b/docs/bump.md
@@ -113,6 +113,34 @@ Generate a **changelog** along with the new version and tag when bumping.
 cz bump --changelog
 ```
 
+### `--prerelease`
+
+The bump is a pre-release bump, meaning that in addition to a possible version bump the new version receives a
+pre-release segment compatible with the bump’s version scheme, where the segment consist of a _phase_ and a
+non-negative number. Supported options for `--prerelease` are the following phase names  `alpha`, `beta`, or
+`rc` (release candidate). For more details, refer to the
+[Python Packaging User Guide](https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases).
+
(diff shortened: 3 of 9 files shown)


Lee-W:
This might not be the behavior we want. May I know why you want commitizen to behave this way?

woile:
Could you update this list with more samples from this PR?

https://github.com/commitizen-tools/commitizen/blob/master/tests/test_version_scheme_pep440.py


@Lee-W the discussion in #688 talks about this change (if I understand correctly)

jenstroeger:
> @Lee-W the discussion in #688 talks about this change (if I understand correctly)

@Lee-W @woile yes that’s the issue where we talked about bumping pre-release versions.

rgarrigue:
Hi. We'd be quite interested in this fix. Is it possible to push it forward if @eduardocardoso can't ?

jenstroeger:
> Hi. We'd be quite interested in this fix. Is it possible to push it forward if @eduardocardoso can't ?

@rgarrigue I talked with Eduardo and he’s busy; I can take a look at this next week?

rgarrigue:
Would be cool, thanks !

WarKaa:
+1 Is it scheduled for release soon?

jenstroeger:
@WarKaa I’ll noodle on the PR tomorrow…

jenstroeger:
> @Lee-W Could you please rebase from the main branch? 

I rebased the PR on today’s main branch (commit e9647c7).

> @woile Could you update this list with more samples from this PR?
> 
> https://github.com/commitizen-tools/commitizen/blob/master/tests/test_version_scheme_pep440.py

You mean you’d like me to add a few more scenarios/examples to one of the lists of test cases?

Things left to do:

- [ ] Add a few more tests that test for `alpha`, `beta`, and/or `rc` pre-release suffixes

woile:
> You mean you’d like me to add a few more scenarios/examples to one of the lists of test cases?

Yes, to model all the new behavior on this pr 👍🏻

jenstroeger:
@Lee-W @woile I noticed this test case:

https://github.com/commitizen-tools/commitizen/blob/af11332b7ced5959493eb48323fbf38216e9a6c7/tests/test_version_scheme_pep440.py#L30

which I think isn’t right: if the current version is `1.0.0rc0` and I add a `fix:` commit and then bump the next release version (i.e. _no_ pre-release) then the next version should be `1.0.1`, no?

Similarly here:

https://github.com/commitizen-tools/commitizen/blob/af11332b7ced5959493eb48323fbf38216e9a6c7/tests/test_version_scheme_pep440.py#L63

where I expected `1.0.1` after the bump 🤔

Unfortunately, I didn’t find details on this problem over at the discussions for [semver](https://github.com/semver/semver) or [conventional commits](https://github.com/conventional-commits/conventionalcommits.org), so we can either raise that discussion or improvise. Intuitively, though, I think the above tests should be “fixed”?

woile:
My understanding is that a release candidate can still receive patches and fixes until it's ready, even breaking changes. That's why you'd go from `1.0.0a0` to `1.0.0`. There can be fixes between each "pre-release".

For example, the popular bootstrap does something like this. [see releases](https://github.com/twbs/bootstrap/releases)

Screenshot 2023-11-06 at 08 39 37

Same for axum (a popular rust web framework), [see releases](https://github.com/tokio-rs/axum/releases?page=6)

jenstroeger:
> For example, the popular bootstrap does something like this. [see releases](https://github.com/twbs/bootstrap/releases) […] Same for axum (a popular rust web framework), [see releases](https://github.com/tokio-rs/axum/releases?page=6)

Neither of them uses conventional commits and actually computes the next release string from a given tag and commit range. Both kinda make things up manually.

> My understanding is that a release candidate can still receive patches and fixes until it's ready, even breaking changes. That's why you'd go from `1.0.0a0` to `1.0.0`. There can be fixes between each "pre-release".

Let’s look at this from another angle: we have a final release version `1.0.0` and then push a `fix:` commit followed by a `feat:` commit. That means that given the release version and two commits, the next release version would be `1.1.0` — right?

Now suppose we want to publish a release candidate with every commit. Thus, we should see the following: `1.0.0` bumps to `1.0.1rc0` after the `fix:` commit, then bumps to `1.1.0rc1` (or `rc0` depending on convention?) after the `feat:` commit, then bumps to final release `1.1.0`.

So with every release candidate we need to take into account the history since the last final release (which is what this PR actually does [here](https://github.com/getpassport/commitizen/blob/8522b8ca552750658780b86fe5fbb6f74e90447d/commitizen/commands/bump.py#L233-L247)).

Which also means that the above test

https://github.com/commitizen-tools/commitizen/blob/af11332b7ced5959493eb48323fbf38216e9a6c7/tests/test_version_scheme_pep440.py#L30

is somewhat meaningless because we don’t know the history since the last final release version.

woile:
> is somewhat meaningless because we don’t know the history since the last final release version

If you are in `1.0.0rc0` you know that either the API is being stabilized or there have been breaking changes.  Any new patch or fix will be accumulated to the pre-release or will finally be released as `1.0.0`.


Help me understand, and let’s look at this from another angle again. As a user, I want to start releasing candidates for my next major version: `2.0.0`.
How would it look if we release a candidate coming from `1.9` for every following commit:
- breaking
- fix
- feat
- breaking
- fix
- feat

Could you list the versions until we make stable `2.0`?

woile:
Screenshot 2023-11-06 at 13 05 48

btw this is docusaurus using semver and cc
https://github.com/facebook/docusaurus/releases

jenstroeger:
> As a user, I want to start releasing candidates for my next major version: `2.0.0`.

The purpose of conventional commit messages (and commitizen, for that matter) is to track versioning for me, so that I don’t have to get [sentimental about versioning](https://web.archive.org/web/20230917180209/http://sentimentalversioning.org/). Unless I have a strict regiment over my developer team, in reality it’s quite tricky to exactly nail down that the next version will be this fix or that feature change and align all commits in a way such that the prediction is met exactly.

I would prefer to address issues and merge PRs and let the tools compute the next version for me based on the given commit history. I wouldn’t want to manipulate my commit history such that I can hit some arbitrary version number I fancy this month.

> btw this is docusaurus using semver and cc https://github.com/facebook/docusaurus/releases

Well… and they shipped a breaking change with the first set of commit which bumped from `2.4.3` to `3.0.0-rc.0` (correct) and then a bunch of more commits, none of which would possibly bump `3.0.0` again. They did the same with `2.0.0` but no release candidates in between major versions.

woile:
Let's focus on your initial question, that's what I'm trying to understand, the sentimental and other things are irrelevant, I want to understand the bumping logic for release candidates.

Now again, how is this:

> 2.0.0rc0 and all following commits (patch, minor, major) wouldn’t change that and only increment the rc.

different than this?

https://github.com/commitizen-tools/commitizen/blob/af11332b7ced5959493eb48323fbf38216e9a6c7/tests/test_version_scheme_pep440.py#L30

why would you want to go to 1.0.1 here? or are you suggesting that after `2.0.0rc0` you send a `fix:` and you skip `2.0.0` all together? would you go to `2.0.1`? you are still under a release candidate, `0..n` new changes should allow you to go to `2.0.0`

jenstroeger:
Let’s step through two separate scenarios: first without pre-releases, then with pre-releases.

### No pre-release versioning

Without pre-release a next version is defined by the conventional commit messages of the commits between the previous version tag and the latest commit. For example, suppose the following commit log:

- commit `bump: release 1.8.3 → 1.9.0` has a tag `"v1.9.0"` and that’s my previous release, then
- commit `fix: some fix`, then
- commit `feat: some feature`.

Now we bump the next final version which is based on the `fix:` and the `feat:` commits and therefore we get

- commit `bump: release 1.9.0 → 1.10.0` with tag `"v1.10.0"` 

That’s my new published release version.

### With pre-release versioning

Now I want to add pre-releases into this flow, and for this example I care only about release candidates (i.e. `rc` suffix). Furthermore, my workflow is fully automated so that whenever I merge a PR into my `staging` branch the release candidate is published automatically — see also comment https://github.com/commitizen-tools/commitizen/pull/800#issuecomment-1791951787 for an illustration. The above example would now be as follows:

- commit `bump: release 1.8.3 → 1.9.0` has a tag `"v1.9.0"` and that’s my previous release, then
- commit `fix: some fix` merges into the `staging` branch
- automatic commit `bump: release 1.9.0 → 1.9.1rc0` is added to `staging` with tag `"v1.9.1rc0"`

That’s my current, published version. If I now choose to create a final release (by merging `staging` into `main`) then that final version would be `1.9.1` which would be correct based on the commit history since the last final release `1.9.0`. However, instead of a final release we push another commit that triggers another release:

- commit `feat: some feature` merges into the `staging` branch
- automatic commit `bump: release 1.9.1rc0 → 1.10.0rc1` is added to `staging` with tag `"v1.10.0rc1"` 

And now I’m ready to publish the next final release of my package by merging `staging` to `main`, and that would bump to `1.10.0` and that’s correct based on the commit history since the last final release `1.9.0`.

noirbizarre:
I didn't started the code review yet but I read all linked threads and issues.

I am very mixed about this because you seem to expect the tool to work as you expect on a process which is not compliant with a prerelease description and some might consider a bleeding edge case.

## Why ?

Both [SemVer](https://semver.org/#spec-item-9) and [Pep440](https://peps.python.org/pep-0440/#pre-releases) describe a pre-release as a future release being stabilized. Pep440 even talk about pre-release cycle.
You describe a process on which you accept a feature on a pre-release, which doesn't seem to match the description and you expect a tool dedicated to release automation to have the edge case as nominal case.

Most of the time, when you enter a pre-release cycle, it goes with a branching model into which you don't accept feature anymore, only fixes to stabilize (we literary talk about a feature freeze). And to be fair, all companies and project I work on, for or with are using the same kind of process (feature freeze > pre-release cycle > release) so having this behavior by default would really be a major issue for all those.

This leads me to the following reasoning: a proper pre-release process shouldn't accept new features and as a consequence a tool handling prerelease shouldn't automatically update to the next version but warn that it shouldn't happen.

So in my opinion, this behavior should never ever be the default, but as I understand your use-case, an explicit opt-in option enabling this behavior seems acceptable.

## Proposal

To be honest, if I go down the full reasoning, to properly handle pre-release management, I think (but this is only my opinion):
- warn on bump if this case happen, something like: "You are trying to bump a pre-release which received new feature(s)" (but keep the actual behavior to avoid a breakage)
- provide some commit and pre-commit tooling to may enforce that (prevent commit if trying to commit something which is considered major/minor change in your commitizen plugin), but same, I would make this opt-in and explicit (I don't want people not being able to commit without explicitly knowing why)
- provide an explicit option+flag allowing that behavior (something like `allow-features-on-prereleases`, to be defined) for people wanting this behavior

@woile @Lee-W WDYT ?

@jenstroeger does it seem an acceptable plan ? From what I understand (I didn't read the PR code yet), this PR is providing the third point but as default behavior. Is the same PR but with a setting and a CLI flag to enable OK to you ?

jenstroeger:
Thanks @noirbizarre. While I understand your point (and agree with it to some extent) I also think that tools should not dictate process.

It’s not `cz` job to enforce this or that workflow; instead, it’s job is to bump version based on the current version and a range of commits. That’s it.

Now that bump has flags ([docs](https://commitizen-tools.github.io/commitizen/bump/#usage)): `--prerelease` and `--devrelease` which impact how the next bump is computed. And that’s what this discussion is about.

noirbizarre:
Note:
I think can already customize yourself this behavior by doing your own version scheme and overriding the `bump` method.

Cf.
 - https://github.com/commitizen-tools/commitizen/blob/master/commitizen/version_schemes.py#L96-L106
 - https://commitizen-tools.github.io/commitizen/bump/#version_scheme

It doesn't mean that is use case doesn't need to be discussed, but that you can already customize it (while we need to agree on how this PR and behavior should be handled).

It initially wasn't meant to, but this is the beauty of making things extensible: you don't know what will come from the community with those extensions. Here's a perfect illustration.

noirbizarre:
> No, I used some process to illustrate why the discussed behavior would make sense. The discussed behavior is how cz bump --prerelease beta (for example) is expected to behave in the face of a previous final release and one or more previous pre-releases.

Oh, I understood, and I tried to show you how it is your expectations' by showing you counter examples (the fact that PEP440 on which is based this soft is saying otherwise) and the fact that all 3 reviewers are reluctant to merge the PR in the current state might indicate that your expectations are not shared.

> You might want to look at @woile’s comment https://github.com/commitizen-tools/commitizen/pull/800#issuecomment-1657057450 where he suggests using another command-line switch to refine the behavior of cz bump under a --prerelease switch.
> 
> Again, that’s what we’re discussing here.

I did and I even linked the answer in my previous response, there:

> (seems like the same idea as https://github.com/commitizen-tools/commitizen/pull/800#issuecomment-1791998382 but with the opposite flag and nominal case)

As I said, I have no issue with your workflow, I have an issue with making it the nominal behavior. Same as I would have an issue with making the "prevent feature commit on pre-release" the nominal behavior. Both are something which are tied to teams' usage and expectations.
I strongly think these kinds of changes should be opt-in or extensions based, not the normal behavior, even more if it goes against the specs we are trying to help with.

commitizen/version_schemes.py
```diff
@@ -212,9 +216,21 @@ def bump(
...
         else:
-            base = self.increment_base(increment)
+            base = self.increment_base(increment, force_bump)
             dev_version = self.generate_devrelease(devrelease)
-            pre_version = self.generate_prerelease(prerelease, offset=prerelease_offset)
+            release = list(self.release)
+            if len(release) < 3:
+                release += [0] * (3 - len(release))
+            current_semver = ".".join(str(part) for part in release)
+            if base == current_semver:
```
noirbizarre, line 225:
```suggestion
            current_base = ".".join(str(part) for part in release)
            if base == current_base:
```

commitizen/commands/bump.py
```diff
@@ -395,3 +411,33 @@ def _get_commit_args(self):
         if self.no_verify:
             commit_args.append("--no-verify")
         return " ".join(commit_args)
+
+    def find_previous_final_version(
```
noirbizarre, line 415: Good API addition 👍🏼

commitizen/commands/bump.py
```diff
@@ -228,15 +232,27 @@ def __call__(self):  # noqa: C901
...
+            force_bump = False
+            if current_version.is_prerelease:
+                last_final = self.find_previous_final_version(current_version)
+                if last_final is not None:
+                    commits = git.get_commits(last_final)
+                    increment = self.find_increment(commits)
+                    semver = last_final.increment_base(
+                        increment=increment, force_bump=True
+                    )
+                    if semver != current_version.base_version:
```
noirbizarre, line 244: To keep the same naming convention (and try to contain semver wording to the version scheme implementation)

```suggestion
                    base = last_final.increment_base(
                        increment=increment, force_bump=True
                    )
                    if base != current_version.base_version:
```

noirbizarre (commented):
After review, I am OK with it (except for a few nitpicky naming suggestions 👇🏼), tests are passing, no changes for existings 👍🏼

woile:
Hey @jenstroeger don't take it wrong, but I still have trouble understanding your comment on this particular case, why should it be removed? what is wrong with it?

https://github.com/commitizen-tools/commitizen/blob/af11332b7ced5959493eb48323fbf38216e9a6c7/tests/test_version_scheme_pep440.py#L30

I follow the logic of your other examples, but they are always for non-breaking changes. And you said this test case is irrelevant. But this test case, would be for a breaking change, how would the logic be?

If my understanding is correct, from any given version `X`, you can go to 3 possible pre-release versions (`X + patch + pr`, `X + minor + pr`, `X + break + pr`), and any new pre-release won't go beyond one of those 3 pre-release versions (depending on the given semver increment given). The final version will be one of those possibilities without the pre-release. Now if the example is `1.0.0rc0` why wouldn't the next version be `1.0.0`?

This is how I think about the problem

```mermaid
graph LR;
    a[1.9.0]-->d[major: 2.0.0rc0]
    a[1.9.0]-->c[minor: 1.10.0rc0]
    a[1.9.0]-->b[patch: 1.9.1rc0]
    d --> e[major: 2.0.0rc1]
    c --> e
    c --> f[minor: 1.10.0rc1]
    b --> e
    b --> g[patch: 1.9.1rc1]
    b --> f
    e --> f1[2.0.0]
    f --> f1
    f --> f2[1.10.0]
    g --> f1
    g --> f2
    g --> f3[1.9.1]
```

And from `2.0.0rc0` I see only one option: `2.0.0`. According [to your comment](https://github.com/commitizen-tools/commitizen/pull/799#issuecomment-1794205637), you propose going to `2.0.1`


Apologies again for my struggle 😅 🙏🏻

jenstroeger:
Hi @woile — no worries, it’s great we have these discussions to clarify how `cz` envisions to bump. Unfortunately, none of the bumping behavior is specified by SemVer or Conventional Commits. (We should probably spark a discussion there?)

~What I’m trying to say is that given a pre-release tag and a commit, we can’t actually compute a next pre-release or final release tag because in both cases we need the history since the last final release. In your graph above, for every node that’s two or more commits away from the last final release (in the above case `1.9.0`) it is impossible to bump without considering the history since the last final release.~ After some staring and thinking and scratching my head, I think you’re right.

Looking at your graph, I think I see what you’re saying: if my current pre-release tag is

- `2.0.0rc1` then both minor and patch are `0` and therefore the accumulated commit history must contain at least one breaking change, any other minor and patch commit are included already and we just keep incrementing the `rcX` counter.
- `1.10.0.rc1` then only the patch version is `0` and therefore the accumulated commit history must contain at least one minor bump and no breaking change, and other minor and patch commit are included already and we just keep incrementing the `rcX`, unless the commit is a breaking change in which case we’d go to `2.0.0rc2` (or final release `2.0.0`).
- `1.9.1rc1` then only the patch version has been bumped and therefore the accumulated commit history must contain only patch bumps and no minor or breaking change, and the next pre-release bump can bump anything depending on the commit.

That makes sense 👍🏼  

With that in mind, the test case

https://github.com/commitizen-tools/commitizen/blob/af11332b7ced5959493eb48323fbf38216e9a6c7/tests/test_version_scheme_pep440.py#L30

tells us that between the last final release and this pre-release there was one commit (because the `rc0` counter is `0`) which was a breaking change (because we’re now at `1.0.0rc0` and both minor and patch are `0`) and thus a patch is already covered and we either bump to

- `1.0.0rc1` for the next pre-release, or
- `1.0.0` for the next final release.

Also, I think it makes a lot of sense to keep the pre-release increment counting up _every time_, for example if I’m at `1.10.0rc2` and I commit a `test:` or `chore:` then the next pre-release is `1.10.0rc3`.

commitizen/version_schemes.py
jenstroeger: Commit https://github.com/commitizen-tools/commitizen/pull/799/commits/9da9302445fe0ac5a4169dd0759c243a6885b28c.

commitizen/commands/bump.py
jenstroeger: Commit https://github.com/commitizen-tools/commitizen/pull/799/commits/9da9302445fe0ac5a4169dd0759c243a6885b28c.

woile:
Okay we are aligned then 👍🏻 It's good to go

Does it also close #757 and #751 ?

> test: or chore: then the next pre-release is 1.10.0rc3

probably yes, but wondering if it should be related to

jenstroeger:
@woile I have a change to the documentation pending, and a few more tests. Let me add tests to check the two other issues and then we’re good to merge after.

But also keep in mind that with this change we’ll have to merge #800 — or we can pull #800 into this PR?

jenstroeger:
I added a bunch of tests and had to refactor some code: `BaseVersion.bump()` didn’t handle all pre-release scenarios.

There’s one more open issue with the following tests:
```python
(("3.1.4", None, "alpha", 0, None), "3.1.4a0"),  # No pump, just adding the pre-release suffix.
(("3.1.4a0", "PATCH", "alpha", 0, None), "3.1.4a1"),  # UNEXPECTED!
(("3.1.4a0", "MINOR", "alpha", 0, None), "3.2.0a0"),
(("3.1.4a0", "MAJOR", "alpha", 0, None), "4.0.0a0"),
```
These tests pass.

So here we have a final version `3.1.4` and without any proper release commit (e.g. a `chore:`) we “bump” to pre-release version `3.1.4a0`. If the next commit is a `fix:` then we bump to `3.1.4a1` but should bump to `3.1.5a0` 🤔

I can push my changes for discussion, if it helps?

commitizen/commands/bump.py
jenstroeger: We should probably rename the function to `find_base_version()` because “base version” seems to be the term used in other places.

jenstroeger:
@woile @noirbizarre sorry for the delay: I didn’t receive an email notification for the 👍🏼 on the message. Anyway, I pushed my local updates which contain:

- A refactor of the bump code: lowered bumping into the `BaseVersion.bump()` method and cleaned up the callers of that function. That came out of existing tests that call `Pep440.bump()` instead of going through the CLI `bump` command line (which are separate tests I’ve not yet expanded).
- More tests that play through various scenarios that bump pre-releases, including the unexpected behavior I mentioned above.

Please do take a close look at the code and the current behavior, and let me know what you think. At this point, I believe that pre-release bumping _must_ take into account the last base version and not just the current tag.

woile (approved):

woile:
Regarding #800 I think you can pull it in this one as well

jenstroeger:
> Regarding #800 I think you can pull it in this one as well

Yes I agree.

I need to check, but also we should make sure that final published versions contain the _full_ changelog since the last final published version — not just the changelog since the last tag (which can be any one of the pre-releases).

noirbizarre:
Good to me, agreed on base version precedence over the tag.
I also agree for #800 inclusion.

For the `(("3.1.4a0", "PATCH", "alpha", 0, None), "3.1.4a1")` unexpected case, I'll tend to say that if we merge this PR, it needs to be consistent all the way and so result in:
```python
(("3.1.4a0", "PATCH", "alpha", 0, None), "3.1.5a0"),
```
But I don't have a strong opinion on this because to me this call is an edge case. I consider IRL this case needs to be safeguarded by a process decided projects maintainers: either you are in a stabilization phase where you increment a prerelease number, either you have published the stable version and you are doing a patch.

jenstroeger:
I rebased the PR, merged the changes from #800, and updated the documentation.

To address the following unexpected behavior:
```python
(("3.1.4", None, "alpha", 0, None), "3.1.4a0"),  # No pump, just adding the pre-release suffix.
(("3.1.4a0", "PATCH", "alpha", 0, None), "3.1.4a1"),  # UNEXPECTED!
```
I think we can’t uphold the assumption that a bump is _always_ context free. Instead, a bump needs to take into account the previous history back to the most recent _final_ release. (@woile that changes our previous conversation in [this comment](https://github.com/commitizen-tools/commitizen/pull/799#issuecomment-1801216513).)

> @woile Does it also close https://github.com/commitizen-tools/commitizen/issues/757 and https://github.com/commitizen-tools/commitizen/issues/751 ?

I’ll have to check.

woile:
I'll take a last look after holidays and merge

jenstroeger:
> @woile I'll take a last look after holidays and merge

No worries. Only point for discussion would be the unexpected behavior mentioned above.

chadrik:
This looks great.  Would love to see this merged soon!

PR #800, opened 2023-07-27
feat: preserve prerelease linearity

eduardocardoso:
## Description

Forbid going back in prerelease level, after a `beta` or `rc` prerelease is made the next prerelease without a version number bump will be the same type if a lower or equal level is provided.
a -> b -> rc

## Checklist

## Expected behavior

After a prerelease it will only allow to keep or increase the prerelease level.
A bump from `b0` with `--prerelease alpha` will generata a `b1` release instead of going back to `a0`.

Start at versions 0.1.0
1. `git commit --allow-empty -m 'fix: bug'`
2. `cz bump --prerelease alpha` → bumps to `0.1.1a0`
3. `cz bump --prerelease beta` → bumps to `0.1.1b0`
4. `cz bump --prerelease alpha` → bumps to `0.1.1b1`
5. `cz bump --prerelease rc` → bumps to `0.1.1rc0`
6. `cz bump --prerelease alpha` → bumps to `0.1.1rc1`
7. `cz bump --prerelease beta` → bumps to `0.1.1rc2`

### DIFF
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, ClassVar, Type, cast
 import importlib_metadata as metadata
 from packaging.version import InvalidVersion  # noqa: F401: Rexpose the common exception
 from packaging.version import Version as _BaseVersion
+from packaging.version import _parse_letter_version
 
 from commitizen.config.base_config import BaseConfig
 from commitizen.defaults import MAJOR, MINOR, PATCH
@@ -148,10 +149,21 @@ class BaseVersion(_BaseVersion):
         This function might return something like 'alpha1'
         but it will be handled by Version.
         """
+
+        pre_order = ["a", "b", "rc"]
+
         if not prerelease:
             return ""
 
         # version.pre is needed for mypy check
+
+        if self.is_prerelease and self.pre:
+            letter_version = _parse_letter_version(prerelease, offset)
+            if letter_version:
+                current_index = pre_order.index(self.pre[0])
+                new_index = pre_order.index(letter_version[0])
+                prerelease = pre_order[max(current_index, new_index)]
+
         if self.is_prerelease and self.pre and prerelease.startswith(self.pre[0]):
             prev_prerelease: int = self.pre[1]
             new_prerelease_number = prev_prerelease + 1
--- a/tests/commands/test_bump_command.py
+++ b/tests/commands/test_bump_command.py
@@ -211,8 +211,20 @@ def test_bump_command_prelease(mocker: MockFixture):
     mocker.patch.object(sys, "argv", testargs)
     cli.main()
 
-    tag_exists = git.tag_exist("0.2.0a0")
-    assert tag_exists is True
+    assert git.tag_exist("0.2.0a0")
+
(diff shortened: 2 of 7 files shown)


Lee-W:
Step 4 might be confusing. The logic makes sense. But we might need to design it in another way. Adding a `--force-increase` flag might make more sense? @woile @noirbizarre Would love to see what you think 🙂

woile:
I'm not a big fan of this change, because things start getting confusing. I'd rather raise an error of the wrong value is used. But, I also struggle to see if it's worth to enforce it. Maybe behind a setting "enforce_prereleases"

What's the motivation for this PR?

jenstroeger:
> What's the motivation for this PR?

This PR builds on top of PR https://github.com/commitizen-tools/commitizen/pull/799, and came from our [branching model](https://nvie.com/posts/a-successful-git-branching-model/) at work: PR branches merge into “downstream” branches and release packages on each branch. So,

- the `development` branch (publishes pre-release `alpha`) merges into
- the `testing` branch (publishes pre-release `beta`) which merges into
- the `staging` branch (publishes pre-release `rc`) which merges into
- the `main` branch (publishes the final release).

Each of these downstream merges automatically creates a bump commit on the target branch and a tag on the bump commit. However, to ensure a consistent history across branches, upstream branches _always_ automatically rebase on top of downstream branches. I illustrated @eduardocardoso’s example above:

cz-800

On the right side of the image you have the growing `git log` history where you can see the `bump:` commits; that history is _the same_ on all branches.

Because we continuously rebase downstream branches, we quickly get into a place where the `development` branch has a `bump: 0.1.1rc0` commit (including tag) — the next commit to `development` _must_ bump that `rc0` to preserve linearity: it can not bump below `rc0`. Does that make sense?

woile:
I see 🤔 I'm fine with it as long as proper documentation is added, I can already foresee users complaining that doing `cz bump --prerelease alpha → bumps to 0.1.1b1` raises the wrong prerelease, and I won't know why 😆 

I think we are taking an opinionated approach on how it should be done, and that's what I would like to see reflected in the docs. I'm not a big user of prereleases myself, so I trust you on this model.

jenstroeger:
Agreed — and it probably makes sense to add a command-line switch that enables [precedence](https://semver.org/#spec-item-9) bumping. For example if the current version is `0.1.1b0` then

- `cz bump --prerelease beta` bumps to `0.1.1b1`
- `cz bump --prerelease alpha` is an error because a beta tag was expected
- `cz bump --prerelease` (no argument!) bumps to to `0.1.1b1` or
- `cz bump --prerelease alpha --consider-precedence` also bumps to `0.1.1b1`

We can either make the argument to `--prerelease` optional in which case it detects and bumps based on the tag with the highest precedence, or we add an argument that explicitly enables bumping based on the tag with the highest precedence.

woile (approved):
LGTM

woile (changes requested):
Please add a new section about `pre-releases` inside the bump command explaining "prerelease linearity" 🙏🏻

jenstroeger:
We can close this PR: I have merged it into PR #799 with commit https://github.com/commitizen-tools/commitizen/pull/799/commits/b9e768ed093c06136f6ead06e13e2f74e8d31058.

Lee-W:
Close as suggested

Issue #554, opened 2022-08-10
Give the user the possibility to commit without add files to staging before.


buhtz:
### Description

I am new to cz.

Usually I commit via `git commit -a`. So every tracked and un-staged file is staged automaticllly.

It seems to me that this is not possible with cz. I got the message "No files added to staging!".

I also tried `cz commit -- -a` which gives me an "unrecognized command" error.

### Possible Solution

Give the user the possibility to commit without add files to staging before.

### Additional context

#418

Lee-W:
I'm week against this feature. IMO, deciding which files should be include in a commit is an important thing.  @Woile what do you think?

Woile:
I think we've decided a long time ago not to use `-a`. I suggest 2 alternatives:

1. If you really wanna use cz commit, then create an alias: `alias czc="git add -u && cz commit"`
2. Use `fix/feat` in your commits and skip the interactive prompt: `git commit -am "feat: add shiny thing"`

On the other hand we added `-s` to commit, maybe we can add `-a` as well which is a commonly used param 🤔

Lee-W:
If @Woile 's good with it, I'm good with it as well :) I guess we'll need to make some change to https://github.com/commitizen-tools/commitizen/blob/master/commitizen/git.py and https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/commit.py. It'd be great if you could sent us a PR :)

Lee-W:
I think this one is finished in #821. Close this one. Thanks for @ttw225's reminder

PR #821, opened 2023-08-20
feat(commands): add cz commit arg to automatically stage every tracked and un-staged file

Xiao75896453:
## Description
Adding `cz commit` arg to  automatically stage every tracked and un-staged file

## Checklist

## Expected behavior

command `cz commit -a` will automatically stage every tracked and un-staged file .

1. cz commit -a

## Additional context

#554

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -92,8 +92,8 @@ def tag(tag: str, annotated: bool = False, signed: bool = False) -> cmd.Command:
     return c
 
 
-def add() -> cmd.Command:
-    c = cmd.run("git add .")
+def add(args: str = "") -> cmd.Command:
+    c = cmd.run(f"git add {args}")
     return c
 
 
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -92,7 +92,7 @@ def tag(tag: str, annotated: bool = False, signed: bool = False) -> cmd.Command:
     return c
 
 
-def add():
+def add() -> cmd.Command:
     c = cmd.run("git add .")
     return c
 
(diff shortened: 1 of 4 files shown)


commitizen/git.py
```diff
@@ -92,6 +92,11 @@ def tag(tag: str, annotated: bool = False, signed: bool = False) -> cmd.Command:
     return c
 
 
+def add() -> cmd.Command:
+    c = cmd.run("git add .")
```
woile, line 96: I don't think this is what `-a` does. In this case even non-tracked files are being added. `-a` should happen on `git commit -a`

Xiao75896453: Thank you! I misunderstood what git commit -a usage. I will modify this PR.

woile:
I wonder what's the advantage of doing `git add -u` vs doing `git commit -a`?
The plus side of `git commit -a` is that we call git once, does `git add` give some kind of different ergonomics?

Xiao75896453:
I do `git add -u` instead of `git commit -a `because of `if git.is_staging_clean() and not dry_run`.
If doing `git commit -a`, I think I may need to modify `if git.is_staging_clean() and not dry_run` and handle the simultaneous existence of `git commit -s` and `git commit -a.`
I feel it is more intuitive to use `git add -u`

Lee-W:
> I do `git add -u` instead of `git commit -a `because of `if git.is_staging_clean() and not dry_run`. If doing `git commit -a`, I think I may need to modify `if git.is_staging_clean() and not dry_run` and handle the simultaneous existence of `git commit -s` and `git commit -a.` I feel it is more intuitive to use `git add -u`

yep, agree with your point. @woile what do you think?

woile:
Sounds good

woile (approved):

Lee-W (approved):

Issue #785, opened 2023-07-14
Configure `cz commit` to add `--signoff` by default


schlotter:
### Description

Many projects require developers to sign-off every commit in order to acknowledge the project's Developer Certificate of Origin (DCO). Thus, it would be useful to configure commitizen in a way so that each commit is automatically signed-off, so that the `-s`/`--signoff` options do not manually be supplied to `cz commit`.

### Possible Solution

1. Developer sets `always_signoff` setting to `true`.
2. Developer runs `cz commit` without any arguments.
3. Developer answers all questions of commitizen.
4. A Git commit is created and its message contains
   ```
   Signed-off-by: Firstname Lastname <firstname.lastname@example.com>
   ```

Lee-W:
Hi @schlotter, thanks for proposing this issue. It sounds good to me. Would be much appreciated if you're interested in sending a PR 🙂

Lee-W:
Thanks @JamieLee0510 for the reminder. Just verified this issue was merged long a ago and resolved

PR #824, opened 2023-08-24
feat(defaults.py): add always_signoff config option for commits

schlotter:
## Description
Add new configuration option `always_signoff`. When is set to `true`, commits will be always signed off. The `-s` option does not have to specified to `cz commit`.

## Checklist

## Expected behavior
* When `always_signoff` configuration option is set to `true`, always sign off commits created with `cz commit`, even if `-s` option has not been specified.

1. Add `always_signoff` option to `true` in configuration file.
2. Create commit using `cz commit` (without `-s` option).
3. Verify the resulting commit message contains `Signed-off-by:`. 

## Additional context
#785

### DIFF
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -63,6 +63,7 @@ _settings = {
     "post_bump_hooks": ["scripts/slack_notification.sh"],
     "prerelease_offset": 0,
     "encoding": "utf-8",
+    "always_signoff": False,
 }
 
 _new_settings = {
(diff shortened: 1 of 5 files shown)


schlotter:
I was not able to follow all contribution guidelines because commands like `poetry run pre-commit install` would fail.

tests/test_conf.py
```diff
@@ -87,6 +88,7 @@
     "post_bump_hooks": ["scripts/slack_notification.sh"],
     "prerelease_offset": 0,
     "encoding": "utf-8",
+    "always_signoff": False,
```
Lee-W, line 91: Could we add a test case to verify that when `always_signoff` is set to true, we're actually running commit with signoff?

Lee-W (approved):
@schlotter Thanks for your contribution! 

@woile I'm good with this PR and plan to merge it these days. Let me know if you have any thought

woile (approved):

PR #842, opened 2023-09-11
ci(github-actions): fix sponsorship action token used

Lee-W:
## Description
it should be PERSONAL_ACCESS_TOKEN instead of PERSONAL_ACCESS_TOKEN_FOR_ORG

## Checklist


woile:
I created a new token for this which requires org permissions according to the GitHub action. I think we need to git add the readme. I'm not sure why it didn't work though

Lee-W:
hmmmm... that's weird. 🤔 might need to do some experiment

PR #903, opened 2023-11-06
ci(github-actions): move the generate sponsor step to docpublish workflow

Lee-W:
## Description

move the "generate sponsor" step to docpublish workflow (not sure whether it's the root cause 🤔)

## Checklist

## Expected behavior

the sponsorship list generated correctly

### DIFF
--- a/.github/workflows/label_pr.yml
+++ b/.github/workflows/label_pr.yml
@@ -10,4 +10,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+      with:
+        sparse-checkout: |
+          .github/labeler.yaml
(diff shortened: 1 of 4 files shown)


.github/workflows/label_pr.yml
```diff
@@ -9,4 +9,5 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@v3
```
noirbizarre, line 12: Seems extracted from #926

noirbizarre (approved):

.github/workflows/label_pr.yml
Lee-W: ah yes, forget to checkout back to main when worked on this

PR #926, opened 2023-12-03
ci(github-actions): fix laber-pr by adding necessary checkout step

Lee-W:
## Description

some of our PR are not correctly labeled, which is due to missing "checkout" step
https://github.com/actions/labeler#using-configuration-path-input-together-with-the-actionscheckout-action

## Checklist


noirbizarre (approved):
LGTM (But need conflict resolution before)

Lee-W:
> LGTM (But need conflict resolution before)

Nice catch! It encounters another conflict after I resolved the first one

PR #963, opened 2024-01-17
feat(provider): add new maven provider for handling versions from pom.xml

rshmhrj:
## Description
Adding a new maven provider
Will also be looking into adding a maven version scheme (SemVer + Qualifier) for use in java projects

## Checklist

- [ x] Add test cases to all the changes you introduce
- [ x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [ x] Test the changes on the local machine manually
- [ x] Update the documentation for the changes

## Expected behavior
On `cz init` have the option to choose `mvn` to read and write versions to the `pom.xml` file

Versioning ref: https://octopus.com/blog/maven-versioning-explained

1. Open a java project with a `pom.xml` in the root dir
  1. Make tag if not already present (`git tag v0.0.1-SNAPSHOT`)
1. Activate poetry venv `source /path_to_poetry_virtualenvs/commitizen-4BNoRlBJ-py3.12/bin/activate`
1. Run cz init directly `python3 /path_to_commitizen_repo/commitizen/cli.py init`

`mvn` is now an option to choose from

image

Can currently successfully add with normal SemVer (minus qualifiers)
image

## Additional context
- [Original Discussion post](https://github.com/commitizen-tools/commitizen/discussions/962)
- not sure how this will pass in CI without some sort of mvn wrapper installed though...

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -47,6 +47,10 @@ class ProjectInfo:
         return os.path.isfile("Cargo.toml")
 
     @property
+    def is_maven_project(self) -> bool:
+        return os.path.isfile("pom.xml")
+
+    @property
     def is_npm_package(self) -> bool:
         return os.path.isfile("package.json")
 
@@ -222,6 +226,7 @@ class Init:
             "commitizen": "commitizen: Fetch and set version in commitizen config (default)",
             "cargo": "cargo: Get and set version from Cargo.toml:project.version field",
             "composer": "composer: Get and set version from composer.json:project.version field",
+            "mvn": "mvn: Get and set version from pom.xml:project.version field",
             "npm": "npm: Get and set version from package.json:project.version field",
             "pep621": "pep621: Get and set version from pyproject.toml:project.version field",
             "poetry": "poetry: Get and set version from pyproject.toml:tool.poetry.version field",
@@ -236,6 +241,8 @@ class Init:
                 default_val = "pep621"
         elif self.project_info.is_rust_cargo:
             default_val = "cargo"
+        elif self.project_info.is_maven_project:
+            default_val = "mvn"
         elif self.project_info.is_npm_package:
             default_val = "npm"
         elif self.project_info.is_php_composer:
--- a/commitizen/providers/__init__.py
+++ b/commitizen/providers/__init__.py
@@ -11,6 +11,7 @@ from commitizen.providers.base_provider import VersionProvider
 from commitizen.providers.cargo_provider import CargoProvider
 from commitizen.providers.commitizen_provider import CommitizenProvider
 from commitizen.providers.composer_provider import ComposerProvider
+from commitizen.providers.mvn_provider import MavenProvider
 from commitizen.providers.npm_provider import NpmProvider
 from commitizen.providers.pep621_provider import Pep621Provider
 from commitizen.providers.poetry_provider import PoetryProvider
@@ -21,6 +22,7 @@ __all__ = [
     "CargoProvider",
     "CommitizenProvider",
     "ComposerProvider",
+    "MavenProvider",
     "NpmProvider",
     "Pep621Provider",
     "PoetryProvider",
--- a/commitizen/providers/mvn_provider.py
+++ b/commitizen/providers/mvn_provider.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+import subprocess
+
+from commitizen.providers.base_provider import VersionProvider
+
+
+class MavenProvider(VersionProvider):
+    """
+    Maven version management
+
+    ref: https://octopus.com/blog/maven-versioning-explained
+
+    Major.Minor.Patch-BuildNumber-Qualifier
+
+    Precedence:
+     - alpha or a
+     - beta or b
+     - milestone or m
+     - rc or cr
+     - snapshot
+     - (the empty string) or ga or final or release
+     - sp
+    """
+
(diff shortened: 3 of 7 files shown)


noirbizarre:
Glad to see people are finding the version provider useful.

BUT, it has been design so extra provider can be created as separate plugins to avoid putting all the maintaince load to the team. I don't know if the team is willing to support lots of providers for stacks they might not use and won't be able to maintain.

My advice would the following: publish this provider as a standalone `commitizen-maven` package and register it on [this documentation page](https://github.com/commitizen-tools/commitizen/blob/master/docs/third-party-commitizen.md) so people can easily discover it. This way it will be easier to maintain and update independantly. From you point of view, it won't change a thing: same code, just published as a separate package, you just need to install it with `commitizen` and it works with the python package, the pre-commit hook with `additional_dependencies` or the action with `extra_requirements`.

@Lee-W @woile What's your poinion on this ?

woile:
I agree, extra providers should be provided on a different repo 🙏🏻

noirbizarre:
I have been thinking a lot about this one.

I think we need to apply our own recommendations, given we already support `npm`, `composer` and `cargo` (I don't include Python providers in this list because commitizen is a Python tool, installable with Python tooling and natively able to read its configuration from official `pyproject.toml`), to be consistent we should either apply the same decision to those 2 provides and publish then as external package, either accept this PR.

Furthermore, I see benefit in externalizing those 3 providers:
- dogfooding: we apply to ourselves what we recommend to other, meaning that most of the time, issues on the pattern would be already fixed
- samples: anyone willing to craft a provider would have 3 examples of existing external providers
- clarity: this thread shows that it is hard to draw the line between what should be accepted and what should be external. This would give a single rule to apply
- fairness: hard to tell, no we won't accept this provider while we already have 3 non-python providers

WDYT ?

rshmhrj:
@noirbizarre Thanks for going through the thought process behind externalizing a provider -- I understand the increased burden and stress of having to make updates to a provider you don't have any experience or proficiency in.

I have a few semi-related questions:

- what is your recommendation for new versions (e.g. SemVer + qualifier (SV+Q) like `3.2.1-SNAPSHOT`) and where they should be developed and maintained?
- ideally the maven provider would be able to work hand-in-hand with that SV+Q version -- should they both be in the same project?
- if mvn and SV+Q are bundled, would it be possible to simultaneously use a different provider with SV+Q?

Lee-W:
> I have been thinking a lot about this one.
> 
> I think we need to apply our own recommendations, given we already support `npm`, `composer` and `cargo` (I don't include Python providers in this list because commitizen is a Python tool, installable with Python tooling and natively able to read its configuration from official `pyproject.toml`), to be consistent we should either apply the same decision to those 3 providers and publish them as external packages, either accept this PR.
> 
> Furthermore, I see benefit in externalizing those 3 providers:
> 
>     * dogfooding: we apply to ourselves what we recommend to other, meaning that most of the time, issues on the pattern would be already fixed
> 
>     * samples: anyone willing to craft a provider would have 3 examples of existing external providers
> 
>     * clarity: this thread shows that it is hard to draw the line between what should be accepted and what should be external. This would give a single rule to apply
> 
>     * fairness: hard to tell, no we won't accept this provider while we already have 3 non-python providers
> 
> 
> WDYT ?

I think this is definitely something we want. We probably would want to remove them after 4.0 as this is a breaking change, but we could start making these providers as standalone packages and add deprecation warning in the main repo

Lee-W:
> @noirbizarre Thanks for going through the thought process behind externalizing a provider -- I understand the increased burden and stress of having to make updates to a provider you don't have any experience or proficiency in.
> 
> I have a few semi-related questions:
> 
>     * what is your recommendation for new versions (e.g. SemVer + qualifier (SV+Q) like `3.2.1-SNAPSHOT`) and where they should be developed and maintained?

Do you mean the version of `commitizen-maven`? I think so.

>     * ideally the maven provider would be able to work hand-in-hand with that SV+Q version -- should they both be in the same project?


I though maven provider is only for finding where maven defines the version. Or am I missing anything?

>     * if mvn and SV+Q are bundled, would it be possible to simultaneously use a different provider with SV+Q?

@noirbizarre Not sure whether we have this support at this moment. 🤔 but may be something we can work on

marcosschroh:
Thanks for the work @rshmhrj . Are you planning to create a package for this new provider?  Without the provider, how are you changing the version in the `pom.xml`? In my case I am adding `metadata` to the `version` attribute in the `pom.xml` and then using it in the commitizen config, for example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>my-sdk</artifactId>
    <version commitizen="true">1.0.0</version>
    ...
</project>
```

and then in `.cz.toml`

```toml
version_files = [
    "pom.xml:commitizen",
]
```

Lee-W:
as it is still marked as draft and the CI has not changed. I'll change it to draft

noirbizarre:
Given the comments above, here's what I propose:
- merge this as soon as CI is passing
- start preparing external providers as standalone libraries for the next major
- maybe plan a transitional release marking the built-in one meant to be extracted as deprecated

Perhaps we should make an org project to plan this properly and not forget anything

bearomorphism:
Hi @rshmhrj , what is the status of this PR? Thanks!

bearomorphism:
This is indeed a good feature, but I'm closing this one since we want to externalize the providers

Thanks

PR #975, opened 2024-01-30
Fix SCM provider to only search tags that are reachable by the current commit

chadrik:
## Description

The scm provider currently uses _all_ of the tags in the repo to determine the current version, but it should instead only consider tags that are upstream from the current commit.   This means commitizen currently consider tags that are in the future or on other branches when determining the version.  This doesn't make sense.  The pep621 provider doesn't determine the current version from the _latest_ commit of `pyproject.toml` or from commits on other branches, it looks at the current state of the repo. 

## Checklist

## Expected behavior

The expected behavior is that the current version should be determined by the "highest" version tag reachable by the current commit.  In order to determine "highest" I had to add sortability to `VersionProtocol`, but luckily this is already implemented by `BaseVersion`.

### DIFF
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -4,7 +4,7 @@ import re
 import sys
 import warnings
 from itertools import zip_longest
-from typing import TYPE_CHECKING, ClassVar, Protocol, Type, cast, runtime_checkable
+from typing import TYPE_CHECKING, Any, ClassVar, Protocol, Type, cast, runtime_checkable
 
 import importlib_metadata as metadata
 from packaging.version import InvalidVersion  # noqa: F401: Rexpose the common exception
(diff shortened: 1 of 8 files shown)


woile (approved):

chadrik:
codecov is failing, but I don't know what else to test.  Any ideas?

commitizen/version_schemes.py
```diff
@@ -93,6 +93,24 @@ def micro(self) -> int:
         """The third item of :attr:`release` or ``0`` if unavailable."""
         raise NotImplementedError("must be implemented")
 
+    def __lt__(self, other: Any) -> bool:
```
woile, line 96: Why were these functions added? I don't see where they are used. Could you add a comment explaining a bit?

- Why are they important?
- When should a developer implement them?

chadrik: The comparators have been added to `VersionProtocol` so that versions can be sorted, e.g. to find the highest version. 

This was actually already used in `bump.Bump.find_previous_final_version`, but the author simply ignored the mypy error:

```python
        final_versions = sorted(final_versions)  # type: ignore
        current_index = final_versions.index(current_version)
        previous_index = current_index - 1
        if previous_index < 0:
            return None
        return final_versions[previous_index]
```

By adding the comparators to the protocol, I was able to remove the `# type: ignore` above.

My change also requires the ability for versions to be sorted, here:

```python
    def get_version(self) -> str:
        matcher = self._tag_format_matcher()
        matches = sorted(
            version
            for t in get_tags(reachable_only=True)
            if (version := matcher(t.name))
        )
        if not matches:
            return "0.0.0"
        return str(matches[-1])
```

I added some tests for sortability.

chadrik: > When should a developer implement them?

since they are part of the `VersionProtocol`, they should always be implemented.  They are required for `bump` and scm provider to work correctly.

chadrik:
I keep adding tests but the codecov is not improving.  Any suggeestions?

commitizen/version_schemes.py
woile: Thanks for the information. I should've been more clear, I meant a docstring, not a comment on github 😅 My question was aimed more at adding docstrings to help those developers implementing a custom `VersionProtocol`. If you have time to add them in a following PR that would be awesome, but I think we can merge this

Issue #845, opened 2023-09-14
Tag name validation issue on `cz bump --changelog`


josuerzslzr:
### Description

When generating changelog, `get_version_tags` function from `commitizen/changelog.py` is validating existing tags as if they were version names with no additional text on them. This behavior conflicts with the feature of customizing `tag_format`

### Steps to reproduce

1. Use some custom prefixed tag_format in .cz.yaml, 
`tag_format: some-project-release-$version`

2. Do initial version commit, bump and generate changelog: `cz bump --changelog`
4. Do subsequent fix/feat commit, bump and generate changelog: `cz bump --changelog`

ER: Version bumped, changelog generated.
AR: Error saying `No tag found to do an incremental changelog`

### Current behavior

get_version_tags function from commitizen/changelog.py is validating tag names as if they where version names, hence it's not able to return existing custom tags, an error appears saying `No tag found to do an incremental changelog`

### Desired behavior

Allow the changelog to generate regardless of tag name custom prefixes, possible solution could be to exclude the tag name prefix when checking with `scheme(tag.name)`

### Environment

Commitizen 3.8.2
Python 3.10.12

hermanho:
When can fix it? I cannot use commitizen in a monorepo repo and the config tag_format is useless.

bearomorphism:
Triage from #1964: Looks fixed on master (4.15.1). `commitizen/tags.py` now contains a `TagRules` class that handles custom `tag_format` (including arbitrary prefixes like `some-project-release-$version`) via regex-based `is_version_tag` / `extract_version`. Could anyone confirm a fresh repro is now passing? If so, this can be closed (and #1149 along with it).

bearomorphism:
**Verification update (re #1964)**

I reproduced this against current master (4.15.1):

Config (`.cz.yaml`): `cz_conventional_commits` with `tag_format: "some-project-release-$version"` and `update_changelog_on_bump: true`.

After two bumps:

`git tag --list`:
```
some-project-release-0.2.0
some-project-release-0.2.1
```

`CHANGELOG.md`:
```markdown
## some-project-release-0.2.1 (2026-05-09)

### Fix

- another change

## some-project-release-0.2.0 (2026-05-09)

### Feat

- initial commit
```

Both bumps succeeded; no `No tag found to do an incremental changelog` error.

**Verdict: FIXED** — closing-eligible. `TagRules` in `commitizen/tags.py` now properly handles custom prefixed `tag_format` via regex matching.

[...]

PR #566, opened 2022-08-22
fix(commands/changelog): Fixed issue #561 cz bump could not find the latest version tag with custom tag_format

ssiuhk:
## Description
fix(command_changelog): Fixed issue #561 cz bump could not find the latest version tag with custom tag_format
- Updated args to _find_incremental_rev to use latest_tag_version instead of latest_version
- Updated test case for test_changelog_command.py from .next to .rc as python packaging lib doesn't take 'next' as a valid version

## Checklist

- [ ] Add test cases to all the changes you introduce
- [X] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [X] Test the changes on the local machine manually
- [ ] Update the documentation for the changes

## Expected behavior
The changelog command will now use the tag_format_$version to lookup the git tags instead of just $version, which may fail if git tag has a long length which significally reduce the similarity

## Steps to Test This Pull Request
Configuration
```
$ cat .cz.toml | grep tag_format
tag_format = "SOME_TAG_PREFIX_$version"
```

Commits
```
$ git tag
SOME_TAG_PREFIX_1.0.0

$ git log
commit 11ba99fd2907be046e814662e3057c4bdd9f3db0 (HEAD -> master)
Author: Sam SIU <23556929+ssiuhk@users.noreply.github.com>
Date:   Fri Aug 19 19:35:08 2022 +0000

    feat(Testing): Testing something

commit db20a16f6ed45ef1470b131f65c9006b094c4a58 (tag: SOME_TAG_PREFIX_1.0.0)
Author: Sam SIU <23556929+ssiuhk@users.noreply.github.com>
Date:   Fri Aug 19 18:54:18 2022 +0000

    First release
```

Command run
```
$ cz bump --dry-run --changelog
bump: version 1.0.0 → 1.1.0
tag to create: SOME_TAG_PREFIX_1.1.0
increment detected: MINOR

## SOME_TAG_PREFIX_1.1.0 (2022-08-19)

### Feat

- **Testing**: Testing something
```

PR #995, opened 2024-02-28
fix(changelog): handle custom tag_format in changelog generation

grahamhar:
When the tag_format does not follow the allowed schemas patterns then changlog generation fails.

## Description

I am working in a mono repo with multiple `.cz.toml` configs (one per component) using `tag_format` to create tags for each component so they can be released independent of each other. When trying to generate the changelog for each component errors are generated see #845.

I was unsure how to add a test case for this if you have ideas please let me know and I will be happy to add.

## Checklist

## Expected behavior

changelogs will now be generated correctly when running `cz bump --changelog`

### DIFF
--- a/commitizen/changelog_formats/asciidoc.py
+++ b/commitizen/changelog_formats/asciidoc.py
@@ -27,9 +27,9 @@ class AsciiDoc(BaseFormat):
             return None
 
         if partial_matches.get("prerelease"):
-            partial_version += f"-{partial_matches['prerelease']}"
+            partial_version = f"{partial_version}-{partial_matches['prerelease']}"
         if partial_matches.get("devrelease"):
-            partial_version += f"{partial_matches['devrelease']}"
+            partial_version = f"{partial_version}{partial_matches['devrelease']}"
         return partial_version
 
     def parse_title_level(self, line: str) -> int | None:
--- a/commitizen/changelog_formats/asciidoc.py
+++ b/commitizen/changelog_formats/asciidoc.py
@@ -18,7 +18,19 @@ class AsciiDoc(BaseFormat):
         matches = list(re.finditer(self.version_parser, m.group("title")))
         if not matches:
             return None
-        return matches[-1].group("version")
+        if "version" in matches[-1].groupdict():
+            return matches[-1].group("version")
+        partial_matches = matches[-1].groupdict()
+        try:
+            partial_version = f"{partial_matches['major']}.{partial_matches['minor']}.{partial_matches['patch']}"
+        except KeyError:
+            return None
+
+        if partial_matches.get("prerelease"):
+            partial_version += f"-{partial_matches['prerelease']}"
+        if partial_matches.get("devrelease"):
+            partial_version += f"{partial_matches['devrelease']}"
+        return partial_version
 
     def parse_title_level(self, line: str) -> int | None:
         m = self.RE_TITLE.match(line)
--- a/commitizen/changelog_formats/base.py
+++ b/commitizen/changelog_formats/base.py
@@ -8,8 +8,8 @@ from typing import IO, Any, ClassVar
 
 from commitizen.changelog import Metadata
 from commitizen.config.base_config import BaseConfig
-from commitizen.version_schemes import get_version_scheme
 from commitizen.defaults import get_tag_regexes
+from commitizen.version_schemes import get_version_scheme
 
 from . import ChangelogFormat
 
--- a/commitizen/changelog_formats/base.py
+++ b/commitizen/changelog_formats/base.py
@@ -9,6 +9,7 @@ from typing import IO, Any, ClassVar
 from commitizen.changelog import Metadata
 from commitizen.config.base_config import BaseConfig
 from commitizen.version_schemes import get_version_scheme
+from commitizen.defaults import get_tag_regexes
 
 from . import ChangelogFormat
 
@@ -32,20 +33,7 @@ class BaseFormat(ChangelogFormat, metaclass=ABCMeta):
     def version_parser(self) -> Pattern:
         tag_regex: str = self.tag_format
         version_regex = get_version_scheme(self.config).parser.pattern
-        TAG_FORMAT_REGEXS = {
-            "$version": version_regex,
-            "$major": r"(?P<major>\d+)",
-            "$minor": r"(?P<minor>\d+)",
-            "$patch": r"(?P<patch>\d+)",
-            "$prerelease": r"(?P<prerelease>\w+\d+)?",
-            "$devrelease": r"(?P<devrelease>\.dev\d+)?",
-            "${version}": version_regex,
-            "${major}": r"(?P<major>\d+)",
-            "${minor}": r"(?P<minor>\d+)",
-            "${patch}": r"(?P<patch>\d+)",
-            "${prerelease}": r"(?P<prerelease>\w+\d+)?",
-            "${devrelease}": r"(?P<devrelease>\.dev\d+)?",
-        }
+        TAG_FORMAT_REGEXS = get_tag_regexes(version_regex)
         for pattern, regex in TAG_FORMAT_REGEXS.items():
             tag_regex = tag_regex.replace(pattern, regex)
         return re.compile(tag_regex)
--- a/commitizen/changelog_formats/base.py
+++ b/commitizen/changelog_formats/base.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import re
 from abc import ABCMeta
 from re import Pattern
 from typing import IO, Any, ClassVar
@@ -25,30 +26,29 @@ class BaseFormat(ChangelogFormat, metaclass=ABCMeta):
         # See: https://bugs.python.org/issue44807
         self.config = config
         self.encoding = self.config.settings["encoding"]
-        self.tag_format = self.config.settings.get("tag_format")
+        self.tag_format = self.config.settings["tag_format"]
 
     @property
     def version_parser(self) -> Pattern:
+        tag_regex: str = self.tag_format
         version_regex = get_version_scheme(self.config).parser.pattern
-        if self.tag_format != "$version":
-            TAG_FORMAT_REGEXS = {
-                "$version": version_regex,
-                "$major": "(?P<major>\d+)",
-                "$minor": "(?P<minor>\d+)",
-                "$patch": "(?P<patch>\d+)",
-                "$prerelease": "(?P<prerelease>\w+\d+)?",
-                "$devrelease": "(?P<devrelease>\.dev\d+)?",
-                "${version}": version_regex,
-                "${major}": "(?P<major>\d+)",
-                "${minor}": "(?P<minor>\d+)",
-                "${patch}": "(?P<patch>\d+)",
-                "${prerelease}": "(?P<prerelease>\w+\d+)?",
-                "${devrelease}": "(?P<devrelease>\.dev\d+)?",
-            }
-            version_regex = self.tag_format
-            for pattern, regex in TAG_FORMAT_REGEXS.items():
-                version_regex = version_regex.replace(pattern, regex)
-        return rf"{version_regex}"
+        TAG_FORMAT_REGEXS = {
+            "$version": version_regex,
+            "$major": r"(?P<major>\d+)",
+            "$minor": r"(?P<minor>\d+)",
+            "$patch": r"(?P<patch>\d+)",
+            "$prerelease": r"(?P<prerelease>\w+\d+)?",
+            "$devrelease": r"(?P<devrelease>\.dev\d+)?",
+            "${version}": version_regex,
+            "${major}": r"(?P<major>\d+)",
+            "${minor}": r"(?P<minor>\d+)",
+            "${patch}": r"(?P<patch>\d+)",
+            "${prerelease}": r"(?P<prerelease>\w+\d+)?",
+            "${devrelease}": r"(?P<devrelease>\.dev\d+)?",
+        }
+        for pattern, regex in TAG_FORMAT_REGEXS.items():
+            tag_regex = tag_regex.replace(pattern, regex)
+        return re.compile(tag_regex)
 
     def get_metadata(self, filepath: str) -> Metadata:
         if not os.path.isfile(filepath):
--- a/commitizen/changelog_formats/base.py
+++ b/commitizen/changelog_formats/base.py
@@ -25,10 +25,30 @@ class BaseFormat(ChangelogFormat, metaclass=ABCMeta):
         # See: https://bugs.python.org/issue44807
         self.config = config
         self.encoding = self.config.settings["encoding"]
+        self.tag_format = self.config.settings.get("tag_format")
 
     @property
     def version_parser(self) -> Pattern:
-        return get_version_scheme(self.config).parser
+        version_regex = get_version_scheme(self.config).parser.pattern
+        if self.tag_format != "$version":
+            TAG_FORMAT_REGEXS = {
+                "$version": version_regex,
+                "$major": "(?P<major>\d+)",
+                "$minor": "(?P<minor>\d+)",
+                "$patch": "(?P<patch>\d+)",
+                "$prerelease": "(?P<prerelease>\w+\d+)?",
+                "$devrelease": "(?P<devrelease>\.dev\d+)?",
+                "${version}": version_regex,
+                "${major}": "(?P<major>\d+)",
+                "${minor}": "(?P<minor>\d+)",
+                "${patch}": "(?P<patch>\d+)",
+                "${prerelease}": "(?P<prerelease>\w+\d+)?",
+                "${devrelease}": "(?P<devrelease>\.dev\d+)?",
+            }
+            version_regex = self.tag_format
+            for pattern, regex in TAG_FORMAT_REGEXS.items():
+                version_regex = version_regex.replace(pattern, regex)
+        return rf"{version_regex}"
 
     def get_metadata(self, filepath: str) -> Metadata:
         if not os.path.isfile(filepath):
--- a/commitizen/changelog_formats/markdown.py
+++ b/commitizen/changelog_formats/markdown.py
@@ -30,9 +30,9 @@ class Markdown(BaseFormat):
             return None
 
         if matches.get("prerelease"):
-            partial_version += f"-{matches['prerelease']}"
+            partial_version = f"{partial_version}-{matches['prerelease']}"
         if matches.get("devrelease"):
-            partial_version += f"{matches['devrelease']}"
+            partial_version = f"{partial_version}{matches['devrelease']}"
         return partial_version
 
     def parse_title_level(self, line: str) -> int | None:
--- a/commitizen/changelog_formats/markdown.py
+++ b/commitizen/changelog_formats/markdown.py
@@ -19,7 +19,21 @@ class Markdown(BaseFormat):
         m = re.search(self.version_parser, m.group("title"))
         if not m:
             return None
-        return m.group("version")
+        if "version" in m.groupdict():
+            return m.group("version")
+        matches = m.groupdict()
+        try:
+            partial_version = (
+                f"{matches['major']}.{matches['minor']}.{matches['patch']}"
+            )
+        except KeyError:
+            return None
+
+        if matches.get("prerelease"):
+            partial_version += f"-{matches['prerelease']}"
+        if matches.get("devrelease"):
+            partial_version += f"{matches['devrelease']}"
+        return partial_version
 
     def parse_title_level(self, line: str) -> int | None:
         m = self.RE_TITLE.match(line)
--- a/commitizen/changelog_formats/restructuredtext.py
+++ b/commitizen/changelog_formats/restructuredtext.py
@@ -46,7 +46,6 @@ class RestructuredText(BaseFormat):
             third = third.strip().lower()
             title: str | None = None
             kind: TitleKind | None = None
-
             if self.is_overlined_title(first, second, third):
                 title = second
                 kind = (first[0], third[0])
@@ -67,10 +66,25 @@ class RestructuredText(BaseFormat):
                 # Try to find the latest release done
                 m = re.search(self.version_parser, title)
                 if m:
-                    version = m.group("version")
-                    meta.latest_version = version
-                    meta.latest_version_position = index
-                    break  # there's no need for more info
+                    matches = m.groupdict()
+                    if "version" in matches:
+                        version = m.group("version")
+                        meta.latest_version = version
+                        meta.latest_version_position = index
+                        break  # there's no need for more info
+                    try:
+                        partial_version = (
+                            f"{matches['major']}.{matches['minor']}.{matches['patch']}"
+                        )
+                        if matches.get("prerelease"):
+                            partial_version += f"-{matches['prerelease']}"
+                        if matches.get("devrelease"):
+                            partial_version += f"{matches['devrelease']}"
+                        meta.latest_version = partial_version
+                        meta.latest_version_position = index
+                        break
+                    except KeyError:
+                        pass
         if meta.unreleased_start is not None and meta.unreleased_end is None:
             meta.unreleased_end = (
                 meta.latest_version_position if meta.latest_version else index + 1
--- a/commitizen/changelog_formats/restructuredtext.py
+++ b/commitizen/changelog_formats/restructuredtext.py
@@ -77,9 +77,13 @@ class RestructuredText(BaseFormat):
                             f"{matches['major']}.{matches['minor']}.{matches['patch']}"
                         )
                         if matches.get("prerelease"):
-                            partial_version += f"-{matches['prerelease']}"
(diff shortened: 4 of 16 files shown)


woile:
You can add tests similar to this:
https://github.com/commitizen-tools/commitizen/blob/master/tests/commands/test_changelog_command.py#L1457-L1491
where you can also configure the `toml` file and simulate a user

grahamhar:
I have taken a stab at the test case.

commitizen/commands/changelog.py
```diff
@@ -164,21 +164,17 @@ def __call__(self):
...
-                latest_tag_version: str = bump.normalize_tag(
-                    changelog_meta.latest_version,
-                    tag_format=self.tag_format,
-                    scheme=self.scheme,
-                )
                 start_rev = self._find_incremental_rev(
-                    strip_local_version(latest_tag_version), tags
+                    strip_local_version(changelog_meta.latest_version), tags
                 )
-
```
Lowaiz, line 181: Hello,
Removing this logic is breaking the search for custom tags (with a tag-format like `example-${version}`).
The first bump (with `--changelog` option) will have no problem and it will create the `CHANGELOG.md` file but any bump afterward will result in an error: `commitizen.exceptions.NoRevisionError: No tag found to do an incremental changelog`.

This is due to the `_find_incremental_rev` function that look at similarity between a given tag and the tags list. But without the normalization, it calculate similarity between a stripped tag (without any format so like `0.1.1`) and the formatted tag (like `example-0.1.0`) resulting in a similarity below threshold.

So, except if you have another edge case that was justifying this removal, I think it should be kept.

grahamhar: Hi,

Thanks for the feedback, I'll put this to a draft status and try to find a solution that doesn't break the current way of working. My gut feeling is it might need an additional flag maybe something like `--strict-tag-matching` what are your thoughts on taking that approach?

I guess the fact you found that error might mean there is a missing test case, I will try adding that first to replicate the error with my changes so I have something to validate against.

Lowaiz: I used your version with a revert on the selected part ([here](https://github.com/Lowaiz/commitizen/)), and it seems to works pretty damn well for me, that's why I was asking for the "why" of the change.
I was able to bump and generate for multiple cz.toml file and multiple tag-format (in a monorepo with 3 sub-packages).

The normalizing part was introduced by that [PR](https://github.com/commitizen-tools/commitizen/pull/566/files), exactly fixing the error I got. 

I really think that you just need to keep the normalizing part, and we should be OK.

grahamhar: Hi,

Thanks for clarifying I misunderstood your point.

Lowaiz: Oh I see the problem.

The difference in behavior between `example-${version}` and  `${version}-example` is due to the regex used in the `changelog_format/markdown.py`.
On my example, the regex only match the version number, but with your it parse the whole tag, probably because the suffix `-example` match some condition (like `-alpha` match for a alpha release) in the regex.

My suggestion would be to use the tag_format to build a regex to extract the tag from the line, and then from the extracted tag, use the version regex to extract the version

grahamhar: A few questions/points after thinking further on this before trying to make further changes. 

Tests of the changelog use examples that don't appear to match the format of those written by the commitizen, is this to cover that consumers started using commitizen with an existing changelog?

If I am understanding the logic correctly we pull the version out of the changelog by matching for the tag format in the changelog, then take that version to match against the tags to find the latest tag in the changelog to get the starting revision, again is this to handle changelings that might not have originally been created by commitizen? If not should it be safe to assume that we could just match on tag in the changelog to get the starting revision when doing incremental changelogs?

Lowaiz: Just to clarify why I'm on this PR, I'm not a maintainer of this project but I'm greatly interested by the monorepo feature, so I try to make it happen a bit faster. 

> Tests of the changelog use examples that don't appear to match the format of those written by the commitizen, is this to cover that consumers started using commitizen with an existing changelog?

I think yes, I saw a PR or Issue saying that, can't remember where.


> If I am understanding the logic correctly we pull the version out of the changelog by matching for the tag format in the changelog, then take that version to match against the tags to find the latest tag in the changelog to get the starting revision, again is this to handle changelings that might not have originally been created by commitizen? If not should it be safe to assume that we could just match on tag in the changelog to get the starting revision when doing incremental changelogs?

I can't really tell, for both questions. 
For the logic I think you're right.
But I think I see a flaw in the current logic:
- if the `tag_format` does not contain `$version` or `${version}`, the substitution does not include the regex group `version` so the extraction fails in the changelog metadat fetch.
- Another [ongoing PR](https://github.com/commitizen-tools/commitizen/pull/692) seems to start addressing this flow, but I think it does not cover all uses cases.
- if a `tag_format` does not contains the complete version infos (major, minor, patch, pre-rel, dev-rel), we can't make the back-matching from the tag to the version, ultimately, I really think the complete version should be used to generate changelog and not the formatted tag, or we need to ensure that the format contains all infos, and make a function to build a regex from the tag (a bit like the mentioned PR). Other possibility could be to not base the changelog generation on the available info inside the current changelog. (at least when bumping version, as we can get the lat git tag and the upcoming one, there is no need to base the last version to the version available in the changelog)

I think it's a complicated question for now, maybe the maintainers could help us a bit to lead to a future-proof solution: @woile @Lee-W @noirbizarre, reading the conversation, do you have opinion or an idea on what to do to better handle tag/tag_format and changelogs generation ?

Lee-W: I'm still trying to read through all the discussion, but here're some short answers to the questions

> if the tag_format does not contain $version or ${version}, the substitution does not include the regex group version so the extraction fails in the changelog metadata fetch.

We should catch the case that `tag_format` does not contain a version.

> if a tag_format does not contains the complete version infos (major, minor, patch, pre-rel, dev-rel), we can't make the back-matching from the tag to the version, ultimately, I really think the complete version should be used to generate changelog and not the formatted tag, or we need to ensure that the format contains all infos, and make a function to build a regex from the tag (a bit like the mentioned PR). Other possibility could be to not base the changelog generation on the available info inside the current changelog. (at least when bumping version, as we can get the last git tag and the upcoming one, there is no need to base the last version to the version available in the changelog)

Yes, I think `we need to ensure that the format contains all infos, and make a function to build a regex from the tag ` is what we should do.

Lee-W: Not sure whether this would be a breaking change 🤔 if that's the case, `--strict-tag-matching`  is something we could try.

grahamhar: We might not need to do a breaking change for this (the functionality was broken already).

I have added handling for different tag_formats other than the standard `$version` to the markdown format. If someone could take a look and let me know if this is a valid approach I will implement the same for all the format types.

I have also taken a stab at documenting how to work with commitizen in a monorepo.

Lee-W: Is there anything that needs clarification for this one? (I guess not?) or should I start reviewing?

grahamhar: I just want to check the way I have handled the different possible tag_formats in markdown.py seems OK. If it is I will make the appropriate changes to the other formats taking the same approach. Once I have completed that it will be ready for review

Lee-W: I don't see any major flaw at the first glance but will try to take a deeper look this weekend.

grahamhar: Sorry it took a while for me to get this finished off. Ready for review when you have time.

Lee-W: Thanks! I was not able to review it last week. 😞  Let's see whether I can at least check a portion of it this week. 💪

commitizen/changelog_formats/base.py
```diff
@@ -24,10 +25,29 @@ def __init__(self, config: BaseConfig):
...
+            "$patch": r"(?P<patch>\d+)",
+            "$prerelease": r"(?P<prerelease>\w+\d+)?",
+            "$devrelease": r"(?P<devrelease>\.dev\d+)?",
+            "${version}": version_regex,
+            "${major}": r"(?P<major>\d+)",
+            "${minor}": r"(?P<minor>\d+)",
+            "${patch}": r"(?P<patch>\d+)",
+            "${prerelease}": r"(?P<prerelease>\w+\d+)?",
+            "${devrelease}": r"(?P<devrelease>\.dev\d+)?",
+        }
```
noirbizarre, line 47: Those are exactly the same as in `changelog:get_version_tag` so I suggest to factorize this into a public factory, something like:
```python
TAG_FORMAT_REGEXS = {
    "$major": r"(?P<major>\d+)",
    "$minor": r"(?P<minor>\d+)",
    "$patch": r"(?P<patch>\d+)",
    "$prerelease": r"(?P<prerelease>\w+\d+)?",
    "$devrelease": r"(?P<devrelease>\.dev\d+)?",
    "${version}": version_regex,
    "${major}": r"(?P<major>\d+)",
    "${minor}": r"(?P<minor>\d+)",
    "${patch}": r"(?P<patch>\d+)",
    "${prerelease}": r"(?P<prerelease>\w+\d+)?",
    "${devrelease}": r"(?P<devrelease>\.dev\d+)?",
}

def tag_format_regexps_for(version: Pattern) -> dict[str, Pattern]:
    return {
        "$version": version,
		"${version}": version,
		**TAG_FORMAT_REGEXS
    }
```

This way:
- single source of thrust
- custom format implementations can reuse `TAG_FORMAT_REGEXS` and `tag_format_regexps_for`

commitizen/providers/scm_provider.py
```diff
@@ -31,6 +31,12 @@ class ScmProvider(VersionProvider):
         "$patch": r"(?P<patch>\d+)",
         "$prerelease": r"(?P<prerelease>\w+\d+)?",
         "$devrelease": r"(?P<devrelease>\.dev\d+)?",
+        "${version}": r"(?P<version>.+)",
+        "${major}": r"(?P<major>\d+)",
+        "${minor}": r"(?P<minor>\d+)",
+        "${patch}": r"(?P<patch>\d+)",
+        "${prerelease}": r"(?P<prerelease>\w+\d+)?",
+        "${devrelease}": r"(?P<devrelease>\.dev\d+)?",
```
noirbizarre, line 39: This is a new accepted syntax.
Documentation and conventional commit message should reflect that

noirbizarre, line 39: Very similar to the 2 previous factorizable `TAG_FORMAT_REGEXS` so I think it should be factorized too,

noirbizarre (changes requested):
Seems good but:
- `TAG_FORMAT_REGEXS` should be factorized and declared once
- the new `${}` syntax should be documented

grahamhar:
> Seems good but:
> 
> - `TAG_FORMAT_REGEXS` should be factorized and declared once
> 
> - the new `${}` syntax should be documented

I think I have implemented the suggestions

commitizen/changelog_formats/asciidoc.py
```diff
@@ -18,7 +18,19 @@ def parse_version_from_title(self, line: str) -> str | None:
...
+        if "version" in matches[-1].groupdict():
+            return matches[-1].group("version")
+        partial_matches = matches[-1].groupdict()
+        try:
+            partial_version = f"{partial_matches['major']}.{partial_matches['minor']}.{partial_matches['patch']}"
+        except KeyError:
+            return None
+
+        if partial_matches.get("prerelease"):
+            partial_version += f"-{partial_matches['prerelease']}"
```
Lee-W, line 30:
```suggestion
            partial_version = f"{partial_version}-{partial_matches['prerelease']}"
```

commitizen/changelog_formats/asciidoc.py
```diff
@@ -18,7 +18,19 @@ def parse_version_from_title(self, line: str) -> str | None:
...
+        if partial_matches.get("devrelease"):
+            partial_version += f"{partial_matches['devrelease']}"
```
Lee-W, line 32:
```suggestion
            partial_version = f"{partial_version}{partial_matches['devrelease']}"
```

commitizen/changelog_formats/markdown.py
```diff
@@ -19,7 +19,21 @@ def parse_version_from_title(self, line: str) -> str | None:
...
+        matches = m.groupdict()
+        try:
+            partial_version = (
+                f"{matches['major']}.{matches['minor']}.{matches['patch']}"
+            )
+        except KeyError:
+            return None
+
+        if matches.get("prerelease"):
+            partial_version += f"-{matches['prerelease']}"
```
Lee-W, line 33:
```suggestion
            partial_version = f"{partial_version}-{matches['prerelease']}"
```

commitizen/changelog_formats/markdown.py
```diff
@@ -19,7 +19,21 @@ def parse_version_from_title(self, line: str) -> str | None:
...
+        if matches.get("devrelease"):
+            partial_version += f"{matches['devrelease']}"
```
Lee-W, line 35:
```suggestion
            partial_version = f"{partial_version}{matches['devrelease']}"
```

commitizen/providers/scm_provider.py
```diff
@@ -24,14 +25,7 @@ class ScmProvider(VersionProvider):
...
 
-    TAG_FORMAT_REGEXS = {
-        "$version": r"(?P<version>.+)",
-        "$major": r"(?P<major>\d+)",
-        "$minor": r"(?P<minor>\d+)",
-        "$patch": r"(?P<patch>\d+)",
-        "$prerelease": r"(?P<prerelease>\w+\d+)?",
-        "$devrelease": r"(?P<devrelease>\.dev\d+)?",
-    }
+    TAG_FORMAT_REGEXS = get_tag_regexes(r"(?P<version>.+)")
```
Lee-W, line 28: What will this match? It lookw like this key does not exists in the dict in `get_tag_regexes`

commitizen/changelog_formats/restructuredtext.py
```diff
@@ -67,10 +66,25 @@ def get_metadata_from_file(self, file: IO[Any]) -> Metadata:
...
+                        version = m.group("version")
+                        meta.latest_version = version
+                        meta.latest_version_position = index
+                        break  # there's no need for more info
+                    try:
+                        partial_version = (
+                            f"{matches['major']}.{matches['minor']}.{matches['patch']}"
+                        )
+                        if matches.get("prerelease"):
+                            partial_version += f"-{matches['prerelease']}"
```
Lee-W, line 80:
```suggestion
                            partial_version = "{partial_version}-{matches['prerelease']}"
```

commitizen/changelog_formats/restructuredtext.py
```diff
@@ -67,10 +66,25 @@ def get_metadata_from_file(self, file: IO[Any]) -> Metadata:
...
+                        if matches.get("devrelease"):
+                            partial_version += f"{matches['devrelease']}"
```
Lee-W, line 82:
```suggestion
                            partial_version = "{partial_version}{matches['devrelease']}"
```

commitizen/changelog_formats/restructuredtext.py
```diff
@@ -67,10 +66,25 @@ def get_metadata_from_file(self, file: IO[Any]) -> Metadata:
...
+                        meta.latest_version = partial_version
+                        meta.latest_version_position = index
+                        break
+                    except KeyError:
```
Lee-W, line 86: This is try-catch block that catches a bit too broad. On which line do we expect the exception to happened? could we add a few simple comparison to reduce this block?

commitizen/changelog_formats/textile.py
```diff
@@ -16,7 +16,21 @@ def parse_version_from_title(self, line: str) -> str | None:
...
-        return m.group("version")
+        if "version" in m.groupdict():
+            return m.group("version")
+        matches = m.groupdict()
+        try:
+            partial_version = (
+                f"{matches['major']}.{matches['minor']}.{matches['patch']}"
+            )
+        except KeyError:
+            return None
```
Lee-W, line 27:
```suggestion
        if not all([version_segment in matches for version_segment in ("major", "minor", "patch")]):
            return None
            
        partial_version = (
            f"{matches['major']}.{matches['minor']}.{matches['patch']}"
        )
```

commitizen/changelog_formats/textile.py
```diff
@@ -16,7 +16,21 @@ def parse_version_from_title(self, line: str) -> str | None:
...
+
+        if matches.get("prerelease"):
+            partial_version += f"-{matches['prerelease']}"
```
Lee-W, line 30:
```suggestion
            partial_version = "{partial_version}-{matches['prerelease']}"
```

commitizen/changelog_formats/textile.py
```diff
@@ -16,7 +16,21 @@ def parse_version_from_title(self, line: str) -> str | None:
...
+        if matches.get("devrelease"):
+            partial_version += f"{matches['devrelease']}"
```
Lee-W, line 32:
```suggestion
            partial_version = "{partial_version}{matches['devrelease']}"
```

commitizen/defaults.py
```diff
@@ -133,3 +133,20 @@ class Settings(TypedDict, total=False):
 )
 change_type_order = ["BREAKING CHANGE", "Feat", "Fix", "Refactor", "Perf"]
 bump_message = "bump: version $current_version → $new_version"
+
+
+def get_tag_regexes(version_regex: str) -> dict[str | Any, str | Any]:
```
Lee-W, line 138: When will the key return `Any`?

commitizen/providers/scm_provider.py
grahamhar: This is the regex being passed in to be used for $version or ${version} in the return dictionary of part names to regex mappings. In the refactor to bring in a single definition of the TAG_FORMAT_REGEXS I needed to accept this as an input due to in this one situation the regex for it here already being that value, I am unsure why we have that one here and couldn't work it out so just replicated the current behaviour.

commitizen/defaults.py
grahamhar: Fixed

tests/commands/test_changelog_command.py
```diff
@@ -1522,6 +1522,120 @@ def test_changelog_template_extras_precedance(
...
+    with open(config_path, "a", encoding="utf-8") as f:
+        f.write('\ntag_format = "${version}custom"\n')
+    create_file_and_commit("feat: new file")
+    git.tag("v0.2.0")
+    create_file_and_commit("feat: another new file")
+    git.tag("0.2.0")
+    git.tag("random0.2.0")
+    testargs = ["cz", "bump", "--changelog", "--yes"]
+    mocker.patch.object(sys, "argv", testargs)
+    cli.main()
```
Lee-W, line 1599:
```suggestion
    # bump to 0.2.0custom
    cli.main()
```

tests/commands/test_changelog_command.py
```diff
@@ -1522,6 +1522,120 @@ def test_changelog_template_extras_precedance(
...
+    wait_for_tag()
+    create_file_and_commit("feat: another new file")
+    cli.main()
```
Lee-W, line 1602:
```suggestion
    # bump to 0.3.0custom
    cli.main()
```

tests/commands/test_changelog_command.py
Lee-W: Read a few times to understand fully. would be great if we can have comments here 🙂

tests/test_changelog_format_asciidoc.py
```diff
@@ -136,3 +166,34 @@ def test_get_matadata(
     changelog.write_text(content)
 
     assert format.get_metadata(str(changelog)) == expected
+
+
+@pytest.mark.parametrize(
+    "format_with_tags, tag_string, expected, ",
```
Lee-W, line 172:
```suggestion
    "format_with_tags, tag_string, expected",
```

tests/test_changelog_format_markdown.py
```diff
@@ -136,3 +166,40 @@ def test_get_matadata(
...
+        pytest.param(
+            "${major}-${minor}-${patch}-${prerelease}-example",
+            "1-0-0-a1-example",
+            "1.0.0-a1",
+        ),
+        pytest.param(
+            "${major}-${minor}-${patch}-${prerelease}${devrelease}-example",
+            "1-0-0-a1.dev1-example",
+            "1.0.0-a1.dev1",
+        ),
```
Lee-W, line 194: feel like we can make it reuseable somewhere 🤔

Lee-W (commented):
Sorry for taking so long to review. Most of my comments are just nitpicks. It would be great if we could fix them, but I'm good with it if we cannot. As this is somewhat a larger change, I think it would be better if we can have @woile and @noirbizarre take a look as well 🙂

Lee-W:
We probably need to resolve the conflict as well 👀 . We're now using screenshots as part of our doc. you can generate the latest screenshot through `poetry run python scripts/gen_cli_help_screenshots.py`

docs/tutorials/monorepo_guidance.md
```diff
@@ -0,0 +1,40 @@
...
+
+```
+src
+    library-b
+      .cz.toml
+    library-z
+      .cz.toml
+```
+
+Each component will have its own changelog, commits will need to use scopes so only relevant commits are included in the
```
woile, line 21: What if you are not using conventional commits and you don't have scopes?

grahamhar: Sorry, I guess I made an assumption here that everyone uses them. Should I just reword this to reflect that assumption?

grahamhar:
> Sorry for taking so long to review. Most of my comments are just nitpicks. It would be great if we could fix them, but I'm good with it if we cannot. As this is somewhat a larger change, I think it would be better if we can have @woile and @noirbizarre take a look as well 🙂 

No worries on the time taken, I will address all the points but might take me a week due to other commitments

Lee-W:
> When attempting to run the `gen_cli_help_screenshots.py` script requested above I just get this output which appears to just be given usage errors of `cz`

The screenshot doesn't look like an error to me 🤔 so basically, this is a command we use to generate the latest cz help message. so if the screenshots are generated, I think we are good 🙂 

> I'm not sure if you are already aware, but with the push hook in place I got an error for a commit not in my PR

Even if you try with rebase? There was one commit with an ill-formatted message. But our hook should only check the commits in this branch, I think 🤔 It Looks like the CI is passing now. Is this still an issue?

grahamhar:
> > When attempting to run the `gen_cli_help_screenshots.py` script requested above I just get this output which appears to just be given usage errors of `cz`
> 
> The screenshot doesn't look like an error to me 🤔 so basically, this is a command we use to generate the latest cz help message. so if the screenshots are generated, I think we are good 🙂
> 
> > I'm not sure if you are already aware, but with the push hook in place I got an error for a commit not in my PR
> 
> Even if you try with rebase? There was one commit with an ill-formatted message. But our hook should only check the commits in this branch, I think 🤔 It Looks like the CI is passing now. Is this still an issue?

Maybe my PR didn't need any changes to the generated images, possibly just confusion on my side.

I have just made sure my fork is updated and as I have no changes I just tried running the hook

```
 pre-commit run --hook-stage pre-push                      
Check hooks apply to the repository.......................(no files to check)Skipped
Check for useless excludes................................(no files to check)Skipped
check vcs permalinks......................................(no files to check)Skipped
fix end of files..........................................(no files to check)Skipped
trim trailing whitespace..................................(no files to check)Skipped
debug statements (python).................................(no files to check)Skipped
don't commit to branch........................................................Passed
check for merge conflicts.................................(no files to check)Skipped
check toml................................................(no files to check)Skipped
check yaml................................................(no files to check)Skipped
detect private key........................................(no files to check)Skipped
blacken-docs..............................................(no files to check)Skipped
Run codespell to check for common misspellings in files...(no files to check)Skipped
commitizen check branch.......................................................Failed
- hook id: commitizen-branch
- exit code: 23

fatal: ambiguous argument 'origin/HEAD..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'



format....................................................(no files to check)Skipped
linter and test...........................................(no files to check)Skipped

```

Doing some further investigation my fork doesn't have origin/HEAD set. After running

`git remote set-head origin master`

The hook now works :)

Lee-W:
> The hook now works :)

> The hook now works :)

Niiiiiiice! I will probably not be around for half a month, but I will try to take a look after I'm back from traveling. Thanks for actively helping out.

noirbizarre:
Sorry for the looong absence.
I'll take the time to do a final review by the end of the week (as this one is a big one, and I need to reread and remember all what have been said)

woile:
This could potentially fix #1149 right?

woile (approved):
LGTM @Lee-W @noirbizarre any thoughts?

woile:
I'm planning on merging this tomorrow

Lee-W:
I'll be almost out till mid-Oct 😞 @woile Thanks for taking care of this!

PR #996, opened 2024-02-29
fix(commands/bump): prevent using incremental changelog when it is set to false in config

### LINKED ISSUE #883: changelog_incremental is set as default when calling changelog from cz bump
### Description

`cz bump` with `changelog_incremental = false` on `pyproject.toml` doesn't have any effect when calling `cz bump --changelog` or `cz bump with `update_changelog_on_bump = true`.

### Steps to reproduce

1. Define changelog_incremental = false and update_changelog_on_bump = true
2. Run `cz bump`

### Current behavior

It runs as if `changelog_incremental = true`.

### Desired behavior

Run `cz ch` without `incremental`.

### Screenshots

image
https://github.com/commitizen-tools/commitizen/blob/9d0878c201c579aaf7d7c0219f19f95897bd01fc/commitizen/commands/bump.py#L272C1-L279C14

### Environment

```bash
Commitizen Version: 3.10.0
Python Version: 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)]
Operating System: Windows
```

josix:
Fix #883 

## Description

Since the default behavior of `cz bump --changelog` is to write incremental changelog, which is different from the default value of `changelog_incremental`, it is hard to know the value is set from user or the default setting, To address it, I created one more member in `BaseConfig`  to record which property is defined from users so that we could distinguish the value is from default setting or users.

## Checklist

## Expected behavior
`cz bump --changelog` should work as `false` when user configure `changelog_incremental` to `false`.

1. Modify the CHANGELOG file
image

2. Configure `pyproject.toml` with `changelog_incremental = false`
3. Run `cz bump --changelog`
4. Check the CHANGELOG, which the new added content should be replaced
image

## Additional context
ditto.


commitizen/config/base_config.py
```diff
@@ -10,11 +10,16 @@ def __init__(self):
...
         self.encoding = self.settings["encoding"]
         self._path: Path | None = None
+        self._settings_from_configs: Settings = {}
 
     @property
     def settings(self) -> Settings:
         return self._settings
 
+    @property
+    def mutated_settings(self) -> Settings:
```
woile, line 20: Could you add a docstring to this function explaining it a bit further? thanks
Right now it seems it does the same as `mutated_settings`, but I think the *intention* is different, right?

noirbizarre:
Just opening the debate: wouldn't it be easier and more consistent to fix/align the default `bump` behavior to the default settings (even if this is a breaking change) ?

By adding a fourth settings source of trust (defaults, user settings, cli flags and now mutated settings, fifth if you add the fact that plugins can override some settings), I feel that we are fixing a symptom but making the problem more complex.

josix:
> Just opening the debate: wouldn't it be easier and more consistent to fix/align the default `bump` behavior to the default settings (even if this is a breaking change) ?
> 
> By adding a fourth settings source of trust (defaults, user settings, cli flags and now mutated settings, fifth if you add the fact that plugins can override some settings), I feel that we are fixing a symptom but making the problem more complex.

Yeah, I think that would be a better way to address the issue if possible. The implementation would be easier and also could prevent overriding the behaviors when running the `bump --changelog`.

woile:
> Just opening the debate: wouldn't it be easier and more consistent to fix/align the default bump behavior to the default settings (even if this is a breaking change) ?

Could you make a proposal for this? It sounds interesting, would be nice if we can simplify the settings

Lee-W:
I'm a bit confused here. I thought we could solve it by reading the value from config?

josix:
> I thought we could solve it by reading the value from config?

I think there are total three combinations of configurations here:

1. user didn't specify `changelog_incremental`, the Config would store the value as `False` and the value passing into `Changelog` should be `True`
2. user did specify `changelog_increamental: true`, the Config would store the value as `True`, so the value passing into `Changelog` would also be `True`
3. user did specify `changelog_increamental: false`, the Config would store the value as `False`, so the value passing into `Changelog` would also be `False`

The 1st & 3rd cases would cause ambiguity when we're reading the value from `Config`, we couldn't tell if the False value is set from the default value or user configuration, according to different source of the value we would take different value to pass into the initialization of `Changelog`.  To address this, I introduce one more property that trace the value set from users to help us differentiate them in this PR.

Lee-W:
> we couldn't tell if the False value is set from the default value or user configuration,

I though the one from config should overwrite the default? Or did I miss anything?

Lowaiz:
Hello,

Do you guys have any news on this fix ?
This setting is blocking the `changelog_merge_prerelease` option, as the arg `incremental` take the precedence, causing us to split the bump and the changelog generation in order to have proper release changelog with every changes from release-x to release-x+1 concatenated, and not with every intermediate pre-release.

Lee-W (commented):
I probably won't say I love this idea, but it is an acceptable workaround before we introduce a better config system. @josix, could you please rebase it and add some tests to verify it?

bearomorphism:
@josix what is the status of this PR? Thanks!

josix:
Oh I missed this comment, let me pick this up

bearomorphism:
Thanks!

commitizen/commands/bump.py
```diff
@@ -310,11 +310,14 @@ def __call__(self) -> None:
...
         if self.changelog_flag:
+            incremental_setting = self.config.settings.get("changelog_incremental")
             changelog_args = {
                 "unreleased_version": new_tag_version,
                 "template": self.template,
                 "extras": self.extras,
-                "incremental": True,
+                "incremental": incremental_setting
+                if incremental_setting is not None
+                else True,
```
bearomorphism, line 323: This probably also works and is less confusing

```suggestion
                "incremental": self.config.settings.get("changelog_incremental", True),
```

josix: yeah, thanks for pointing it out.

josix: weird, seems that it will let tests fail, let me investigate it.

josix: Oh, I found the root cause, `dict.get` only returns the fallback when the key is missing, so `.get(..., True)` would return None (not True). The explicit check is required to default to incremental changelog generation while still honoring an explicit `False`.

Lee-W: It does remind me that Airflow has an `ARG_NOT_SET` thing. not sure whether we should have similiar mechanism

Lee-W: I think we'll also need more context in this comment. same as the value in `default.py`. This is probably the best we can do for the time being, but worth thinking of a abetter solution in the following major versions. also would be great if we have a unit test to cover it

Issue #572, opened 2022-08-29
Cannot commit using tmp file if computer name has a space in it


schmi-schmi:
### Description

When cz c is used, the questions appear as they should. Only when all questions are answered does the same error appear:
`error: pathspec 'Micks\AppData\Local\Temp\tmpsxrs5lc0' did not match any file(s) known to git`

I discovered within Commitizen a tmp file is being generated to save the commit message is a converted format so that it can be committed using the contents of the file.

I found the bug on my side so I can suggest a solution. In the git.py:95 file, the code looked as follows:
```
def commit(message: str, args: str = "") -> cmd.Command:
    f = NamedTemporaryFile("wb", delete=False)
    f.write(message.encode("utf-8"))
    f.close()
    c = cmd.run(f"git commit {args} -F {f.name}")
    os.unlink(f.name)
    return c
```
This should be changed to:
```
def commit(message: str, args: str = "") -> cmd.Command:
    f = NamedTemporaryFile("wb", delete=False)
    f.write(message.encode("utf-8"))
    f.close()
    c = cmd.run(f'git commit {args} -F "{f.name}"')
    os.unlink(f.name)
    return c
```

The error stopped appearing and I was able to complete the commit successfully. Hope this helps. :)

### Steps to reproduce

> cz c

Note: the computer name must include a space.

### Current behavior

Produces a pathspec error ( common in git when a file cannot be found but is being used to commit )

### Desired behavior

cz c should complete without the error appearing.

### Screenshots

I printed out the full pathspec as it is stored after creating the tmp file. The actual pathspec is above, while the error shows what is being used for the commit.
![image](https://user-images.githubusercontent.com/68070794/187147208-a8325733-3102-48dc-a419-105e30b85ab7.png)

### Environment

- commitizen version: `2.32.2`
- python version: `3.9.10`
- Operating System: `Windows`

Additional environment setup required:
- computer name must include a space

Woile:
Hey thanks for finding this error! if you want feel free to send us a PR with this fix.

Cliying94:
Take this issue.  @Lee-W

Lee-W:
@Cliying94 Sure. Let me assign it to you. Thanks for your help!

Lee-W:
Closed by: #1039

PR #1039, opened 2024-03-30
Resolve tempfile path spaces issue in git commit function

ttw225:
## Checklist

## Expected behavior

When the file path used in a commit contains spaces, the commit command should execute correctly. This fix addresses the issue where spaces in file names or paths previously caused errors with the git commit command.

1. Create a file with spaces in its name or path using touch "temp file".
2. Add the file to git with git add "temp file".
3. Execute the commit using cz c.

## Additional context

The test included here serves as a straightforward workaround, employing mocker for testing.
@Cliying94 is on the path to crafting a detailed integration test that begins with the commit operation.

Related: #572

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -293,3 +293,32 @@ def test_create_tag_with_message(tmp_commitizen_project):
         assert git.get_tag_message(tag_name) == (
             tag_message if platform.system() != "Windows" else f"'{tag_message}'"
         )
+
+
+@pytest.mark.parametrize(
+    "file_path,expected_cmd",
+    [
+        (
+            "/tmp/temp file",
+            'git commit --signoff -F "/tmp/temp file"',
+        ),  # File contains spaces
+        (
+            "/tmp dir/temp file",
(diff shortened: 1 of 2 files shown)


tests/test_git.py
```diff
@@ -293,3 +293,32 @@ def test_create_tag_with_message(tmp_commitizen_project):
...
+        (
+            "/tmp dir/temp file",
+            'git commit --signoff -F "/tmp dir/temp file"',
+        ),  # Path contains spaces
+        (
+            "/tmp/tempfile",
+            'git commit --signoff -F "/tmp/tempfile"',
+        ),  # Path does not contain spaces
+    ],
+)
```
Lee-W, line 314: nitpick: instead of using commands. [ids](https://docs.pytest.org/en/stable/example/parametrize.html#different-options-for-test-ids) might be a even better solution

ttw225: Agree! ids is a better way to describe the test cases

Lee-W (approved):

Lee-W:
Closes: #572

Issue #760, opened 2023-06-13
"cz init" should list exisitng tag in reverse order 


Lee-W:
### Description

When we run `cz init` in a project with existing tags, when it comes to the question `? Please choose the latest tag:  (Use arrow keys)`. We list the tags in ascending order. However, the latest tag is most likely with the largest number. Thus list the tags in reverse order makes more sense.

Ellie-Yen:
@Lee-W  i'll take the issue

Lee-W:
@Ellie-Yen Sure thing!

PR #1040, opened 2024-03-30
fix(command-init): "cz init" should list exisitng tag in reverse order

Ellie-Yen:
## Description
When we run cz init in a project with existing tags, when it comes to the question ?` Please choose the latest tag:  (Use arrow keys)`. We list the tags in ascending order. However, the latest tag is most likely with the largest number. Thus list the tags in reverse order makes more sense.

Closes: #760 

## Checklist

## Expected behavior
in `Please choose the latest tag: `
when latest tags are ["0.0.1", "0.0.2"],
the choices should be ["0.0.2", "0.0.1"] 

run test cases in
tests/commands/test_init_commands.py

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -186,9 +186,11 @@ class Init:
                 out.error("No Existing Tag. Set tag to v0.0.1")
                 return "0.0.1"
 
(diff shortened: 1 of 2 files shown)


Lee-W (approved):
LGTM! Great work!

PR #1041, opened 2024-03-30
improve test case for verifying doing "git commit" on path with space

Cliying94:
related: #572

## Description
add-a-test-for-handling-blank-with-path-in-git-commit

* mock action:
Whether the path includes blank or not, the git commit action can parse anyway and correctly execute `commit` under commitizen/commitizen/git.py and go through the util function `create_file_and_commit`

## Checklist

## Expected behavior
if the project path is including blank, the program should work.

run `./scripts/test` or run the single text `test_git_commit_command_with_varied_paths_handles_blank` under `tests/test_git.py`

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -4,6 +4,7 @@ import inspect
 import os
 import platform
 import shutil
+import subprocess
 
 import pytest
 from commitizen import cmd, exceptions, git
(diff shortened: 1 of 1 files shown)


Lee-W:
I would suggest changing the scope of the commit message to `git` instead (i.e., `test(git): ....`)

Issue #818, opened 2023-08-20
[Refactor] move cz_conventional_commit defaults out of defaults.py


Lee-W:
### Description

As described in https://github.com/commitizen-tools/commitizen/issues/425#issuecomment-1537112868, many of our values in [commitizen/defaults.py](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/defaults.py) are for cz_conventional_commit only but not necessary for the whole commitizen.

ITingHung:
@Lee-W can I take this?

Lee-W:
@ITingHung Sure!

Lee-W:
Closed by #1042

PR #1042, opened 2024-03-30
move cz_conventional_commit defaults out of project wide default

ITingHung:
## Description

commit_parser in [commitizen/defaults.py](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/defaults.py) is only for cz_conventional_commit rather than all commitizens. This PR is to move commit_parser from commitizen/defaults.py to [conventional_commits.py](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/cz/conventional_commits/conventional_commits.py)

## Checklist

## Expected behavior

commit_parser in commitizen/defaults.py is removed, and will be directly retrieved from ConventionalCommitsCz.

## Additional context

Related to #818


Lee-W (approved):
Nice refactoring. Will merge it once CI passes.

PR #1047, opened 2024-04-01
feat(cli): add config option to specify config file path

### LINKED ISSUE #656: Add option --config to set the pyproject.toml configuration file path.
### Description

I was not able to find how to use commitizen with a pyproject.toml file that is not in the root folder.

Thus I would like to suggest the addition of the configuration option --config that could be used in pre-commit config like this:

```
-   repo: https://github.com/commitizen-tools/commitizen
    rev: v2.31.0
    hooks:
    -   id: commitizen
        additional_dependencies: ['toml']
        stages: [commit-msg]
        args:
        - --config=config/pyproject.toml
```

### Possible Solution

_No response_

### Additional context

_No response_

### Additional context

_No response_

rockleona:
## Description

CLI now have an option `--config` can specify the config file path,
if the config file is not in root folder, user can specify the located file as well. 

## Checklist

## Expected behavior

`cz` now have an option to specify config file,
if the given file is existed, will only read configuration from the given file,
else, a error would be raise since the given file do not exist.

Execute any `cz` command with `--config` option, if the config file do not exist, a error would raised.

## Additional context

resolve #656

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -94,7 +94,7 @@ data = {
     "arguments": [
         {
             "name": "--config",
-            "help": "specify file path if config file is not in root folder",
+            "help": "the path of configuration file",
         },
         {"name": "--debug", "action": "store_true", "help": "use debug mode"},
         {
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -92,6 +92,10 @@ data = {
     ),
     "formatter_class": argparse.RawDescriptionHelpFormatter,
     "arguments": [
+        {
+            "name": "--config",
+            "help": "specify file path if config file is not in root folder",
+        },
         {"name": "--debug", "action": "store_true", "help": "use debug mode"},
         {
             "name": ["-n", "--name"],
@@ -534,9 +538,7 @@ def parse_no_raise(comma_separated_no_raise: str) -> list[int]:
 
 
 def main():
-    conf = config.read_cfg()
     parser = cli(data)
-
     argcomplete.autocomplete(parser)
     # Show help if no arg provided
     if len(sys.argv) == 1:
@@ -575,6 +577,11 @@ def main():
             )
         extra_args = " ".join(unknown_args[1:])
         arguments["extra_cli_args"] = extra_args
+
+    if args.config:
+        conf = config.read_cfg(args.config)
+    else:
+        conf = config.read_cfg()
 
     if args.name:
         conf.update({"name": args.name})
--- a/commitizen/config/__init__.py
+++ b/commitizen/config/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from pathlib import Path
 
 from commitizen import defaults, git
-from commitizen.exceptions import ConfigFileNotFound
+from commitizen.exceptions import ConfigFileNotFound, ConfigFileIsEmpty
 
 from .base_config import BaseConfig
 from .json_config import JsonConfig
@@ -14,37 +14,23 @@ from .yaml_config import YAMLConfig
 def read_cfg(filepath: str | None = None) -> BaseConfig:
     conf = BaseConfig()
 
-    git_project_root = git.find_git_project_root()
-
     if filepath is not None:
-        given_cfg_path = Path(filepath)
-
-        if not given_cfg_path.exists():
+        if not Path(filepath).exists():
             raise ConfigFileNotFound()
 
-        with open(given_cfg_path, "rb") as f:
-            given_cfg_data: bytes = f.read()
+        cfg_paths = (path for path in (Path(filepath),))
+    else:
+        git_project_root = git.find_git_project_root()
+        cfg_search_paths = [Path(".")]
+        if git_project_root:
+            cfg_search_paths.append(git_project_root)
 
-        given_cfg: TomlConfig | JsonConfig | YAMLConfig
+        cfg_paths = (
+            path / Path(filename)
+            for path in cfg_search_paths
+            for filename in defaults.config_files
+        )
 
-        if "toml" in given_cfg_path.suffix:
-            given_cfg = TomlConfig(data=given_cfg_data, path=given_cfg_path)
-        elif "json" in given_cfg_path.suffix:
-            given_cfg = JsonConfig(data=given_cfg_data, path=given_cfg_path)
-        elif "yaml" in given_cfg_path.suffix:
-            given_cfg = YAMLConfig(data=given_cfg_data, path=given_cfg_path)
-
-        return given_cfg
-
-    cfg_search_paths = [Path(".")]
-    if git_project_root:
-        cfg_search_paths.append(git_project_root)
-
-    cfg_paths = (
-        path / Path(filename)
-        for path in cfg_search_paths
-        for filename in defaults.config_files
-    )
     for filename in cfg_paths:
         if not filename.exists():
             continue
@@ -61,7 +47,9 @@ def read_cfg(filepath: str | None = None) -> BaseConfig:
         elif "yaml" in filename.suffix:
             _conf = YAMLConfig(data=data, path=filename)
 
-        if _conf.is_empty_config:
+        if filepath is not None and _conf.is_empty_config:
+            raise ConfigFileIsEmpty()
+        elif _conf.is_empty_config:
             continue
         else:
             conf = _conf
--- a/commitizen/config/__init__.py
+++ b/commitizen/config/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from pathlib import Path
 
 from commitizen import defaults, git
+from commitizen.exceptions import ConfigFileNotFound
 
 from .base_config import BaseConfig
 from .json_config import JsonConfig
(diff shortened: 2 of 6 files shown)


woile:
What's the motivation for this change?

Lee-W:
> > What's the motivation for this change?
> 
> Simply just found out #656 is quite interesting for me, but actually I can not imagine about the usage or in what scenario user will use this option.

Someone who wants to use a single config for all projects might need it. or mono repo maybe?

commitizen/cli.py
```diff
@@ -92,6 +92,10 @@ def __call__(
     ),
     "formatter_class": argparse.RawDescriptionHelpFormatter,
     "arguments": [
+        {
+            "name": "--config",
+            "help": "specify file path if config file is not in root folder",
```
Lee-W, line 97:
```suggestion
            "help": "the path of configuration file",
```

commitizen/config/__init__.py
```diff
@@ -3,17 +3,39 @@
...
 
 
-def read_cfg() -> BaseConfig:
+def read_cfg(filepath: str | None = None) -> BaseConfig:
     conf = BaseConfig()
 
     git_project_root = git.find_git_project_root()
+
+    if filepath is not None:
+        given_cfg_path = Path(filepath)
```
Lee-W, line 20:
```suggestion
        cfg_path = Path(filepath)
```

cfg_path alone should be expressive enough

commitizen/config/__init__.py
```diff
@@ -3,17 +3,39 @@
...
+            given_cfg_data: bytes = f.read()
+
+        given_cfg: TomlConfig | JsonConfig | YAMLConfig
+
+        if "toml" in given_cfg_path.suffix:
+            given_cfg = TomlConfig(data=given_cfg_data, path=given_cfg_path)
+        elif "json" in given_cfg_path.suffix:
+            given_cfg = JsonConfig(data=given_cfg_data, path=given_cfg_path)
+        elif "yaml" in given_cfg_path.suffix:
+            given_cfg = YAMLConfig(data=given_cfg_data, path=given_cfg_path)
```
Lee-W, line 35: I think we can extract the common part here (also used below)

docs/README.md
```diff
@@ -107,6 +107,7 @@ For more information about the topic go to https://conventionalcommits.org/
 
 optional arguments:
   -h, --help            show this help message and exit
+  --config              specify file path if config file is not in root folder
```
Lee-W, line 110:
```suggestion
  --config              the path of configuration file
```

tests/test_conf.py
```diff
@@ -158,6 +179,32 @@ def test_load_empty_pyproject_toml_and_cz_toml_with_config(_, tmpdir):
             cfg = config.read_cfg()
             assert cfg.settings == _settings
 
+    def test_load_pyproject_toml_not_in_root_folder(_, tmpdir):
```
Lee-W, line 182:
```suggestion
    def test_load_pyproject_toml_from_config_argument(_, tmpdir):
```

this name can be applied to other test cases

Lee-W (changes requested):
Hi @rockleona, thanks so much for implementing this. I left a few comments.

commitizen/config/__init__.py
```diff
@@ -3,28 +3,35 @@
...
-def read_cfg() -> BaseConfig:
+def read_cfg(filepath: str | None = None) -> BaseConfig:
     conf = BaseConfig()
 
-    git_project_root = git.find_git_project_root()
-    cfg_search_paths = [Path(".")]
-    if git_project_root:
-        cfg_search_paths.append(git_project_root)
+    if filepath is not None:
+        cfg_paths = (Path(filepath) for _ in range(1))
```
rockleona, line 18: Hi @Lee-W , since test script would have a error not to use `list` nor `tuple`, I'm wondering any better solution of this, how would you suggest?

Lee-W: hmmmm... It seems I'm not able to reproduce it. Could you please share the steps and error message?

commitizen/config/__init__.py
```diff
@@ -39,7 +46,9 @@ def read_cfg() -> BaseConfig:
         elif "yaml" in filename.suffix:
             _conf = YAMLConfig(data=data, path=filename)
 
-        if _conf.is_empty_config:
+        if filepath is not None and _conf.is_empty_config:
+            raise ConfigFileIsEmpty()
```
Lee-W, line 50: I'm not sure whether we really need to raise an error here. @woile @noirbizarre what do you think?

tests/test_conf.py
```diff
@@ -158,6 +179,40 @@ def test_load_empty_pyproject_toml_and_cz_toml_with_config(_, tmpdir):
...
+            json_cfg_by_class = config.JsonConfig(data=JSON_STR, path=_not_root_path)
+            assert cfg.settings == json_cfg_by_class.settings
+
+    def test_load_cz_yaml_not_from_config_argument(_, tmpdir):
+        with tmpdir.as_cwd():
+            _not_root_path = tmpdir.mkdir("not_in_root").join(".cz.yaml")
+            _not_root_path.write(YAML_STR)
+
+            cfg = config.read_cfg(filepath="./not_in_root/.cz.yaml")
+            yaml_cfg_by_class = config.YAMLConfig(data=JSON_STR, path=_not_root_path)
```
Lee-W, line 205: Should it be YAML_STR instead?

commitizen/config/__init__.py
Lee-W: Got it. I might try 

```suggestion
        cfg_paths = (path for _ in (Path(filepath),))
```

for readability.

commitizen/config/__init__.py
rockleona: IMO, if the file path is been given it must be have some content inside, otherwise there's no need to specify the config file path.

However if the empty config always acceptable, I'll remove the line as well.

tests/test_conf.py
rockleona: Yes, thanks for pointing out.

commitizen/config/__init__.py
Lee-W: After a second thought, I think the current implementation is better. If the user want to ignore it, they can always use no-raise

commitizen/config/__init__.py
```diff
@@ -3,28 +3,35 @@
...
-def read_cfg() -> BaseConfig:
+def read_cfg(filepath: str | None = None) -> BaseConfig:
     conf = BaseConfig()
 
-    git_project_root = git.find_git_project_root()
-    cfg_search_paths = [Path(".")]
-    if git_project_root:
-        cfg_search_paths.append(git_project_root)
+    if filepath is not None:
+        cfg_paths = (_ for _ in (Path(filepath),))
```
Lee-W, line 18:
```suggestion
        cfg_paths = (path for path in (Path(filepath),))
```

commitizen/config/__init__.py
```diff
@@ -3,28 +3,35 @@
...
 
-    cfg_paths = (
-        path / Path(filename)
-        for path in cfg_search_paths
-        for filename in defaults.config_files
-    )
     for filename in cfg_paths:
-        if not filename.exists():
+        if filepath is not None and not filename.exists():
+            raise ConfigFileNotFound()
```
Lee-W, line 33: I feel like we probably would want to move it to the front of the method. Otherwise, we need to do multiple unnecessary check when filepath is not provided

Lee-W (commented):
I think we're really close to merging! In addition to the comment to address, I would suggest you squash the latest 2 commits into the first one as they're minor fixes and do not really need to appear in the changelog. Thanks!

commitizen/config/__init__.py
rockleona: Got it, I've changed this line before assign `cfg_paths` generators.

commitizen/config/__init__.py
```diff
@@ -3,26 +3,34 @@
...
+def read_cfg(filepath: str | None = None) -> BaseConfig:
     conf = BaseConfig()
 
-    git_project_root = git.find_git_project_root()
-    cfg_search_paths = [Path(".")]
-    if git_project_root:
-        cfg_search_paths.append(git_project_root)
+    if filepath is not None:
+        if not Path(filepath).exists():
+            raise ConfigFileNotFound()
```
Lee-W, line 19: nice refactoring!

Lee-W (approved):
LGTM. @woile @noirbizarre I'll merge this one these days. Please let me know if you want to take a deeper look

noirbizarre (approved):
Good to me.

Non blocking suggestion: test and ensure that user home path expansion works (users WILL use home-relative forms for sure so handling cases like `~/path/to/config` from start will avoid inevitable issues)

Lee-W:
> Good to me.
> 
> Non blocking suggestion: test and ensure that user home path expansion works (users WILL use home-relative forms for sure so handling cases like `~/path/to/config` from start will avoid inevitable issues)

Tested on my end. Works fine. Gonna merge it 🚀

PR #1048, opened 2024-04-01
cli(script): gen screenshot for cli helps

Ellie-Yen:
## Description

Add a script to generate screenshot of all cli help contents. 

## Checklist

## Expected behavior

Execute `scripts/gen_cli_help_screenshots.py` will generate screenshots of cli help instructions

1. Execute `scripts/gen_cli_help_screenshots.py`
2. Check each of content insides `commitizen.cli`'s `data` is match its screenshot in `docs/images/cli_help/` folder.

## Additional context

Considering the purpose of the file is more closed to a script, put it into scripts folder.
Should add test and update the documentation accordinately after the change is accepted.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ mkdocs-material = "^9.1.6"
 deprecated = "^1.2.13"
 types-deprecated = "^1.2.9.2"
 types-python-dateutil = "^2.8.19.13"
+rich = "^13.7.1"
 
 
 [tool.poetry.scripts]
--- a/scripts/gen_cli_help_screenshots.py
+++ b/scripts/gen_cli_help_screenshots.py
@@ -1,0 +1,42 @@
+import os
+import subprocess
+from pathlib import Path
+
+from rich.console import Console
+
+from commitizen.cli import data
+
+project_root = Path(__file__).parent.parent.absolute()
+images_root = project_root / Path("docs") / Path("images") / Path("cli_help")
+
+
+def gen_cli_help_screenshots() -> None:
+    """Generate the screenshot for help message on each cli command and save them as svg files."""
+    if not os.path.exists(images_root):
+        os.makedirs(images_root)
+        print(f"Created {images_root}")
+
+    help_cmds = _list_help_cmds()
+    for cmd in help_cmds:
+        file_name = f"{cmd.replace(' ', '_').replace('-', '_')}.svg"
+        _export_cmd_as_svg(cmd, f"{images_root}/{file_name}")
+
+
+def _list_help_cmds() -> list[str]:
+    cmds = [f"{data['prog']} --help"] + [
+        f"{data['prog']} {sub_c['name'] if isinstance(sub_c['name'], str) else sub_c['name'][0]} --help"
+        for sub_c in data["subcommands"]["commands"]
+    ]
+
+    return cmds
+
+
+def _export_cmd_as_svg(cmd: str, file_name: str) -> None:
+    stdout = subprocess.run(cmd, shell=True, capture_output=True).stdout.decode("utf-8")
+    console = Console(record=True, width=80)
+    console.print(f"$ {cmd}\n{stdout}")
+    console.save_svg(file_name, title="")
+
+
+if __name__ == "__main__":
+    gen_cli_help_screenshots()
--- a/docs/images/cli_help/cz___help.svg
+++ b/docs/images/cli_help/cz___help.svg
@@ -1,0 +1,198 @@
+<svg class="rich-terminal" viewBox="0 0 994 904.0" xmlns="http://www.w3.org/2000/svg">
+    <!-- Generated with Rich https://www.textualize.io -->
+    <style>
+
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Regular"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Regular.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Regular.woff") format("woff");
+        font-style: normal;
+        font-weight: 400;
+    }
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Bold"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Bold.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Bold.woff") format("woff");
+        font-style: bold;
+        font-weight: 700;
+    }
+
+    .terminal-911685263-matrix {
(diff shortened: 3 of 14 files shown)


pyproject.toml
```diff
@@ -78,6 +78,7 @@ types-python-dateutil = "^2.8.19.13"
 [tool.poetry.scripts]
 cz = "commitizen.cli:main"
 git-cz = "commitizen.cli:main"
+gen-cli-help-screenshot = "scripts.tasks:run_generate_help_screenshots"
```
woile, line 81: I think this script would make it to the final user, and instead we should simply use it from the scripts folders, what do you think?

scripts/tasks.py
```diff
@@ -0,0 +1,46 @@
+import os
+from pathlib import Path
+
+from invoke import task
```
Lee-W, line 4: We've not yet used `invoke` in this project. Not sure whether we should.

@woile I guess this might not be what we want? (This is something similar to makefile I personally like to use. If my memory serves me right, we don't want to introduce such thing in commitizen?)

scripts/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+from invoke.context import Context
+from rich.console import Console
```
Lee-W, line 6: We'll probably need to add `rich` into pyproject.toml. You can achieve it through

```shell
poetry add rich --group dev
```

Lee-W (commented):
Hi @Ellie-Yen , thank you for your immediate PR! This looks promising! But there are a few things I will need your help with before we can merge it.

1. address the comments we have in separate files
2. reword the commit messages into `ci` instead of `feat` as they're not commitizen features but more like a documentation improvement


After we merge this one, there're a few steps we want to do
1. Add these images to document (somewhere like https://commitizen-tools.github.io/commitizen/init/)
2. Automate the update images scripts in github actions

Thanks again for this cool automation

pyproject.toml
Ellie-Yen: It's a good idea, thank you ~!!!

scripts/tasks.py
Ellie-Yen: ok !
I made a change in
 [ci(scripts): a script to gen cli help screenshots](https://github.com/commitizen-tools/commitizen/pull/1048/commits/b1dc3f721bfa9061d25e09904c9ee50773606093)
 
 Please check it, thank you ~~

scripts/tasks.py
Ellie-Yen: Thanks for reminder ! 
This is updated here: [build(pyproject.toml): add rich in dev.dependencies](https://github.com/commitizen-tools/commitizen/pull/1048/commits/f5c6fd7e932cdee5aa60e294c407849c59099ca1)

scripts/tasks.py
```diff
@@ -0,0 +1,46 @@
+import os
```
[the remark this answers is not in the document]
Ellie-Yen: Yes, the file is renamed to `gen_cli_help_screenshots.py` now after remove `invoke`.
[ci(scripts): a script to gen cli help screenshots](https://github.com/commitizen-tools/commitizen/pull/1048/commits/b1dc3f721bfa9061d25e09904c9ee50773606093)

scripts/gen_cli_help_screenshots.py
```diff
@@ -0,0 +1,46 @@
...
+
+project_root = Path(__file__).parent.parent.absolute()
+images_root = project_root / Path("docs") / Path("images") / Path("cli_help")
+
+
+def gen_cli_help_screenshots() -> None:
+    """Generate the screenshot for help message on each cli command and save them as svg files."""
+    if not os.path.exists(images_root):
+        os.makedirs(images_root)
+        print(f'created {images_root}')
```
Lee-W, line 17:
```suggestion
        print(f"Created {images_root}")
```

scripts/gen_cli_help_screenshots.py
```diff
@@ -0,0 +1,46 @@
...
+    console.print(f"$ {cmd}\n{stdout}")
+    console.save_svg(file_name, title="")
+
+
+# entrypoint
+def main():
+    gen_cli_help_screenshots()
+
+if __name__ == "__main__":
+    main()
```
Lee-W, line 46:
```suggestion
if __name__ == "__main__":
    gen_cli_help_screenshots()
```

scripts/gen_cli_help_screenshots.py
```diff
@@ -0,0 +1,46 @@
...
+        file_name = f"{cmd.replace(' ', '_').replace('-', '_')}.svg"
+        _export_cmd_as_svg(cmd, f"{images_root}/{file_name}")
+
+def _list_help_cmds() -> list[str]:
+    cmds = [
+        f"{data['prog']} --help"
+    ] + list(
+        map(lambda sub_c: f"{data['prog']} {sub_c['name'] if isinstance(sub_c['name'], str) else sub_c['name'][0]} --help",
+            data['subcommands']['commands'])
+        )
```
Lee-W, line 30:
```suggestion
    cmds = [f"{data['prog']} --help"] + [
        f"{data['prog']} {sub_c['name'] if isinstance(sub_c['name'], str) else sub_c['name'][0]} --help"
        for sub_c in data["subcommands"]["commands"]
    ]
```

I think list comprehension would be easier to understand. `map` is suggested when using only built-in function such as `max`, `min`

scripts/gen_cli_help_screenshots.py
Lee-W: we can run ruff or black against these files to improve style consistency

Lee-W (commented):
Thank you for the prompt update!  We're really close to merging.

scripts/gen_cli_help_screenshots.py
Ellie-Yen: Okay. I run `ruff fomat` on this file and its style is  consistent now.
Btw, I notice only `commitizen` and `tests` are included  in the `scripts/format` as below.
Do you think including scripts and other folders is a good idea ?

image

scripts/gen_cli_help_screenshots.py
Ellie-Yen: fixed, thanks !

scripts/gen_cli_help_screenshots.py
Ellie-Yen: I agree. It's more readable than combination of map + list

Lee-W (approved):

pyproject.toml
Lee-W: We could make it a future improvement 👍

scripts/gen_cli_help_screenshots.py
Lee-W: yep, I think that what we should do

Lee-W:
> This looks super cool, do you think it's possible to generate animated gifs as well?

We could make it a future improvement 👍
These help texts might not be as useful. but for demonstrating the interactive TUI, it could be really helpful. I'm thinking of using https://docs.asciinema.org/

Ellie-Yen:
Thanks for reviewing ~
I'll try https://docs.asciinema.org/  soon and let's see if it could get better !

Issue #1053, opened 2024-04-04
Turn on mypy check_untyped_defs = true


Lee-W:
### Description

mypy keeps warning us to use `check_untyped_defs`. When turning it on, I found out there're a few type annotation issues that could be improved and could potentially be an actual issue.

---

### Possible Solution

Fix the type annotation error/warning one by one

### Additional context

list of modules needed to be fixed by 3.21.3

* [ ] commitizen/config/yaml_config.py
* [ ] commitizen/commands/init.py
* [ ] commitizen/config/toml_config.py
* [ ] commitizen/config/json_config.py
* [x] commitizen/cz/base.py
* [ ] tests/test_factory.py
* [ ] commitizen/commands/changelog.py
* [x] commitizen/commands/example.py
* [ ] commitizen/commands/commit.py
* [ ] commitizen/git.py
* [ ] tests/test_changelog.py
* [x] commitizen/cli.py
* [ ] tests/test_cz_customize.py
* [ ] tests/test_cz_conventional_commits.py
* [ ] tests/test_cz_base.py
* [ ] commitizen/commands/info.py
* [x] commitizen/commands/check.py

AdrianDC:
**Possible configurations, commented ones raise new errors to resolve:**

```diff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -171,12 +171,19 @@ convention = "google"
 
 [tool.mypy]
 files = "commitizen"
-disallow_untyped_decorators = true
+# check_untyped_defs = true # TODO
+# disallow_any_generics = true # TODO
+# disallow_incomplete_defs = true # TODO
 disallow_subclassing_any = true
-warn_return_any = true
+# disallow_untyped_calls = true # TODO
+disallow_untyped_decorators = true
+# disallow_untyped_defs = true # TODO
+no_implicit_optional = true
+strict_equality = true
 warn_redundant_casts = true
-warn_unused_ignores = true
+warn_return_any = true
 warn_unused_configs = true
+warn_unused_ignores = true
```

Lee-W:
Sounds like a good idea!

PR #1054, opened 2024-04-05
mypy improvement

Lee-W:
## Description

This is part of https://github.com/commitizen-tools/commitizen/issues/1053, which intends to catch errors earlier through a more strict mypy setting.

## Checklist

## Additional context

#1053


commitizen/cz/customize/customize.py
```diff
@@ -68,16 +68,16 @@ def message(self, answers: dict) -> str:
         else:
             return message_template.render(**answers)
 
-    def example(self) -> str | None:
-        return self.custom_settings.get("example")
+    def example(self) -> str:
+        return self.custom_settings.get("example") or ""
```
woile, line 72: You can set the default for the `.get`, right

woile (approved):
Lgtm

commitizen/cz/customize/customize.py
Lee-W: ah, yes, miss this one (also missed the reviewed...). Let me fix it now!

Lee-W: mypy doesn't seem to be smart enough to make it right. We probably need to rewrite the setting thing a bit to make it work 🤔 But let's keep it for now

PR #1072, opened 2024-04-18
feat(schemes): adds support for SemVer 2.0 (dot in pre-releases) (fix #1025)

### LINKED ISSUE #1025: Generated pre-release part of SemVer differs from SemVer spec
### Description

When using SemVer schema and bumping to pre-release version, the version generated is like `1.0.0-a0`.

This differs from examples from [SemVer spec 9](https://semver.org/#spec-item-9), which is `1.0.0-alpha.0`, using "alpha" instead of "a", and having a dot between "a" and the sequential number.

Current behavior may break the version comparison when the number has more than 1 digits. 
For example, `1.0.0-a10` is less than `1.0.0-a2` according to [spec 11](https://semver.org/#spec-item-11), because `a10` is less than `a2` lexically.  

### Steps to reproduce

1. Setup commitizen using SemVer schema
2. Run `cz bump --pr alpha`

### Current behavior

Bumped into `x.y.z-a0`

### Desired behavior

Bumped into `x.y.z-a.0` or `x.y.z-alpha.0`

### Screenshots

_No response_

### Environment

Commitizen Version: 3.18.3
Python Version: 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Operating System: Linux

noirbizarre:
## Description
This PR adds a `semver2` version scheme which is applying strict `prerelease` formatting as described in [SemVer 2 Item 9](https://semver.org/spec/v2.0.0.html#spec-item-9) and [Semver 2.0.0 Item 11](https://semver.org/spec/v2.0.0.html#spec-item-11), meaning:
- each prelease identifier should be separated by a dot
- dash is part of the identifier, not a separator
- precedence rules should be applied on identifiers

As a consequence, the `1.2.3-alpha0` form is not respecting those rules (`1.2.3-alpha10` has not proper precedence over `1.2.3-alpha2`), nor `1.2.3-alpha.0-dev.0` as its prerelease identifiers will be `alpha`, `0-dev`, `0` while it should be `alpha`, `0`, `dev` and `0`.

Note that Semver 2 is underspecified on this point (there is no formal description of what are the expected prerelease identifiers, only example) and so this PR chose to stay close to the examples (use of `alpha` and `beta` instead of `a` and `b`), but this is choice, not a specification requirement.

## Checklist

## Expected behavior

SemVer 2.0 versions are properly parsed and bumped, the resulting version is SemVer 2,0 compliant too.

1. Defined `version_scheme = "semver2"` in you commitizen config
2. Bump a dev release, it should have the form `x.y.z-dev.D`
3. Bump an alpha release, it should have the form `x.y.z-alpha.A`
4. Bump an alpha dev release, it should have the form `x.y.z-alpha.A.dev.D`

## Additional context

Fixes #1025

> [!Note]
> I think it's time for removing the `version-type` deprecated by `vesion-scheme` (in another PR)

### DIFF
--- a/commitizen/version_schemes.py
+++ b/commitizen/version_schemes.py
@@ -310,7 +310,7 @@ class SemVer(BaseVersion):
     """
     Semantic Versioning (SemVer) scheme
 
-    See: https://semver.org/
+    See: https://semver.org/spec/v1.0.0.html
     """
 
     def __str__(self) -> str:
@@ -324,9 +324,8 @@ class SemVer(BaseVersion):
         parts.append(".".join(str(x) for x in self.release))
 
         # Pre-release
-        if self.pre:
-            pre = "".join(str(x) for x in self.pre)
-            parts.append(f"-{pre}")
+        if self.prerelease:
+            parts.append(f"-{self.prerelease}")
 
         # Post-release
         if self.post is not None:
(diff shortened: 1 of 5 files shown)


woile (approved):
Awesome work again 🎉 
I'm in for a breaking change cleaning up some old deprecated flags.

PR #1076, opened 2024-04-21
feat: add an argument to limit the length of commit message

kevin1kevin1k:
## Description

Re-opened another PR for #191 as the long-hanging https://github.com/commitizen-tools/commitizen/pull/557 has been closed.
Check if the length of commit message exceeds the specified limit.
The limit can be specified via, for example, `-l 72` or `--message-length-limit 72`.

## Checklist

## Expected behavior

If one runs `cz commit -l/--message-length-limit N`, the commit would fail (with the exception `CommitMessageLengthExceededError`) when the message is longer than `N` characters.
Note that for `ConventionalCommitsCz`, the limit applies only from the prefix to the subject.
In other words, everything after the first line (the body and the footer) are not counted in the length.

1. `poetry run python commitizen/cli.py -n cz_conventional_commits c -l 10` (or `cz -n cz_conventional_commits c -l 10` when merged)
2. ENTER (type of change is `fix: ` with length 5)
3. ENTER (no scope)
4. type `123456` with length 6
5. ENTER (no body)
6. ENTER (not breaking change)
7. ENTER (no footer)

As 5+6 = 11 > 10 the process should exit with message `Length of commit message exceeds limit (11/10)` and error code 32.

## Additional context

Closes: #191

### DIFF
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -66,7 +66,7 @@ class Commit:
         message = cz.message(answers)
         message_len = len(message.partition("\n")[0].strip())
         message_length_limit: int = self.arguments.get("message_length_limit", 0)
-        if message_length_limit > 0 and message_len > message_length_limit:
+        if 0 < message_length_limit < message_len:
             raise CommitMessageLengthExceededError(
                 f"Length of commit message exceeds limit ({message_len}/{message_length_limit})"
             )
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -61,7 +61,9 @@ class Commit:
 
         if not answers:
             raise NoAnswersError()
-        return cz.message(answers)
+
+        message_length_limit: int = self.arguments.get("message_length_limit", 0)
+        return cz.message(answers, message_length_limit=message_length_limit)
 
     def __call__(self):
         dry_run: bool = self.arguments.get("dry_run")
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -64,7 +64,7 @@ class Commit:
             raise NoAnswersError()
 
         message = cz.message(answers)
-        message_len = len(message.partition("\n")[0])
+        message_len = len(message.partition("\n")[0].strip())
         message_length_limit: int = self.arguments.get("message_length_limit", 0)
         if message_length_limit > 0 and message_len > message_length_limit:
             raise CommitMessageLengthExceededError(
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -12,6 +12,7 @@ from commitizen.cz.exceptions import CzException
 from commitizen.cz.utils import get_backup_file_path
 from commitizen.exceptions import (
     CommitError,
+    CommitMessageLengthExceededError,
     CustomError,
     DryRunExit,
     NoAnswersError,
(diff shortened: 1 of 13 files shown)


commitizen/commands/commit.py
```diff
@@ -61,7 +61,9 @@ def prompt_commit_questions(self) -> str:
 
         if not answers:
             raise NoAnswersError()
-        return cz.message(answers)
+
+        message_length_limit: int = self.arguments.get("message_length_limit", 0)
+        return cz.message(answers, message_length_limit=message_length_limit)
```
kevin1kevin1k, line 66: One interesting subtlety here: there is a reason why `_check_message_length_limit` is not called here, but instead called in `message` of each child of `BaseCommitizen`.
It's currently because for conventional commits the whole message may consist of multiple lines, where the limit should only be applied to the first line.
There is a possible alternative: call `_check_message_length_limit` and only take stuff before the first line break (`\n`).
But I assume the current form allows more flexible settings for each child of `BaseCommitizen`, so did not adopt that alternative.

Lee-W: Yep, I think we should take care of only the first line. and this behavior should be clearly documentated

kevin1kevin1k: No problem.
The new commit simplifies accordingly, and the behavior is also reflected in commit.md.
A missing test `test_commit_command_with_message_length_limit` is also added.

commitizen/commands/commit.py
```diff
@@ -61,7 +62,16 @@ def prompt_commit_questions(self) -> str:
 
         if not answers:
             raise NoAnswersError()
-        return cz.message(answers)
+
+        message = cz.message(answers)
+        message_len = len(message.partition("\n")[0])
```
Lee-W, line 67: I didn't know we could use `partition` 👀 Great job

Lee-W: should we `strip()` as well?

docs/commit.md
```diff
@@ -36,3 +36,14 @@ You can use `cz commit --retry` to reuse the last commit message when the previo
...
 a new commit message to be prompted.
+
+### Commit message length limit
+
+The argument `-l` (or `--message-length-limit`) followed by a positive number can limit the length of commit messages.
+An exception would be raised when the message length exceeds the limit.
+For example, `cz commit -l 72` will limit the length of commit messages to 72 characters.
+By default the limit is set to 0, which means no limit on the length.
+
+Note that the limit applies only to the first line of message.
```
Lee-W, line 47:
```suggestion
**Note that the limit applies only to the first line of the message.**
```

Lee-W (commented):
Great job @kevin1kevin1k ! Left a few nitpicks. But we're pretty close to merge this one

Lee-W:
Also, great work on the detailed PR description.

commitizen/commands/commit.py
kevin1kevin1k: Sure, thanks for the suggestion!

docs/commit.md
kevin1kevin1k: Indeed this should be highlighted.

Lee-W (approved):

Lee-W:
Great job @kevin1kevin1k !

@woile @noirbizarre I'm planing to merge this one these days. Let me know if you want to take a deeper look. Thanks!

commitizen/commands/commit.py
```diff
@@ -61,7 +62,16 @@ def prompt_commit_questions(self) -> str:
 
         if not answers:
             raise NoAnswersError()
-        return cz.message(answers)
+
+        message = cz.message(answers)
+        message_len = len(message.partition("\n")[0].strip())
+        message_length_limit: int = self.arguments.get("message_length_limit", 0)
+        if message_length_limit > 0 and message_len > message_length_limit:
```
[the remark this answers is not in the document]
kevin1kevin1k: Thanks for the polishing!

Issue #950, opened 2023-12-29
Help wanted: commitizen 3.x not supoort generate CHANGELOG with tag format like v1.0.0-release


dhmemi:
### Description

Commitizen 3.x not supoort generate CHANGELOG with tag format like `v1.0.0-release`.

My repo tag could not be change as it released. Any idea to solve this problem in order to generate the changelog correctly? 

### Steps to reproduce

```
PS E:\codes\aidi-sdk> cz ch                           
InvalidVersion GitTag('v0.7.1-release', '0689497d1cf9d036cb510cfed3a96135d9456419', '2023-12-08')
InvalidVersion GitTag('v0.7.0-release', 'c81dd1acd809ea2df66154bc5ecb64fb625a40e6', '2023-12-07')
```

### Current behavior

no

### Desired behavior

mo

### Environment

`cz version`: 3.13.0

rshmhrj:
This is also happening for maven type versions (using cz version 3.13.0):

```bash
cz init
Welcome to commitizen!

Answer the questions to configure your project.
For further configuration visit:

https://commitizen-tools.github.io/commitizen/config/

? Please choose a supported config file:  .cz.toml
? Please choose a cz (commit rule): (default: cz_conventional_commits) cz_conventional_commits
? Choose the source of the version: commitizen: Fetch and set version in commitizen config (default)
? Is master_0.0.2-SNAPSHOT.74 the latest tag? No
? Please choose the latest tag:  v0.0.1-SNAPSHOT
? Choose version scheme:  semver
Traceback (most recent call last):
  File "/usr/local/bin/cz", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/Cellar/commitizen/3.13.0/libexec/lib/python3.12/site-packages/commitizen/cli.py", line 570, in main
    args.func(conf, arguments)()
  File "/usr/local/Cellar/commitizen/3.13.0/libexec/lib/python3.12/site-packages/commitizen/commands/init.py", line 100, in __call__
    version = get_version_scheme(self.config, version_scheme)(tag)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/opt/python-packaging/lib/python3.12/site-packages/packaging/version.py", line 200, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
[...]

PR #1081, opened 2024-04-24
Fix semver not fully covered

Lee-W:
## Description

`1.0.0-reallyweird` is a valid semver but not yet covered by commitizen. Possible solution could be https://github.com/commitizen-tools/commitizen/issues/950#issuecomment-2072666867

Closes: #950 

## Checklist

### DIFF
--- a/tests/test_version_scheme_semver.py
+++ b/tests/test_version_scheme_semver.py
@@ -64,6 +64,7 @@ weird_cases = [
     (("1.0.0alpha1", None, "alpha", 0, None), "1.0.0-a2"),
     (("1", None, "rc", 0, None), "1.0.0-rc0"),
     (("1.0.0rc1+e20d7b57f3eb", "PATCH", None, 0, None), "1.0.0"),
+    (("1.0.0-reallyweird", "PATCH", "reallyweird", 0, None), "1.0.0"),
 ]
 
 # test driven development
--- a/tests/test_version_scheme_semver2.py
+++ b/tests/test_version_scheme_semver2.py
@@ -5,6 +5,8 @@ import pytest
 
 from commitizen.version_schemes import SemVer2, VersionProtocol
 
+# current_version, increment, prerelease, prerelease_offset, devrelease
+
 simple_flow = [
     (("0.1.0", "PATCH", None, 0, None), "0.1.1"),
     (("0.1.0", "PATCH", None, 0, 1), "0.1.1-dev.1"),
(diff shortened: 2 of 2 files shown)


Lee-W:
Currently, only test cases are added. Will need to find some time to wrap it up. Feel free to take over if anyone's interested in it 🙂

Lee-W:
I think this one would be a bit tricky 🤔 prerelease can be anything for semver. Thus, bump prerelease might not make sense to semver. e.g., what should be the next version of `1.1.0-test` @woile @noirbizarre WDYT?

woile:
I completely missed this. I don't know the answer, but I think we could stick to adding the `0` at the end (?)

```
1.1.0-test.0
1.1.0-test.1
1.1.0-test.2
```

Similar to semver:
```
1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1
```

And we could assume anything unknown is before alpha (?)

Lee-W:
> I completely missed this. I don't know the answer, but I think we could stick to adding the `0` at the end (?)
> 
> ```
> 1.1.0-test.0
> 1.1.0-test.1
> 1.1.0-test.2
> ```
> 
> Similar to semver:
> 
> ```
> 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1
> ```
> 
> And we could assume anything unknown is before alpha (?)

Sounds like a good idea 🤔  I'll get back to this one when I have more bandwidth

dhmemi:
> I completely missed this. I don't know the answer, but I think we could stick to adding the `0` at the end (?)
> 
> ```
> 1.1.0-test.0
> 1.1.0-test.1
> 1.1.0-test.2
> ```
> 
> Similar to semver:
> 
> ```
> 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1
> ```
> 
> And we could assume anything unknown is before alpha (?)


Perhaps this could be done by having the user configure a prerelease word list and bump to the next item in the list. If the current prerelease is already the last word, upgrade the previous patch version.
In the case where the user has not given a word list, ignore `prerelease`, and retain any subsequent text after `1.0.0-` as is ?

Lee-W:
Sounds like a valid idea 👀

tests/test_version_scheme_semver.py
```diff
@@ -64,6 +64,7 @@
     (("1.0.0alpha1", None, "alpha", 0, None), "1.0.0-a2"),
     (("1", None, "rc", 0, None), "1.0.0-rc0"),
     (("1.0.0rc1+e20d7b57f3eb", "PATCH", None, 0, None), "1.0.0"),
```
gsemet, line 66: this is not semver but PEP440

Issue #1115, opened 2024-05-15
missing help text when subcommand is used (e.g., `cz commit --help`)


Lee-W:
### Description

When running `cz --help`, we have the help text for each subcommand (e.g., `cz commit`), but `cz commit --help` does not have the description `create new commit`

### Steps to reproduce

1. `cz --help`
2. `cz commit --help`hope

### Current behavior

missing help text in subcommandthe 

### Desired behavior

the help text should appear

### Screenshots

<img width="1714" alt="image" src="https://github.com/commitizen-tools/commitizen/assets/5144808/1e1f0f2e-39ad-4e2e-91bd-7410b50c0ff7">

### Environment

Commitizen Version: 3.24.0
Python Version: 3.12.2 (main, Apr 19 2024, 14:41:00) [Clang 15.0.0 (clang-1500.3.9.4)]
Operating System: Darwin

marcosdotme:
@Lee-W Can I act on this?

Lee-W:
sure! that would be awesome. assigning this to you

marcosdotme:
@Lee-W Waiting review on #1120 🤝🏻

PR #1120, opened 2024-05-16
fix: add description for subcommands

marcosdotme:
## Checklist

- [ ] Update the documentation for the changes (need remake a bunch of gifs/images @Lee-W)

## Expected behavior
You must see the command description when running with `--help` option

1. `cz commit --help`

## Screenshot
image


marcosdotme:
@Lee-W I delete the `tests/commands/test_other_commands.py` an move the tests for the correspondent file without asking you. Let me know if you need that I revert this.

commitizen/cli.py
```diff
@@ -114,11 +114,13 @@ def __call__(
         "commands": [
             {
                 "name": ["init"],
+                "description": "init commitizen configuration",
                 "help": "init commitizen configuration",
```
woile, line 118: Should we remove `help`? is it useful at all?

Lee-W: @marcosdotme Could you please check whether we could just remove "help"? This PR mostly looks good to me 👍

tests/commands/test_bump_command.py
```diff
@@ -1449,3 +1449,15 @@ def test_bump_changelog_contains_increment_only(mocker, tmp_commitizen_project,
...
+def test_bump_command_shows_description_when_use_help_option(
+    mocker: MockFixture, capsys
+):
+    testargs = ["cz", "bump", "--help"]
+    mocker.patch.object(sys, "argv", testargs)
+    with pytest.raises(SystemExit):
+        cli.main()
+
+    out, _ = capsys.readouterr()
+    assert "bump semantic version based on the git log" in str(out).replace("\n", " ")
```
Lee-W, line 1463: none blocker, but we could try something like https://github.com/commitizen-tools/commitizen/blob/f1ece99dc9ca142268b912671b04dfdfe5addb3f/tests/test_bump_update_version_in_files.py#L157-L158 as well

commitizen/cli.py
marcosdotme: I we remove `help`, the command is not shown when running `cz --help`

image

CC: @Lee-W

tests/commands/test_bump_command.py
marcosdotme: I don't know if I understand... I wrote the `test_bump_command_shows_description_when_use_help_option` to assert that the command description shows on running `--help` against the command. Can you explain your idea with other words?

Lee-W: `pytest-regeression` is a tool that you can verify whether the content (e.g., str, data frame) is exactly the same as the last time we generated it. It done so by creating a text file that records the last result. https://pytest-regressions.readthedocs.io/en/latest/overview.html

marcosdotme: I will look to this later, today. Thanks!

marcosdotme: @Lee-W `pytest-regressions` is a game changer! Using the assert only approach isn't enough to check if the command descriptions matches. For example:

Using `assert "create new commit" in str(out).replace("\n", " ")` isn't reliable because if we add an dot to the command description "create new commit." it still matching.

But, using `file_regression` we can check for the entire output and adding the dot to the end of the command description will cause the test to fail.

I replaced all the assertion approach for the `file_regression`, except for two: **changelog** and **bump**.

Analyzing the failure I found the "problem" is in the file `commitizen.version_schemes.py`. Theres an constant called **KNOWN_SCHEMES** that is an set of all available scheme providers. The problem is: set is an unordered struct so the order changes for every call. Since the order changes for every call, we cannot assure using `file_regression`:

image

To fix this we can just replace the set comprehension for an list comprehension:
```diff
- KNOWN_SCHEMES = {ep.name for ep in metadata.entry_points(group=SCHEMES_ENTRYPOINT)}
+ KNOWN_SCHEMES = [ep.name for ep in metadata.entry_points(group=SCHEMES_ENTRYPOINT)]
```

I change the set for list and if passes on all tests, and now we can use `file_regression`. Theres an reason this to be an set? We can change to list?

marcosdotme: I'm working on the changes requested on #1115 (move step from generate_cli_screenshots.yml to docspublish.yml) and theres an another problem caused by using the set comprehension:

image

Even when thats no real changes, **bump** and **changelog** svg are being re-generated due to the order change on **KNOWN_SCHEMES**

Lee-W: > I change the set for list and it passes on all tests, and now we can use file_regression. Theres an reason this to be an set? We can change to list?

I guess that's for performance and the fact that it didn't really need to be a list. But it could be a list in this case. Nice catching!

commitizen/cli.py
Lee-W: Yep, I guess we'll have to keep them then. I'm not sure whether there's a way we can remove this duplication. But let's keep it for another PR.

tests/commands/test_bump_command.py
Lee-W: @marcosdotme Have we pushed the `file_regression` change?

marcosdotme: Waiting more opinions about that, so I can create an issue, resolve then, merge into the branch of this PR an it will automatically pass all tests.

marcosdotme: @Lee-W Not yet! Is in stage. I was waiting for any thoughts about the **KNOWN_SCHEMES** "problem"

Lee-W: I think we're good to change `KNOWN_SCHEMES` to list and change test cases to file_regressions. Let me know if I missed anything 🙂

Lee-W: Just modify this one should be fine 🙂

marcosdotme: @Lee-W  I will wait to #1126 be merged because this branch will also fail on CI due to trailing whitespaces on ***.svg**

Lee-W: It's merged 🎉

marcosdotme: @Lee-W After run the `./scripts/test` it fails because the missing ":" on commit message:

image

The commit is:

image

How to proceed? 🥲

Lee-W: oh just ignore it

marcosdotme:
@Lee-W Can you help me? Dont know whats happening here lol. All works fine on my local branch, all tests passes:

image

But it still failing on CI

Lee-W:
> @Lee-W Can you help me? Dont know whats happening here lol. All works fine on my local branch, all tests passes:
> 
> image
> 
> But it still failing on CI

Sure, let me take a look 🙂

Lee-W:
I guess it's due to the latest main branch change. could you please run `poetry run pytest --force-regen` and see whether it fixes the CI? If not, I'll take a deeper look

Lee-W:
got it. let me take a look

Lee-W:
I know what's happening. The message of argparse is different in different Python versions.

Lee-W:
Changed in Python 3.10. We could probably generate two sets of expected results and run that test based on different Python versions. We could try https://docs.pytest.org/en/latest/how-to/skipping.html#id1. Not sure whether there's something like parameterize we can use in this case

marcosdotme:
> Changed in Python 3.10. We could probably generate two sets of expected results and run that test based on different Python versions. We could try https://docs.pytest.org/en/latest/how-to/skipping.html#id1. Not sure whether there's something like parameterize we can use in this case. @marcosdotme do you want me to resolve it? or do you want to take a look?

Can you resolve this one?

Lee-W:
> > Changed in Python 3.10. We could probably generate two sets of expected results and run that test based on different Python versions. We could try https://docs.pytest.org/en/latest/how-to/skipping.html#id1. Not sure whether there's something like parameterize we can use in this case. @marcosdotme do you want me to resolve it? or do you want to take a look?
> 
> Can you resolve this one? So I can learn from watching this time

Sure 🙂

Lee-W:
@marcosdotme Done. I decide not to cover < 3.10 as this is not important enough to check

Lee-W (approved):

marcosdotme:
> @marcosdotme Done. I decide not to cover < 3.10 as this is not important enough to check

Ok! Thanks 🤝🏻

Lee-W:
@woile @noirbizarre I'm planning on merging it these days. Please let me know if you'd like to take a look. 🙂

PR #1143, opened 2024-05-24
feat: update default changelog titles

### LINKED ISSUE #1119: Default changelog titled for `cz_conventional_commits`
### Description

In general the Conventional Commits / Angular scheme is suitable for me (as for most of the people write code with VSC, I guess). It is nicely implemented in commitizen and suggested to be set as a default (see #535, [comment](https://github.com/commitizen-tools/commitizen/issues/535#issuecomment-1192676349)). That's great!

The only thing I don't get is the way the changelog sections are formatted:

https://github.com/commitizen-tools/commitizen/blob/3e57007bf1f882a6ed34717c68d3a265e16dcf7a/commitizen/cz/conventional_commits/conventional_commits.py#L35-L40

Is there any specific reason (or convention) for which the section titles are as they are?

Changelogs are supposed to be read by humans (?), so why more human-readable names aren't applied?

Regards! 🙂

### Possible Solution

```python
change_type_map = { 
     "feat": "Features", 
     "fix": "Bug fixes", 
     "refactor": "Code refactoring", 
     "perf": "Performance improvements", 
 } 
```

The `change_type_order` must be updated accordingly.

### Additional context

I am aware of the fact that my issue can be resolved by `cz_customize`. This, however, requires providing extra configurations, e.g. `bump_pattern`, etc.

### Additional context

_No response_

paduszyk:
## Description

This updates the default titles of the changelog's sections in `ConventionalCommitsCz`.

## Checklist

## Expected behavior

Changelogs are more human-readable.

## Additional context

Closes #1119.


Lee-W (approved):

tests/conftest.py
```diff
@@ -253,3 +254,17 @@ def any_changelog_format(config: BaseConfig) -> ChangelogFormat:
...
     return get_changelog_format(config)
+
+
+@pytest.fixture(autouse=True)
+def default_change_type_map(mocker: MockerFixture) -> None:
+    mocker.patch.object(
+        ConventionalCommitsCz,
+        "change_type_map",
+        {
+            "feat": "Feat",
```
woile, line 265: Why are the tests using the old format?

Lee-W:
@paduszyk Hi, would it be possible for us to address https://github.com/commitizen-tools/commitizen/issues/1119#issuecomment-2132866909. I think we're really close to merging!

noirbizarre:
I would flag this change as breaking as it means any user having a custom `change_type_order` with the default `change_type_map` will see its changelog broken without a manual intervention  (aka. updating the `change_type_order` with the new names).
So maybe should include this one in the next major (v4?) ?

Lee-W:
> I would flag this change as breaking as it means any user having a custom `change_type_order` with the default `change_type_map` will see its changelog broken without a manual intervention (aka. updating the `change_type_order` with the new names). So maybe should include this one in the next major (v4?) ?

Sounds good! Let me add the label

tests/conftest.py
paduszyk: I did not want to change the tests. So I applied mocking instead to hard-coding the new values.

IMHO: For tests, the names should not matter, I guess. The point is what the values are actually used for.

Lee-W: It might be a bit confusing as these are no longer the actual `default`_change_type_map when someone's tracing the test cases, they might be misled

Lee-W (changes requested):
I think we need to change the test cases

paduszyk:
> I think we need to change the test cases

It's been a while... Sorry.

@Lee-W Are you sure? This will result in a relatively large diff, taking into account the source itself will remain basically untouched except the update literals...

bearomorphism:
I agree that test cases should be updated, otherwise it will cause confusion to future contributors.

I don't think relatively large diff on test files is an issue.

Issue #1135, opened 2024-05-21
cz commit -s  should fail with deprecation warning BEFORE going through dialog, --help should reflect behavior


yarikoptic:
### Description

Bad Dev Experience if asks for all the questions to only fail after

### Steps to reproduce

Run `cz commit -s -a`

### Current behavior

```shell
❯ cz commit -s -a
? Select the type of change you are committing style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
? What is the scope of this change? (class or file name): (press [enter] to skip)
 
? Write a short and imperative summary of the code changes: (lower case and no period)
 move codespell configuration to pyproject.toml
? Provide additional contextual information about the code changes: (press [enter] to skip)
 So people could just run "codespell" without pre-commit and have centralized configuration for tools (the others are already in pyproject.toml)
? Is this a BREAKING CHANGE? Correlates with MAJOR in SemVer No
? Footer. Information about Breaking Changes and reference issues that this commit closes: (press [enter] to skip)
 

style: move codespell configuration to pyproject.toml

So people could just run "codespell" without pre-commit and have centralized configuration for tools (the others are already in pyproject.toml)

signoff mechanic is deprecated, please use `cz commit -- -s` instead.
fatal: /home/yoh/.tmp/tmpsqo29rpl: '/home/yoh/.tmp/tmpsqo29rpl' is outside repository at '/home/yoh/proj/misc/commitizen'
```

### Desired behavior

```shell
❯ cz commit -s -a
signoff mechanic is deprecated, please use `cz commit -- -s` instead.
```

Also --help should mention `--` -- not there at all

```shell
❯ cz commit --help
usage: cz commit [-h] [--retry] [--no-retry] [--dry-run] [--write-message-to-file FILE_PATH] [-s] [-a] [-l MESSAGE_LENGTH_LIMIT]

options:
  -h, --help            show this help message and exit
  --retry               retry last commit
  --no-retry            skip retry if retry_after_failure is set to true
  --dry-run             show output to stdout, no commit, no modified files
  --write-message-to-file FILE_PATH
                        write message to file before committing (can be combined with --dry-run)
  -s, --signoff         sign off the commit
  -a, --all             Tell the command to automatically stage files that have been modified and deleted, but new files you have not told Git about are not affected.
  -l MESSAGE_LENGTH_LIMIT, --message-length-limit MESSAGE_LENGTH_LIMIT
                        length limit of the commit message; 0 for no limit
[...]

PR #1146, opened 2024-05-28
Shows deprecated warning before prompt questions

marcosdotme:
## Description
Shows deprecated warning before prompt questions to dev, for an better experience.

## Checklist

- [x] Update the documentation for the changes **_(deprecated warning already in docs)_**

### Shows deprecated warning before prompt questions
image

### Shows deprecated warning on command help text
image

### Dont' shows deprecated warning if uses `--signoff` option the right way
image

### DIFF
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -73,10 +73,24 @@ class Commit:
         return message
 
     def __call__(self):
+        signoff: bool = (
+            self.arguments.get("signoff") or self.config.settings["always_signoff"]
+        )
         dry_run: bool = self.arguments.get("dry_run")
         write_message_to_file: bool = self.arguments.get("write_message_to_file")
-
+        retry: bool = self.arguments.get("retry")
+        no_retry: bool = self.arguments.get("no_retry")
+        retry_after_failure: bool = self.config.settings.get("retry_after_failure")
         is_all: bool = self.arguments.get("all")
+
+        if signoff:
+            out.warn(
+                "signoff mechanic is deprecated, please use `cz commit -- -s` instead.\n"
+            )
+            extra_args = self.arguments.get("extra_cli_args", "--") + " -s"
+        else:
+            extra_args = self.arguments.get("extra_cli_args", "")
+
         if is_all:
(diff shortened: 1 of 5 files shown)


marcosdotme:
Relates to #1135

commitizen/commands/commit.py
```diff
@@ -73,10 +73,24 @@ def prompt_commit_questions(self) -> str:
...
     def __call__(self):
+        signoff: bool = (
+            self.arguments.get("signoff") or self.config.settings["always_signoff"]
+        )
         dry_run: bool = self.arguments.get("dry_run")
         write_message_to_file: bool = self.arguments.get("write_message_to_file")
-
+        retry: bool = self.arguments.get("retry")
+        no_retry: bool = self.arguments.get("no_retry")
+        retry_after_failure: bool = self.config.settings.get("retry_after_failure")
```
Lee-W, line 83: Should we add a default value to these `get`?

commitizen/commands/commit.py
```diff
@@ -73,10 +73,24 @@ def prompt_commit_questions(self) -> str:
...
         is_all: bool = self.arguments.get("all")
+
+        if signoff:
+            out.warn(
+                "signoff mechanic is deprecated, please use `cz commit -- -s` instead.\n"
+            )
+            extra_args = self.arguments.get("extra_cli_args", "--") + " -s"
```
Lee-W, line 90: Let's use f-string here

Lee-W (commented):
Would it be possible for us to also document the `--` behavior in help text 🤔

marcosdotme:
> Would it be possible for us to also document the `--` behavior in help text 🤔

Of course. I will add this too

commitizen/commands/commit.py
marcosdotme: Hmmm, we can add a explicit default value but `.get` has a default value to `None` if not specified.

Lee-W: Yep, but `None` is not a `bool`, `False` would be better 🙂

PR #1178, opened 2024-07-09
feat: add argument to limit length of commit message in checks

jwidauer:
## Description
Recently there was an addition to add a message length limit in #1076.
However, those changes only applied to the `cz commit` command.
This PR also adds this new command line argument to the `cz check` command, to make checking for these violations possible.

## Checklist

## Expected behavior
When running `cz check -l N -m MESSAGE` with N being greater than the length of  the header of MESSAGE, the check should fail.


Lee-W (approved):
@jwidauer This is great! Thanks for adding this. I'm planning to merge it these days so that others would have a chance to take a look.

PR #1195, opened 2024-07-30
feat(bump): add functionality to write the next version to stdout

marcelblijleven:
Adds a --get-next flag which determines the next version and writes it to stdout.

## Description

The new `--get-next` flag can be used to determine the next version based on the available commits, and output it to stdout. This can be useful in CI, for example to determine the version for a preview build image.

## Checklist

## Expected behavior

When provided, the `--get-next` flag makes commitizen behave similarly to `--dry-run`, it doesn't change any files and it doesn't create a tag. It determines the next version the same way a dry run does. It writes the next version to stdout, with no other output. 

The flag can't be used with `--changelog`, `--changelog-to-stdout` and `MANUAL_VERSION`, a `NotAllowed` exception will be raised if these options are provided together with `--get-next`.

An example use case is assigning the output to a variable in CI.

1. Make sure you have fetched tags locally
2. Add a commit that is eligible for a bump
3. Run `poetry run cz bump --get-next`
4. Compare the output with `poetry run cz bump --dry-run`

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -14,6 +14,7 @@ from commitizen.exceptions import (
     BumpTagFailedError,
     DryRunExit,
     ExpectedExit,
+    GetNextExit,
     InvalidManualVersion,
     NoCommitsFoundError,
     NoneIncrementExit,
@@ -159,6 +160,7 @@ class Bump:
         manual_version = self.arguments["manual_version"]
         build_metadata = self.arguments["build_metadata"]
         increment_mode: str = self.arguments["increment_mode"]
+        get_next: bool = self.arguments["get_next"]
 
         if manual_version:
             if increment:
@@ -190,6 +192,9 @@ class Bump:
                     "--prerelease-offset cannot be combined with MANUAL_VERSION"
                 )
 
+            if get_next:
+                raise NotAllowed("--get-next cannot be combined with MANUAL_VERSION")
+
         if major_version_zero:
             if not current_version.release[0] == 0:
                 raise NotAllowed(
@@ -202,6 +207,18 @@ class Bump:
                     "--local-version cannot be combined with --build-metadata"
                 )
 
+        # If user specified changelog_to_stdout, they probably want the
+        # changelog to be generated as well, this is the most intuitive solution
+        self.changelog = self.changelog or bool(self.changelog_to_stdout)
+
+        if get_next:
+            if self.changelog:
+                raise NotAllowed(
+                    "--changelog or --changelog-to-stdout is not allowed with --get-next"
+                )
+            # Setting dry_run to prevent any unwanted changes to the repo or files
+            self.dry_run = True
+
         current_tag_version: str = bump.normalize_tag(
             current_version,
             tag_format=tag_format,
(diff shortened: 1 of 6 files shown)


woile:
Thanks for the contribution! It looks good 🎉 . Could you add a comparison in the docs with `dry-run`? Why chose one over the other.

Moreover, should this be part of the `version` command? I wonder if instead we should have: `cz version --project --next` 🤔  I know it might be a bit harder to implement. @Lee-W ?

Lee-W:
The idea looks great! I'll take a look this weekend when leading yet another dev sprint haha

marcelblijleven:
Thanks for the feedback, I've just added some additional information to the documentation.

Lee-W:
did not actually have the time to check it this weekend 🤦‍♂️ will try to do it these days!

commitizen/commands/bump.py
```diff
@@ -190,6 +192,9 @@ def __call__(self) -> None:  # noqa: C901
                     "--prerelease-offset cannot be combined with MANUAL_VERSION"
                 )
 
+            if get_next:
```
Lee-W, line 195: Probably not required in this PR, but I feel we could move these validations out as a separate function 🤔

docs/commands/bump.md
```diff
@@ -285,6 +285,39 @@ You should normally not use this functionality, but if you decide to do, keep in
...
+### `--get-next`
+
+Provides a way to determine the next version and write it to stdout. This parameter is not compatible with `--changelog`
+and `manual version`.
+
+```bash
+cz bump --get-next
+```
+
+Will output the next version, e.g. `1.2.3`. This can be useful for determining the next version based in CI for non
```
Lee-W, line 297:
```suggestion
Will output the next version, e.g., `1.2.3`. This can be useful for determining the next version based on CI for non
```

tests/commands/test_bump_command.py
```diff
@@ -1462,3 +1463,63 @@ def test_bump_command_shows_description_when_use_help_option(
...
+
+    out, _ = capsys.readouterr()
+    assert "0.2.0" in out
+
+    tag_exists = git.tag_exist("0.2.0")
+    assert tag_exists is False
+
+
+@pytest.mark.usefixtures("tmp_commitizen_project")
+def test_bump_get_next__changelog_is_not_allowed(mocker: MockFixture):
```
Lee-W, line 1485: I feel we can combine these `not_allowed` test cases

Lee-W (approved):
I left a few nitpicks. non-blocking

Lee-W:
I'm planning on merging it these days. But please do let me know if you want to take a look 🙂  @woile @noirbizarre

commitizen/commands/bump.py
```diff
@@ -202,6 +207,18 @@ def __call__(self) -> None:  # noqa: C901
...
+        # changelog to be generated as well, this is the most intuitive solution
+        self.changelog = self.changelog or bool(self.changelog_to_stdout)
+
+        if get_next:
+            if self.changelog:
+                raise NotAllowed(
+                    "--changelog or --changelog-to-stdout is not allowed with --get-next"
+                )
+            # Setting dry_run to prevent any unwanted changes to the repo or files
+            self.dry_run = True
```
bearomorphism, line 220: Is `self.dry_run` used anywhere?

woile: https://github.com/commitizen-tools/commitizen/pull/1195/files#diff-920733205b591ccb6aded23a60aba1c20600036d38b5c224e27da47cf09ffb60L310

bearomorphism: `dry_run` and `self.dry_run` are different objects here.

I have tested in my local environment. The assignment is redundant and the `if dry_run:` path is never hit if `get_next` is True because `GetNextExit()` is raised earlier.

Issue #704, opened 2023-04-13
Commitizen-branch with pre-commit prevents from push to origin with exit code 23


vabatta:
### Description

Whenever I try to push a commit to the origin (`git push -u origin main`), an error is preventing me from pushing:
```
commitizen check branch..................................................Failed
- hook id: commitizen-branch
- exit code: 23

fatal: ambiguous argument 'origin/HEAD..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'


error: failed to push some refs to 'https://github.com/<some repo>.git'
```

### Steps to reproduce

1. Run `cz init`.
2. Run `pre-commit install --hook-type commit-msg --hook-type pre-push`
3. Run `git init && git add . && git commit -m "feat: my new commit"`
4. Run `git remote add origin https://github.com/<some repo>.git`
5. Run `git push -u origin main`

### Current behavior

I should be able to push it correctly to origin.

### Desired behavior

I want to push it to origin.

### Environment

```
Commitizen Version: 2.42.1
Python Version: 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110]
Operating System: Linux
```

vabatta:
If you are using dev containers, for me it worked by cleaning my local dev container and let it rebuild.

Corfucinas:
Any quick solution to this bug? (that is not `--no-verify`)

slyons:
I'm still seeing this as well. I added `cz` to an existing repo, but in a new branch that did not exist on `origin` yet, if that helps.

slyons:
For anyone still running into this, here's how you fix it:

```git remote set-head origin --auto```

For whatever reason, `HEAD` was not set in my repo. 

Before:
```shell
$ git branch -a
* dev/initial_structure
  main
  remotes/origin/dev/initial_structure
  remotes/origin/main
```

After:
```shell
$ git branch -a
* dev/initial_structure
  main
  remotes/origin/HEAD -> origin/main
  remotes/origin/dev/initial_structure
  remotes/origin/main
```

PR #1207, opened 2024-08-13
feat(commit): implement questions 'filter' support with handlers

AdrianDC:
## Description

Supported APIs: Common Python, commitizen.cz.utils.* functions

**Example YAML configurations:**
```yaml
commitizen:
  name: cz_customize
  customize:
    questions:
      - ...
      - type: input
        name: scope
        message: 'Scope of the change :'
        filter: 'lambda text: commitizen.cz.utils.required_validator(text, msg="! Error: Scope is required")'
        default: ''
      - type: input
        name: subject
        message: 'Title of the commit (starting in lower case and without period) :'
        filter: 'lambda text: commitizen.cz.utils.required_validator(text.strip(".").strip(), msg="! Error: Title is required")'
        default: ''
      - type: input
        name: body
        message: 'Additional contextual message (Empty to skip) :'
        default: 'Issue: #...'
        filter: 'commitizen.cz.utils.multiple_line_breaker'
```

Additional commits to add missing YAML tests, fix YAML configurations, and then add the keys for this feature.

## Checklist

## Expected behavior

```
? Title of the commit (starting in lower case and without period) :                                                                                                                                                                                    
Hello                                                                                                                                                                                                                                                  
! Error: Title is required
```

Example provided in the description

## Additional context

-


AdrianDC:
* Coverage resolved, 100% over commit.py
* TOML faulty types fixed
* Local containerized tests fixed for 'safe.directory' ownership issues
* Tests for 'cz_customize' using the `.customize` config rather than unrelated `cz_jira`
* Contents of the questions prompts tested for coverage and types around `commands.Commit`

commitizen/commands/commit.py
```diff
@@ -48,9 +51,23 @@ def read_backup_message(self) -> str | None:
...
+        global commitizen
+        import commitizen.cz.utils
+
+        for question in filter(
+            lambda q: isinstance(q.get("filter", None), str), questions
+        ):
+            question_filter = [
+                multiple_line_breaker(question["filter"].replace("\\n", "\n"))
+            ]
+            question["filter"] = eval("\n".join(question_filter))
```
Lee-W, line 69: Hmmm... I don't really like this kind of `eval` stuff. Would like to know how @woile and @noirbizarre  think

Lee-W (changes requested):

Lee-W (changes requested):
I'm sending a "Request changes" to this PR as I'm -1 to this but would like to get feedback from other maintainers

commitizen/commands/commit.py
AdrianDC: I'm not the biggest fan of `eval` either, it's a better choice than `exec` of course, was looking for an alternative
but nothing quite so capable of what we need it to do here to replicate these behaviours.

Otherwise, statically coded things you can call with a single string was an idea,
but is way too specific and limited to cover the simple cases needed here.

---

Makes me thing I'd like to restore my original `.cz.yaml` implementation of custom Conventional Commits,
before / without my evolutions and changes, to share it here for reference.

**To give an idea of my currently used `filter` keys :** https://gitlab.com/RadianDevCore/tools/pre-commit-crocodile/-/blob/47068c6bce3758119db692ff74d0de9370ae199c/.cz.yaml#L133

Lee-W: I think something like https://gitlab.com/RadianDevCore/tools/pre-commit-crocodile/-/blob/47068c6bce3758119db692ff74d0de9370ae199c/.cz.yaml#L147 is still a bit better than using `eval` (yep, we'll need to make it a python module), but would like to gather more ideas from others before we make any decision on this one

AdrianDC: That one is evaluated to the Python function too given it's all the same parser,
the initial alternative was `multiple_line_breaker`, for that one it's fine,
but it would need to somehow preexist for the other 2 for example, but if so and properly named, fine.

To sum up, no eval, but `multiple_line_breaker`, `required_validator_subject_strip` and `required_validator_scope` in `.utils`.
And strings of `if`, `elif`, `else Error` for bindings.

AdrianDC:
**Refactored as discussed : `feat(commit): implement questions 'filter' support with handlers`**

**Supported APIs:**
  - `multiple_line_breaker`
  - `required_validator`
  - `required_validator_scope`
  - `required_validator_subject_strip`
  - `required_validator_title_strip`

**Example YAML configurations:**

```yaml
commitizen:
  name: cz_customize
  customize:
    questions:
      - ...
      - type: input
        name: scope
        message: 'Scope of the change :'
        filter: 'required_validator_scope'
        default: ''
      - type: input
        name: subject
        message: 'Title of the commit (starting in lower case and without period) :'
        filter: 'required_validator_subject_strip'
        default: ''
      - type: input
        name: body
        message: 'Additional contextual message (Empty to skip) :'
        default: 'Issue: #...'
        filter: 'multiple_line_breaker'
```

AdrianDC:
Rebased :wink: .

Please cherry-pick at least the coverage improvements and minor fixes before the filter features review :+1: .

docs/customization.md
```diff
@@ -110,13 +110,13 @@ And the correspondent example for a yaml json file:
 commitizen:
   name: cz_customize
   customize:
-    message_template: "{{change_type}}:{% if show_message %} {{message}}{% endif %}"
+    message_template: '{{change_type}}:{% if show_message %} {{message}}{% endif %}'
```
Lee-W, line 113: May I know why do we want to change from " to ' here?

scripts/test
```diff
@@ -4,7 +4,7 @@ set -e
 export PREFIX='poetry run python -m '
 export REGEX='^(?![.]|venv).*'
 
-${PREFIX}pytest -n 3 --dist=loadfile --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen tests/
+${PREFIX}pytest -n 3 --dist=loadfile --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen "${@}" tests/
```
Lee-W, line 7: What it this for?

commitizen/cz/utils.py
```diff
@@ -6,13 +6,34 @@
 from commitizen.cz import exceptions
 
 
-def required_validator(answer, msg=None):
+def required_validator(answer: str, msg=None) -> str:
```
Lee-W, line 9: We probably would like to add test cases to validate these validators

Lee-W: I know we named it as `required_validator` but would be great if we change them to something like `require_validator`

commitizen/cz/utils.py
```diff
@@ -6,13 +6,34 @@
...
+
+
+def required_validator_title_strip(
+    answer: str,
+    msg: str = "! Error: Title is required",
+) -> str:
+    return required_validator(answer.strip(".").strip(), msg)
+
+
+def multiple_line_breaker(answer: str, sep: str = "|") -> str:
```
Lee-W, line 36:
```suggestion
def break_multiple_line(answer: str, sep: str = "|") -> str:
```

Lee-W (commented):
Sorry for taking so long. Just left some comments

scripts/test
AdrianDC: Usage was detailed in the commit message.

Now deprecated in favor to `poe cover -k test_initialize_cz_customize_failed`, will not be in the rebase

commitizen/cz/utils.py
AdrianDC: Done

docs/customization.md
AdrianDC: Uniformization and cleanups, updated against latest changes

commitizen/cz/utils.py
AdrianDC: I'm against renaming it from `required` to `require`, because on the users' side it translates to `required` fields, some of my users / colleagues already use it to create custom configurations, and should stay factual in my opinion.

Lee-W: Sounds good.

Lee-W:
I'll come back to this one after 4.8.3 release. As for another PR, I'll take a look at these 2 days. Sorry for the long waiting and thanks so much for still being around!

AdrianDC:
Rebased from last year's PR, to update towards [`4.9.1+adriandc.20250914`](https://github.com/AdrianDC/commitizen/commits/4.9.1%2Badriandc.20250914/)
for `pre-commit-crocodile` 8.0.0 (`prek` becoming default btw).

Feel free to edit the PR or cherry-pick at will.

---

I also implemented a `check-commitizen-branch` wrapper to the `commitizen-branch` hook,
handling `set-head` if needed, covering #704, also #593, but fixing upstream needs analysis.

---

_For the record, I'm not a fan of pinning to a patched & customized fork for so long,
blocking my users + coworkers on a specific version of commitizen-tools that I rebase 2-3 times a year,
including now v5-moved fixes and questions features, but works fine since August 2024._

PR #1206, opened 2024-08-13
fix(commit): resolve 'always_signoff' configuration and '-s' CLI issues

AdrianDC:
## Description

The `always_signoff` configuration or the CLI `-s` argument fail upon `git commit` call
due to the passed syntax being `git commit -- -s` rather than `git commit -s`.

The `--` is needed only on the commitizen CLI, and if no git argument is passed, the `--` should not be forced.

> signoff mechanic is deprecated, please use `cz commit -- -s` instead.
> fatal: /tmp/...: '/tmp/... is outside repository at '...'

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [ ] Update the documentation for the changes

## Expected behavior

* Warning and successful if `-s`
* Warning and successful if `-s --`
* No warning and successful if `-- -s`
* No warning and successful if `-- -s`
* No warning and successful with `always_signoff: true`
* No warning and successful if `--` with `always_signoff: true`
* No warning and successful if `-s` with `always_signoff: true`
* No warning and successful if `-s --` with `always_signoff: true`
* No warning and successful if `-- -s` with `always_signoff: true`

## Steps to Test This Pull Request

**.cz.yaml:**
```yaml
---
commitizen:

  # commitizen configurations
  always_signoff: true
```

* `cz c`
* `cz c --`
* `cz c -s`
* `cz c -s --`
* `cz c -- s`

## Additional context

Related to issue #1135 and #1146

**Tested with the `python:3.12` Docker image:** 
```bash
docker run -i --rm --entrypoint bash python:3.12 <<EOF
{
  git clone -b signoff https://github.com/AdrianDC/commitizen.git /tmp/commitizen
  cd /tmp/commitizen/
  pip install poetry
  poetry install
  ./scripts/format
  ./scripts/test
}
EOF
```

PR #1209, opened 2024-08-15
fix(pre-commit-hooks)!: allow empty ranges in 'commitizen-branch'

AdrianDC:
## Description

**fix(pre-commit-hooks)!: move 'check' to 'args' for global options**

Details: 'args: [...]' is not able to access '-nr NO_RAISE' global
         arguments for example from a arguments configuration

**fix(pre-commit-hooks)!: allow empty ranges in 'commitizen-branch'**

Details: Pushing the same history from 'develop' to 'staging' for
         example fails because range 'origin/HEAD..HEAD' is empty
> No commit found with range: 'origin/HEAD..HEAD'

## Checklist

## Expected behavior

Error `No commit found with range: 'origin/HEAD..HEAD'` not raised

Push same history to develop then staging

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -19,7 +19,7 @@
     the fact (e.g., pre-push or in CI) without an expensive check of the entire
     repository history.
   entry: cz
-  args: [check, --rev-range, origin/HEAD..HEAD]
+  args: [--no-raise, '3', check, --rev-range, origin/HEAD..HEAD]
   always_run: true
   pass_filenames: false
   language: python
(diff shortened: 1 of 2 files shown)


woile (commented):
@Lee-W thoughts on this? LGTM

Lee-W:
Looks good to me. but we probably won't want to bump a major version for it. so I would suggest we remove the ! in all the commits

AdrianDC:
Thanks for the feedback.

IMO this should be a ! change, as any user having `args: []` as I used to will have local and CI breakages.

Could go along other fixes such as #1206 and #1207

**Example local usage until released:** https://gitlab.com/RadianDevCore/tools/gcil/-/blob/53e6cb534ff77704ac9b9dc51d755b9965556910/.pre-commit-config.yaml#L54

Lee-W:
If that's the case, we'll need to wait for our next release. @woile we probably can create a v4 branch already?

Lee-W (approved):

Lee-W:
Just change the base to v4. and let's merge it 🚀

Issue #1194, opened 2024-07-29
Configuration flag to prevent bumping with uncommitted changes


ptoews:
### Description

When running `cz bump`, committizen will include all current uncommitted changes (even if not staged) in the bump commit.
This can lead to unexpected changes suddenly being in a new release.
There should be some kind of configuration option to either abort the bumping when there are uncommitted changes (similar to [bump-my-version's `allow_dirty`](https://callowayproject.github.io/bump-my-version/reference/configuration/#allow_dirty)) or stash the changes, update the version, commit, and pop the stash.

Lee-W:
@woile I feel we could change it to add version files only in https://github.com/commitizen-tools/commitizen/blob/01fd0426677b4db07988103d3a0ebd0d10763886/commitizen/commands/bump.py#L366. What do you think?

Lee-W:
@marcusalstrom I would vote for the latter (well, I'm biased. that's what I proposed).

would love to know how others think

woile:
I prefer having the `git.add`, I wonder what's the use case for not including all modifications?

If we asume commitizen runs on a CI, not on your working directory. Then you might do stuff like:

```sh
# modify a file not handled by commitizen but you want it as part of the release
cz bump
```

But I'm open for a non-breaking change

omidmme:
I think the case for not including all modifications is that unintended changes can go unnoticed. Would it in this case be better topass a flag to bump depending on which use-case you have to try and avoid any breaking changes?

bearomorphism:
Triage from #1964: This is a feature request, not a bug. There's no `allow_dirty`-style flag in commitizen. Suggesting we relabel as `type: feature`. Implementation idea: a `--allow-dirty` CLI flag + `allow_dirty` config setting; when false (recommended default in v5?), error out before staging if `git status --porcelain` is non-empty (excluding `version_files` and the changelog).

PR #1261, opened 2024-10-15
fix(bump)!: only add version files and staged changes

marcusalstrom:
## Description
Bump will now only add and commit version files along with any already committed changes.  Any unstaged changes will remain unstaged. 

NOTE: BREAKING CHANGE, since we added a new abstract method to VersionProvider any custom version provider will break.  (open to any suggestions on how to avoid this)

## Checklist

## Expected behavior
Only version files and already staged changes will be committed on bump. 

1. cz bump with unstaged changes
2. only version files (and any staged changes) will be added and committed, unstaged changes will remain unstaged

## Additional context
References issue #1194

### DIFF
--- a/commitizen/providers/base_provider.py
+++ b/commitizen/providers/base_provider.py
@@ -34,6 +34,12 @@ class VersionProvider(ABC):
         Set the new current version
         """
 
+    @abstractmethod
+    def get_files(self) -> list[str]:
+        """
+        Get the version files
+        """
+
 
 class FileProvider(VersionProvider):
     """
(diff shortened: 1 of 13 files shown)


Lee-W:
As this is breaking change, I change the base branch to v4

commitizen/providers/base_provider.py
```diff
@@ -34,6 +34,12 @@ def set_version(self, version: str):
         Set the new current version
         """
 
+    @abstractmethod
+    def get_files(self) -> list[str]:
```
Lee-W, line 38: let's make it a property

Lee-W: maybe renamed as `version_files` would be better?

Lee-W:
In general, I like this idea. @woile @noirbizarre would love to know your thought 🙂

noirbizarre:
No unexpected extra file in release, I like this idea too, but it should include the changelog too if the option is set.

If it was possible to bump a dirty working copy, this is obviously a breaking change, and we should provide a way around this. Maybe a setting and param `--allow-changed-files` as `version files` are parsed for version and updated and some may have a process in which they stage their change to be taken in the bump without wanting version updated in this file

PR #1298, opened 2024-11-26
build(python)!: Drop support for Python 3.8 which reached EOL and add expose support for 3.13

noirbizarre:
## Description

Drop support for Python 3.8 which reached End Of Life (cf. https://devguide.python.org/versions/)
Expose support for Python 3.13 which is already tested in CI but not in the package metadata.

## Checklist

## Expected behavior

- Not functional change expected
- New syntaxes unlocked

CI should be sufficient

## Additional context
N/A


Lee-W (approved):

Lee-W:
oops. it seems this bump major version 😱

@woile @noirbizarre should we revert this and merge this to v4 instead?

Lee-W:
or we can just make the `v4` branch `v5`. sorry for not noticing it 🤦‍♂️

woile:
Let's leave it, technically this is proper semver 😅

PR #1301, opened 2024-11-29
fix(get-next-bump): fix to permit usage of --get-next option even when update_changelog_on_bump is set to true

Lowaiz:
## Description

With the configuration options `update_changelog_on_bump` set to `True`, the command `cz bump --get-next` crashed, due to the mutual exclusive options `--get-next` and `--changelog`/`--changelog-to-stdout`. 

Changing this behavior to still throw a conflict when using both CLI options, but let `--get-next` take the precedence over `update_changelog_on_bump` configuration.

## Checklist

## Expected behavior

Now the configuration options `update_changelog_on_bump` can be set to whatever you want when using the `cz bump --get-next` command.

1. Build a commitizen config with `update_changelog_on_bump` flag set to true.
2. Add some commits to the target test project.
3. Run `cz bump --get-next`
4. The command should work and not throw an error.

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -66,9 +66,8 @@ class Bump:
             },
         }
         self.cz = factory.commiter_factory(self.config)
-        self.changelog = arguments["changelog"] or self.config.settings.get(
-            "update_changelog_on_bump"
-        )
+        self.changelog_flag = arguments["changelog"]
+        self.changelog_config = self.config.settings.get("update_changelog_on_bump")
         self.changelog_to_stdout = arguments["changelog_to_stdout"]
         self.git_output_to_stderr = arguments["git_output_to_stderr"]
         self.no_verify = arguments["no_verify"]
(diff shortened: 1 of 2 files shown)


commitizen/commands/bump.py
```diff
@@ -207,17 +206,24 @@ def __call__(self) -> None:  # noqa: C901
...
                 raise NotAllowed(
                     "--changelog or --changelog-to-stdout is not allowed with --get-next"
                 )
+            # --get-next is a special case, taking precedence over config for 'update_changelog_on_bump'
+            self.changelog_config = False
             # Setting dry_run to prevent any unwanted changes to the repo or files
             self.dry_run = True
+        else:
+            # If user specified changelog_to_stdout, they probably want the
+            # changelog to be generated as well, this is the most intuitive solution
```
Lee-W, line 221: Hmmm... I think otherwise. If the user wants it to be printed, they might want to take a look first before making the change

Lowaiz: This comment and the associated self.changelog_flag = (self.changelog_flag or bool(self.changelog_to_stdout) or self.changelog_config) is from the L210-L212 of the previous code. (with the inclusion of the new `changelog_config` change)

Lowaiz: >  If a user wants to print something, they may want to review it first before making any changes.

And, from what I understand from the the rest of the `bump` code, using `changelog_to_stdout` flag essentially put the changelog generation in dry run. So it will not change anyfile (my MR doesn't seem to change that behavior I guess)

Lowaiz:
@Lee-W @noirbizarre @woile 
Hello,
Can one of you review this PR ?
As I did not receive any update on my response to the comment, I try the direct ping, sorry for the inconvenience  !

Lee-W:
> @Lee-W @noirbizarre @woile Hello, Can one of you review this PR ? As I did not receive any update on my response to the comment, I try the direct ping, sorry for the inconvenience !

Will take a look these days. Thanks for reminding me!

Lee-W (approved):

PR #1327, opened 2025-01-02
Multi-Language Version

Vlor999:
## Description
Sorry for the last pull request, it was a mistake. The idea is the people may want to use commitizen to make commits but some people may not speak English. This is where the idea came to introduce a language option. The idea is just to give the opportunity to users to have better understanding of the selection of possibility. 

## Checklist

## Expected behavior
This new feature must increase the understanding and improve commits for users.

1. poetry shell
2. python3 commitizen/__main__.py -language fr commit
3. python3 commitizen/__main__.py commit
4. python3 commitizen/__main__.py -language en commit
5. exit

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -108,6 +108,12 @@ data = {
             "required": False,
             "help": "comma separated error codes that won't rise error, e.g: cz -nr 1,2,3 bump. See codes at https://commitizen-tools.github.io/commitizen/exit_codes/",
         },
+        {
+            "name": ["-language"],
+            "type": str,
+            "default": "en",
+            "help": "language of the commit message (default: en). Available languages: enGLISH, frENCH. others are welcome",
+        },
     ],
     "subcommands": {
         "title": "commands",
--- a/commitizen/cz/conventional_commits/.cache_multilanguage.txt
+++ b/commitizen/cz/conventional_commits/.cache_multilanguage.txt
@@ -1,0 +1,12 @@
+prefix_en=Select the type of change you are committing
+fix_en=A bug fix. Correlates with PATCH in SemVer
+feat_en=A new feature. Correlates with MINOR in SemVer
+docs_en=Documentation only changes
+style_en=Changes that do not affect the meaning of the code (white-space, formatting,  missing semi-colons, etc)
+refactor_en=A code change that neither fixes a bug nor adds a feature
+perf_en=A code change that improves performance
+test_en=Adding missing or correcting existing tests
+build_en=Changes that affect the build system or external dependencies (example scopes: pip, docker, npm)
+ci_en=Changes to CI configuration files and scripts (example scopes: GitLabCI)
+scope_en=What is the scope of this change? (class or file name): (press [enter] to skip)
+subject_en=Write a short and imperative summary of the code changes: (lower case and no period)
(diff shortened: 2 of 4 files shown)


woile:
Looks like a nice feature. Where's the french translation? How would people provide custom translations for their own rules, like [cz-emoji](https://github.com/adam-grant-hendry/cz-emoji)?

Thoughts @Lee-W @noirbizarre ?

Vlor999:
> Looks like a nice feature. Where's the french translation? How would people provide custom translations for their own rules, like [cz-emoji](https://github.com/adam-grant-hendry/cz-emoji)?
> 
> Thoughts @Lee-W @noirbizarre ?

You can choose your translation on the command line. If it merge by typing : `cz -language fr commit`. With this sentence you will have the explanation in French, if you want other language, you can change the `fr` and if you want to stay with the English version you do not need to write `-language`. For the cz-emoji version, it could be a bit hard because my way of doing is to use a translator API. So if you want to make the cz-emoji, you have to add the datas/sentences into de cache-memories. Sorry if there is some mistakes in my code, I can already see that there is some issue about the python-check, this is my first contribution to an open source projects. I'm here to help you make this feature available if you wish so.

noirbizarre:
That's very interesting! I worked in French administrations and some teams/projects have french as a unique and mandatory language, so I can definitively see the use cases!

I worked a lot with translations and making them opt-in, including for extension is straightforward with https://docs.python.org/fr/3/library/gettext.html

The idea would be:
- keep all `name` keys in english
- call `gettext` on display on the key:
  - either there is a translation provided, then it is displayed instead of the english one
  - either there is none, and the english one is displayed
- extensions/plugins can provide their translation using a shared convention (often `translations/{iso}/messages.po`)

`gettext` has many advantages over `json` as it handles complex logic like pluralization (in Python, given in some other languages there are libs offering this kind of features on JSON).

It is also a good idea to bind the default on `LANG` env var instead of `en` as it will fallback on `en` if the language does not exists.

To help with translations, translation services like transifex offer a free tier for open-source projects, but https://weblate.org is open source itself and free for open-source project.

For emojis, this is a special case which will require some design choices, but given it is already using the [emoji](https://pypi.org/project/emoji/) library which is already supporting localization, it won't be that difficult to handle.

Anyway, I think this feature would be great and I am ready to help on this one if required.

Lee-W:
In general, I like the idea, but we probably have a better tool for dealing with this i18n thing. Instead of doing it ourselves, we probably should delegate the logic to those tools. what we need to do is providing the script for each of the languages

Lee-W:
Hi @Vlor999 do we still need this one? or should we review https://github.com/commitizen-tools/commitizen/pull/1328/files instead?

Vlor999:
Yes, it is more interesting to look for the last commit.ENSIMAGLe 26 janv. 2025 à 11:54, Wei Lee ***@***.***> a écrit :﻿
Hi @Vlor999 do we still need this one? or should we review https://github.com/commitizen-tools/commitizen/pull/1328/files instead?

Vlor999:
@noirbizarre Hello, I'm currently a student and searching for an internship in France. I know this is absolutely not the place but if you are interesting, please contact me on my mail. : 
willem.adnet@grenoble-inp.org

PR #1345, opened 2025-01-30
Improve hooks type annotation and fix missing link

Lee-W:
## Description

* docs(hooks): fix missing link
* style(hooks): improve type annotation

## Checklist

## Expected behavior

Everything should still be the same

None

### DIFF
--- a/docs/tutorials/auto_prepare_commit_message.md
+++ b/docs/tutorials/auto_prepare_commit_message.md
@@ -25,7 +25,7 @@ commitizen and use the generated commit message for the commit.
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


noirbizarre (approved):
👍🏼 

Note for later:
- we should also provide them as pre-commit hook
- maybe we should have a `cz hook install` or something like that to avoid the manual fetch with curl

Lee-W:
> 👍🏼
> 
> Note for later:
> 
>     * we should also provide them as pre-commit hook
> 
>     * maybe we should have a `cz hook install` or something like that to avoid the manual fetch with curl

Yep, I was thinking about that as well (but not yet have time 🥲). added an issue https://github.com/commitizen-tools/commitizen/issues/1355 to track this

Issue #1349, opened 2025-01-31
Add uv provider


Lee-W:
### Description

support updating both `pyprojec.toml` and `uv.lock`

woile:
Should we extend `pep440` to detect the presence of `uv.lock`?

Lee-W:
> Should we extend `pep440` to detect the presence of `uv.lock`?

yep, sounds like a good idea!

lsorber:
We're using Commitizen for [Substrate](https://github.com/superlinear-ai/substrate) and we noticed that you cannot just run `cz bump` with a uv project because bumping the version number causes `uv.lock` to be out of sync with `pyproject.toml`.

Until Commitizen supports `uv.lock`, would you recommend adding `uv lock --offline` to the [pre_bump_hooks](https://commitizen-tools.github.io/commitizen/commands/bump/#pre_bump_hooks) as a workaround?

khrapovs:
Thank you so much for this, I've been waiting for it! But unfortunately you have forgotten to document it for the users like me ;) I think this feature deserves to be mentioned in https://commitizen-tools.github.io/commitizen/config/#version-providers :)

Lee-W:
> Thank you so much for this, I've been waiting for it! But unfortunately you have forgotten to document it for the users like me ;) I think this feature deserves to be mentioned in https://commitizen-tools.github.io/commitizen/config/#version-providers :)

yep, miss this part 🤦‍♂️. just sent a PR for it
https://github.com/commitizen-tools/commitizen/pull/1362

PR #1351, opened 2025-01-31
feat(providers): add uv_provider

Lee-W:
closes: #1349

## Description

as title

## Checklist

## Expected behavior

uv.lock should also updated when running `cz bump`

```shell
uv init
uv add commitizen
cz init
git add .
cz c
cz bump
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,7 @@ npm = "commitizen.providers:NpmProvider"
 pep621 = "commitizen.providers:Pep621Provider"
 poetry = "commitizen.providers:PoetryProvider"
 scm = "commitizen.providers:ScmProvider"
+uv = "commitizen.providers:UvProvider"
 
 [project.entry-points."commitizen.scheme"]
 pep440 = "commitizen.version_schemes:Pep440"
@@ -165,9 +166,7 @@ omit = [
 
 [tool.pytest.ini_options]
 addopts = "--strict-markers"
-testpaths = [
-    "tests/",
-]
+testpaths = ["tests/"]
 
 [tool.tox]
 requires = ["tox>=4.22"]
@@ -178,7 +177,7 @@ description = "Run tests suite against Python {base_python}"
 skip_install = true
 deps = ["poetry>=2.0"]
 commands_pre = [["poetry", "install", "--only", "main,test"]]
-commands = [["pytest", { replace = "posargs", extend = true}]]
+commands = [["pytest", { replace = "posargs", extend = true }]]
 
 [tool.ruff]
 line-length = 88
@@ -227,14 +226,14 @@ poetry_command = ""
 [tool.poe.tasks]
 format.help = "Format the code"
 format.sequence = [
-    {cmd = "ruff check --fix commitizen tests"},
-    {cmd = "ruff format commitizen tests"},
+    { cmd = "ruff check --fix commitizen tests" },
+    { cmd = "ruff format commitizen tests" },
 ]
 
 lint.help = "Lint the code"
 lint.sequence = [
-    {cmd = "ruff check commitizen/ tests/ --fix"},
-    {cmd = "mypy commitizen/ tests/"},
+    { cmd = "ruff check commitizen/ tests/ --fix" },
+    { cmd = "mypy commitizen/ tests/" },
 ]
 
 check-commit.help = "Check the commit message"
@@ -254,7 +253,7 @@ all.sequence = [
     "format",
     "lint",
     "cover",
-    "check-commit"
+    "check-commit",
 ]
 
 "doc:screenshots".help = "Render documentation screeenshots"
@@ -268,10 +267,10 @@ doc.cmd = "mkdocs serve"
 
 ci.help = "Run all tasks in CI"
 ci.sequence = [
-    {cmd="pre-commit run --all-files"},
+    { cmd = "pre-commit run --all-files" },
     "cover",
 ]
-ci.env = {SKIP = "no-commit-to-branch"}
+ci.env = { SKIP = "no-commit-to-branch" }
 
 setup-pre-commit.help = "Install pre-commit hooks"
 setup-pre-commit.cmd = "pre-commit install"
--- a/tests/providers/test_uv_provider.py
+++ b/tests/providers/test_uv_provider.py
@@ -1,0 +1,97 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from commitizen.config.base_config import BaseConfig
+from commitizen.providers import get_provider
+from commitizen.providers.uv_provider import UvProvider
+
+if TYPE_CHECKING:
+    from pytest_regressions.file_regression import FileRegressionFixture
+
+
+PYPROJECT_TOML = """
+[project]
+name = "test-uv"
+version = "4.2.1"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.13"
+dependencies = ["commitizen==4.2.1"]
+"""
+
+UV_LOCK_SIMPLIFIED = """
+version = 1
+revision = 1
+requires-python = ">=3.13"
+
+[[package]]
+name = "commitizen"
+version = "4.2.1"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "argcomplete" },
+    { name = "charset-normalizer" },
+    { name = "colorama" },
+    { name = "decli" },
+    { name = "jinja2" },
+    { name = "packaging" },
+    { name = "pyyaml" },
+    { name = "questionary" },
+    { name = "termcolor" },
+    { name = "tomlkit" },
+]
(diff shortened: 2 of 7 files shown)


tests/providers/test_uv_provider.py
```diff
@@ -0,0 +1,97 @@
...
+[project]
+name = "test-uv"
+version = "4.2.1"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.13"
+dependencies = ["commitizen==4.2.1"]
+"""
+
+UV_LOCK_SIMPLIFIED = """
```
Lee-W, line 23: Somehow I'm not able to extract this as a separate file. pytest-xdist fails.

woile (approved):
It looks good, the only issue is that it doesn't cover workspace.
```
[tool.uv.workspace]
members = ["packages/*"]
exclude = ["packages/seeds"]
```

Lee-W:
> It looks good, the only issue is that it doesn't cover workspace.
> 
> ```
> [tool.uv.workspace]
> members = ["packages/*"]
> exclude = ["packages/seeds"]
> ```

I'm a bit confused why do we need to cover workspace (I've never used this feature so I might miss some context 🤦‍♂️)

woile:
It allows to have multiple python projects, as a monorepo. Each package *probably* (as in, I think it works quite similar to cargo) will add an entry to the lock. I'm working on a fix for cargo. If it works well we can port it to `uv`

pyproject.toml
```diff
@@ -250,12 +249,7 @@ cover.help = "Run the test suite with coverage"
 cover.ref = "test --cov-report term-missing --cov-report=xml:coverage.xml --cov=commitizen"
 
 all.help = "Run all tasks"
-all.sequence = [
-    "format",
-    "lint",
-    "cover",
-    "check-commit"
-]
+all.sequence = ["format", "lint", "cover", "check-commit"]
```
noirbizarre, line 252: I tend to avoid 1-liner list as in the case of addition, change or removal, the entire line is changed, while the single item by line allows to precisely see what has been added, changed or removed.
To avoid being reformatted, add the missing trailing comma:
```suggestion
all.sequence = [
    "format",
    "lint",
    "cover",
    "check-commit",
]
```

commitizen/commands/init.py
```diff
@@ -260,7 +264,7 @@ def _ask_version_provider(self) -> str:
     def _ask_version_scheme(self) -> str:
         """Ask for setting: version_scheme"""
         default = "semver"
-        if self.project_info.is_python:
+        if self.project_info.is_python or self.project_info.has_uv_lock:
```
noirbizarre, line 267: Is there any case supported by `uv` where a `uv.lock` may exists without a `pyproject.toml`?
It seems that is currently unsupported: https://github.com/astral-sh/uv/issues/6722#issuecomment-2315254679
But I might be missing something

pyproject.toml
```diff
@@ -267,11 +261,8 @@ doc.help = "Live documentation server"
 doc.cmd = "mkdocs serve"
 
 ci.help = "Run all tasks in CI"
-ci.sequence = [
-    {cmd="pre-commit run --all-files"},
-    "cover",
-]
-ci.env = {SKIP = "no-commit-to-branch"}
+ci.sequence = [{ cmd = "pre-commit run --all-files" }, "cover"]
```
noirbizarre, line 264: Same, it's harder to read the sequence and will produce harder to read diff in case of change:
```suggestion
ci.sequence = [
    { cmd = "pre-commit run --all-files" }, 
    "cover",
]
```

noirbizarre (commented):
Seems good to me but I wonder why for `uv` we handle the lock while we don't do it for poetry nor any other python provider.
I feel this is introducing a `uv` special treatment that could already be handled by relying on the generic PEP621 support we already have and using [`post_bump_hooks`](https://commitizen-tools.github.io/commitizen/commands/bump/#post_bump_hooks)
> [!NOTE]
> Maybe the `pyproject.toml` formatting should be splitted in a `style` commit as unrelated

noirbizarre:
I investigated a bit, and neither `pdm` nor `poetry` lock the root package itself in the lock which, to me, makes sense.
It is a weird choice from `uv` to lock the package itself, including the version, which doesn't have any sense outside pypi publication, but at lest I understand why this special treatment.

Given I always use dynamic version with `scm`, I need to check how `uv` handles it (cf. https://docs.astral.sh/uv/concepts/cache/#dynamic-metadata 
It is possible that using `scm` with `uv` requires the same lock update and then it might be more interesting to provide this as a reusable hook for both `toml` and `scm` provider.

WDYT ?

noirbizarre:
Also, as it has been a recurrent issue working with some lock files, I would add lock file version check (cf. https://docs.astral.sh/uv/concepts/resolution/#lockfile-versioning).
This would allow to:
- add support for future lockfile versioning while continuing to support previous format and not having commitizen version tied to `uv` version and its lock versioning
- warn when encountering an unsupported lock version, instead of failing or corrupting it

woile:
`uv` copies cargo. I think it's because you may have a repo with multiple
packages (in any of both).

We could just use pep621 and detect all the different lock files, or we
could be explicit, like in this PR, I like it like this

noirbizarre:
But then it means that to have `scm` support with `uv`, we will need to have a separated `UvScmProvider` and then for each combination of provider + known lock file we would need a dedicated class.

My point does not imply that we shouldn't have explicit support for this (I am all for both explicitness and `uv.lock` support) but that after reviewing it, I don't think this should be a dedicated version provider but some kind of reusable `pre_bump_hook`.
It would still be explicit, but it would also work with the `scm` provider.
Plus, this is only true when the package is self installed, which is not always the case.
The `uv.lock` file looks more like a special `version_file` to handle and after thinking a lot about this, I believe having support for "special" files hook would be a great addition.

The way I see it, instead of having:
```toml
version_provider = "uv"
``` 
we should have something like:
```toml
pre_bump_hook = [
   "<hooks prefix>:uv.lock",
]
```
or 
```toml
pre_bump_hook = [
   "canonical.path.to.hook:uvlock",
]
```
or 
```toml
pre_bump_hook = [
   "canonical.path.to.hooks:lock('uv.lock')",
]
```

I see plenty of possibilities with this approach because sometimes, the `version_files` handling is not enough, and it would allow us and users to provide reusable complex files handling.

WDYT ?

pyproject.toml
Lee-W: opps, didn't notice my editor format it for me...

pyproject.toml
Lee-W: right... `,` is better

Lee-W:
> It allows to have multiple python projects, as a monorepo. Each package _probably_ (as in, I think it works quite similar to cargo) will add an entry to the lock. I'm working on a fix for cargo. If it works well we can port it to `uv`

Got it. Yep, maybe we can do it in another PR, and I'll spend some time to look into how it works

Lee-W:
I only have time to browse through the comments. But @noirbizarre 's seems to be a good alternative 🤔 I might need some time to think through it again. But I'll be out for the next 2 weeks 🤦‍♂️ We can merge it if we want this in earlier. Otherwise, I'll continue to work on it after I'm back. Thanks!

woile:
We might have to think this through a bit better. 

On one hand, I'd like things to just work. I don't want users to have to configure many different chunks when using something like `uv`, it should just work, and not fail the next CI pipeline. 

On the other hand, we have `scm` where the user chooses to use the version from git. I think here the user should update all the relevant files using the option `version_files`, but then we have the issue with multi-line files, specially lock files.

Right now, our provider `cargo` is also broken, because the lock is not being updated, and the same seems to be the case for `uv`.

To me, `uv`'s lock file is not special, packages managers, which I usually never run issues with (unlike poetry), have the package version in the lock itself (I don't know if it's related or not to the lock though, so take it with a pinch of salt):
- npm (js)
- cargo (rust)
- uv (python)

So I'd say it's a pretty standard feature.

Could we, perhaps, have both, a working `uv` and potentially "hooks"? 
Compose the `uv` provider with "chunks", which an advanced user could use to compose its own "provider"? Is it too much? WDYT?

---

The discussion for [PEP751](https://peps.python.org/pep-0751/) seems to lean towards *not* including the package version in the lockfile (can someone check that I understood properly?): https://discuss.python.org/t/pep-751-one-last-time/77293/68
I think it will be *optional* for auditing purposes, [see pep](https://peps.python.org/pep-0751/#drop-recording-the-package-version)

noirbizarre:
Agreed, I like this direction this is going.
I have no problem with merging the `uv` provider until we come with a more "standard" and scalable way to handle such kind of structured file formats.

noirbizarre (approved):
LGTM until we have file-specific hooks 🚀

Lee-W:
Ok, then I'll probably merge this later today. Another thing I'm thinking of is multiple provider? not sure whether it makes sense 🤔

noirbizarre:
I think it is easier to stick with the 1 provider only for the version source but support for multiple hooks for specific files.
The main purpose of the Version provider is to designate the source of trust for the version and extract it, setting it back is optional (like in `scm` where it is never set).
As a consequence, having multiple providers seems a bit complicated to handle and a big source of complexity.
However, this is already the purpose of `version_files` to spread the new version in multiple files as a side effect of bump.
We could say that we restrict version provider to read and handle all writes as `version_files` but I don't think it is necessary and would add unnecessary verbosity and redundancy.

commitizen/commands/init.py
Lee-W: AFAIK, no. 🤔 yep we probably could just remove this check. Let me update it now

PR #1354, opened 2025-02-08
Proposal: Dev & CI workflow improvement

noirbizarre:
## Description

This PR is a proposal improvement on the dev & CI workflows:
- Use `poethepoet` as plugin to run tasks
- Use the same tasks in CI
- Use the same dependencies groups in CI
- Only install required dependency groups in each CI jobs
- Add `tox` support to easily test against all supported Python version
- Removes redundant `pre-push` hooks as they already run on commit

On `poetry install`, `poethepoet` is installed locally as a plugin.
Tasks are exposed both as:
- poetry sub commands
- poe tasks

You can run:

```shell
poetry test
poe test
```

Tasks accept extra parameters, so it's now possible to run tests with queries:

```shell
poetry test -k bump
```

So are tox tests (run in parallel)

```shell
poetry test:all -k bump
```
To run the same thing as in GitHub action, run `poetry ci`

You can also preview the documentation with `poetry doc`

### Available tasks
 
image

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `./scripts/format` and `./scripts/test` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes

## Expected behavior

- no GitHub actions regression but slighly faster jobs (less dependencies to install)
- easier onboarding and contribution (all commands are discoverable and accept extra parameters)
- easier contribution (all commands accept extra parameters to fit everyone)
- less temptation to commit with `--no-verify` (pre-commit hooks are lighter and don't prevent push wip branches or in case of flapping test)

## Steps to Test This Pull Request
1. Checkout this PR
2. `poetry install`
3. List tasks with `poetry list` or `poe` (`poetry run poe` if you don't have it globally installed)
4. Try any command

## Additional context

Fix #724

> [!NOTE]
> It makes possible to have the documentation screenshots easily generated on the fly in both GHA workflow and `poetry doc` command and so, not having commit them and to store them in the repo

PR #1357, opened 2025-02-11
ci(github-actions): add check-commit task

Lee-W:
## Description

In https://github.com/commitizen-tools/commitizen/pull/1354, commit check was missed. This PR is to add it back

## Checklist

## Expected behavior

invalid commit should raise an error when running `poetry commit-check`

1. randomly create a commit that does not follow conventional commit
2. `poetry commit-check`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -237,6 +237,9 @@ lint.sequence = [
     {cmd = "mypy commitizen/ tests/"},
 ]
 
+check-commit.help = "Check the commit message"
+check-commit.cmd = "cz -nr 3 check --rev-range origin/master.."
+
 test.help = "Run the test suite"
 test.cmd = "pytest -n 3 --dist=loadfile"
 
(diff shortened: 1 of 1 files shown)


Lee-W:
@noirbizarre could you please take a quick look when you have some bandwidth 🙂 This would prevent us from merging PR with wrong commit format accidentally

noirbizarre (approved):

noirbizarre:
Note that I thought it was already done by commitizen pre-commit hook:https://github.com/commitizen-tools/commitizen/blob/master/.pre-commit-config.yaml#L50-L56

Lee-W:
> Note that I thought it was already done by commitizen pre-commit hook:https://github.com/commitizen-tools/commitizen/blob/master/.pre-commit-config.yaml#L50-L56 (and if not, it might be interesting to add this hook)

It actually is, but it's also quite common for a new contributor to miss this step (and us to accidentally merge it) so better to add an additional safe guard I think 🤔

PR #1358, opened 2025-02-17
fix(bump): manual version bump if prerelease offset is configured

ckagerer:
## Description

If a prerelease offset is configured (I introduced this about 2 years ago with PR #614), no bump is possible with the manual version.
To do this, the configuration must first be changed and then reset after the bump.

For this reason, it makes more sense to remove the check that prevents the manual bump.

## Checklist

	- I'm not sure where best to cover this case. Please give me a hint in which existing test file I could extend this.
- [x] ~Update the documentation for the changes~

## Expected behavior

Bumping a manual version works without changing the configuration file.

1. Use the follwoing Configuration
```toml
[tool.commitizen]
changelog_incremental = true
tag_format = "v$version"
version = "1.2.0b13"
prerelease_offset = 1
```
2. `cz  bump 9.8.7`

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -187,11 +187,6 @@ class Bump:
                     "--major-version-zero cannot be combined with MANUAL_VERSION"
                 )
 
-            if prerelease_offset:
-                raise NotAllowed(
-                    "--prerelease-offset cannot be combined with MANUAL_VERSION"
-                )
-
             if get_next:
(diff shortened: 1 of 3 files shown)


Lee-W (approved):
Thanks @ckagerer ! 

@woile @noirbizarre This PR looks good to me. I'm going to merge it this days if there's no objection

woile:
LGTM

PR #1364, opened 2025-03-02
fix(tags): fixes ImportError on Python >=3.11 (#1363)

jakob-keller:
## Description
Fixes #1363

## Checklist

- [ ] Add test cases to all the changes you introduce
- [ ] Run `poetry all` locally to ensure this change passes linter check and test
- [ ] Test the changes on the local machine manually
- [ ] Update the documentation for the changes

PR #1365, opened 2025-03-02
fix(tags): import `Self` from `typing` for Python 3.11+

noirbizarre:
## Description
`typing-extensions` is not installed for Python 3.11+, so fix the import properly.

> [!Note]
> CI was not failing as `typing-extensions` is a transitive dependency of multiple test/lint dependencies but is missing as soon as `commitizen` is installed alone.

## Checklist

## Expected behavior

`commitizen` import/run does not fail on `Self`git s

## Additional context
I might merge this PR without review as soon as the CI pass as the latest release is broken for Python 3.11+

### DIFF
--- a/commitizen/tags.py
+++ b/commitizen/tags.py
@@ -1,14 +1,13 @@
 from __future__ import annotations
 
 import re
+import sys
 import warnings
 from collections.abc import Sequence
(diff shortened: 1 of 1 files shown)


noirbizarre:
Duplicate #1364

PR #1375, opened 2025-03-21
fix: display invalid tag information

carlossg:
and the regex used for validation

## Description

Starting in 4.4.0 due to #1297 when an invalid tag is present an exception is thrown. I would expect the tag is ignored.

This PR only improves the message, not the root cause

Before

```
packaging.version.InvalidVersion: Invalid version tag
```

After

```
packaging.version.InvalidVersion: Invalid version tag: '2.2.44.aa00000' does not match any configured tag format: re.compile('v?(?P<version>([0-9]+)\\.([0-9]+)(?:\\.([0-9]+))?(?:-([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?(?:\\+[0-9A-Za-z.]+)?(\\w+)?)')
```

## Checklist

1. create a non semver tag, ie. 2.2.44.aa00000
2. run `cz bump --dry-run`

### DIFF
--- a/commitizen/tags.py
+++ b/commitizen/tags.py
@@ -146,7 +146,9 @@ class TagRules:
             m for regex in self.version_regexes if (m := regex.fullmatch(tag.name))
         )
         if not (m := next(candidates, None)):
-            raise InvalidVersion()
+            raise InvalidVersion(
+                f"Invalid version tag: '{tag.name}' does not match any configured tag format"
+            )
         if "version" in m.groupdict():
             return self.scheme(m.group("version"))
 


commitizen/tags.py
```diff
@@ -146,7 +146,8 @@ def extract_version(self, tag: GitTag) -> Version:
             m for regex in self.version_regexes if (m := regex.fullmatch(tag.name))
         )
         if not (m := next(candidates, None)):
-            raise InvalidVersion()
+            regexes = ", ".join([str(r) for r in self.version_regexes])
+            raise InvalidVersion(f"Invalid version tag: '{tag.name}' does not match any configured tag format: {regexes}")
```
noirbizarre, line 150: Maybe something shorter, without the regexp:
```suggestion
            raise InvalidVersion(f"Invalid version tag: '{tag.name}' does not match any configured tag format")
```

With many regexps or when you have many tags that don't match (or worse if you have both many patterns and many unmatched tags), it will break the terminal output making it harder to read and defeating the purpose of this PR.

noirbizarre (changes requested):
I think this would be a good addition, but without the regexps in the message as it will easily flood the output.

Note that #1297 should not have introduced any new raised exception, as it was already raised before and it was the reason I made #1297 initially, PR to provide a workaround to this exception. So I am also interested in finding why it wasn't raised before for you and now it is.

Lee-W (approved):
LGTM. Just fix the style. Would be nice if we could add a unit test. But I'm also good with merge it as it is

Lee-W:
> I think there is already a test for the exception itself, so I am OK to have the wording changed without a new test.
> 
> I'll take a look at the uncaught exception as soon as I have more time.

Then that's merge it and we can get back a bit later

PR #1376, opened 2025-03-21
feat(changelog): expose commit parents

unleashed:
## Description

Expose commit parent(s) digests to the changelog process. This has an application
for projects that want to generate changelogs based on merge commits only, but it
is likely useful in more contexts, for example listing commits that belong to the same
PR under its own section.

## Checklist

## Expected behavior

The changelog generation code exposes the `parents` field with a list of parent
digests in git commits.

You can create a changelog template iterating the `parents` field of changes and
for merge commits you should have at least 2 parents.

## Additional context

Useful in the git workflow I'm currently working on, where merge commits cannot
be told apart by the changelog generation machinery and I need to distinguish them
as they are the only ones that can be taken into account without rewriting the project's
git history due to using varied standards throughout its life.

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -44,13 +44,20 @@ class GitObject:
 
 class GitCommit(GitObject):
     def __init__(
-        self, rev, title, body: str = "", author: str = "", author_email: str = ""
+        self,
+        rev,
+        title,
+        body: str = "",
+        author: str = "",
+        author_email: str = "",
+        parents: list[str] | None = None,
     ):
         self.rev = rev.strip()
         self.title = title.strip()
         self.body = body.strip()
         self.author = author.strip()
         self.author_email = author_email.strip()
+        self.parents = parents or []
 
     @property
     def message(self):
@@ -137,7 +144,9 @@ def get_commits(
     for rev_and_commit in git_log_entries:
         if not rev_and_commit:
             continue
-        rev, title, author, author_email, *body_list = rev_and_commit.split("\n")
+        rev, parents, title, author, author_email, *body_list = rev_and_commit.split(
+            "\n"
+        )
         if rev_and_commit:
             git_commit = GitCommit(
                 rev=rev.strip(),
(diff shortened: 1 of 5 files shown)


commitizen/git.py
```diff
@@ -44,13 +44,20 @@ def __eq__(self, other) -> bool:
...
 class GitCommit(GitObject):
     def __init__(
-        self, rev, title, body: str = "", author: str = "", author_email: str = ""
+        self,
+        rev,
+        title,
+        body: str = "",
+        author: str = "",
+        author_email: str = "",
+        parents: list[str] = [],
```
noirbizarre, line 53: You can't use a list as default value, it will persist between calls, use an optional list instead.
```suggestion
        parents: list[str] | None = None,
```

commitizen/git.py
```diff
@@ -44,13 +44,20 @@ def __eq__(self, other) -> bool:
...
     ):
         self.rev = rev.strip()
         self.title = title.strip()
         self.body = body.strip()
         self.author = author.strip()
         self.author_email = author_email.strip()
+        self.parents = parents
```
noirbizarre, line 60: Then handle the default here:
```suggestion
        self.parents = parents or []
```

noirbizarre (changes requested):
Thanks for the contribution.

Good to me after the default list handling is fixed 👍🏼

docs/customization.md
```diff
@@ -507,6 +507,7 @@ Each `Change` has the following fields:
 | scope | `str | None` | An optional scope |
 | message | `str` | The commit message body |
 | sha1 | `str` | The commit `sha1` |
+| parents | `list[str]` | The parent commit(s) `sha1`(s) |
```
woile, line 510: Please add more information to this section, specially what's it useful for:

> Example usage: generate changelogs based on merge commits only, or listing commits that belong to the same PR under its own section.

commitizen/git.py
unleashed: Applied the suggestion.

commitizen/git.py
unleashed: Done.

docs/customization.md
unleashed: Added similar phrasing just below this table after the note. Let me know if that's ok.

noirbizarre (approved):
LGTM 🚀

woile (approved):

PR #1378, opened 2025-03-30
fix(commands/init): add missing uv provider to "cz init"

Lee-W:
## Checklist


Lee-W:
quick fix during some testing. will try to see whether it makes sense to add a test for it.

woile (approved):

woile:
We  could detect if `uv` is being used by checking the presence of `uv.lock`, and set it as default.

Lee-W:
Yep, we should. This PR is somewhat a work in progress (but already fixes something). I'll try to wrap it up today!

woile (approved):

PR #1381, opened 2025-04-05
feat: filter search for commit change type

Yusin0903:
## Description
Added search filter functionality to commit type selection in Conventional Commits. This allows users to quickly find commit types by typing part of the type name or description, enhancing user experience especially when working with many commit types.

Changes Made
1. Added use_search_filter: True to enable search filtering in the commit type selection.
2. Set use_jk_keys: False to prevent conflicts with search functionality (as j/k keys can be part of the search term).
3. Because the interaction questionary can not test directly so I test to verify search filter functionality and parameter works correctly.
4. Added newline at the end of file to comply with PEP 8 standards.

## Checklist

## Expected behavior
When a user types characters during commit type selection, the list of options is filtered to show only those containing the typed characters. For example, typing "fix" will show the item that contains "fix" .

1. Run cz commit to start the commit process
2. When the commit type selection appears, type what you want to search. eg: "fix"
3. Verify that the list filters to show the all "fix" items 
5. Try other search terms to confirm filtering works as expected

## Additional context

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -170,13 +170,15 @@ commitizen:
 
 | Parameter   | Type   | Default | Description                                                                                                                                                                                     |
 | ----------- | ------ | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `type`      | `str`  | `None`  | The type of questions. Valid type: `list`, `input` and etc. [See More][different-question-types]                                                                                                |
+| `type`      | `str`  | `None`  | The type of questions. Valid types: `list`, `select`, `input` and etc. The `select` type provides an interactive searchable list interface. [See More][different-question-types]                  |
 | `name`      | `str`  | `None`  | The key for the value answered by user. It's used in `message_template`                                                                                                                         |
 | `message`   | `str`  | `None`  | Detail description for the question.                                                                                                                                                            |
-| `choices`   | `list` | `None`  | (OPTIONAL) The choices when `type = list`. Either use a list of values or a list of dictionaries with `name` and `value` keys. Keyboard shortcuts can be defined via `key`. See examples above. |
+| `choices`   | `list` | `None`  | (OPTIONAL) The choices when `type = list` or `type = select`. Either use a list of values or a list of dictionaries with `name` and `value` keys. Keyboard shortcuts can be defined via `key`. See examples above. |
 | `default`   | `Any`  | `None`  | (OPTIONAL) The default value for this question.                                                                                                                                                 |
 | `filter`    | `str`  | `None`  | (OPTIONAL) Validator for user's answer. **(Work in Progress)**                                                                                                                                  |
-| `multiline` | `bool` | `False` | (OPTIONAL) Enable multiline support when `type = input`.                                                                                                                                            |
+| `multiline` | `bool` | `False` | (OPTIONAL) Enable multiline support when `type = input`.                                                                                                                                        |
+| `use_search_filter` | `bool` | `False` | (OPTIONAL) Enable search/filter functionality for list/select type questions. This allows users to type and filter through the choices.                                                  |
+| `use_jk_keys` | `bool` | `True` | (OPTIONAL) Enable/disable j/k keys for navigation in list/select type questions. Set to false if you prefer arrow keys only.                                                                    |
 
 [different-question-types]: https://github.com/tmbo/questionary#different-question-types
 
@@ -445,8 +447,8 @@ Commitizen gives you the possibility to provide your own changelog template, by:
 
 - providing one with your customization class
 - providing one from the current working directory and setting it:
-    - as [configuration][template-config]
-    - as `--template` parameter to both `bump` and `changelog` commands
+  - as [configuration][template-config]
+  - as `--template` parameter to both `bump` and `changelog` commands
 - either by providing a template with the same name as the default template
 
 By default, the template used is the `CHANGELOG.md.j2` file from the commitizen repository.
--- a/tests/test_cz_search_filter.py
+++ b/tests/test_cz_search_filter.py
@@ -1,0 +1,76 @@
+import pytest
+
+from commitizen.config import TomlConfig
+from commitizen.cz.customize import CustomizeCommitsCz
+
+TOML_WITH_SEARCH_FILTER = r"""
+[tool.commitizen]
+name = "cz_customize"
+
+[tool.commitizen.customize]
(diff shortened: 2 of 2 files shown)


commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -209,4 +211,4 @@ def process_commit(self, commit: str) -> str:
         m = re.match(pat, commit)
         if m is None:
             return ""
-        return m.group(3).strip()
+        return m.group(3).strip()
```
Lee-W, line 214: Not sure what's system is used 🤔 but this looks like removing EOF instead

Lee-W: Just fixed it for you

commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -43,9 +43,11 @@ class ConventionalCommitsCz(BaseCommitizen):
     def questions(self) -> Questions:
         questions: Questions = [
             {
-                "type": "list",
+                "type": "select",
                 "name": "prefix",
                 "message": "Select the type of change you are committing",
+                "use_search_filter": True,
+                "use_jk_keys": False,
```
Lee-W, line 50: I don't know whether we should set it as the default, probably not for v4. I would suggest making it configurable. Also, this now only works for `conventional_commits`, we would like it to work for all cz rules. Following up with this discussion, we probably could make the whole questionary behavior configurable, but that could be another PR

Yusin0903: OK, I will modify the search function to work for all cz rules. As for making it configurable, we can open another issue for that. Does that sound good?

Lee-W: I would probably say we should make it configurable first to retain current behavior and let ones who want to play with it opt it

Yusin0903: Hi @Lee-W,

I have tested the search filter functionality using the configuration approach:

```toml
[[tool.commitizen.customize.questions]]
type = "list"
name = "change_type"
message = "Select the type of change you are committing"
use_search_filter = true
use_jk_keys = false
```
It works as expected, so I've:

- Removed the modifications to conventional_commits.py
- Added documentation for the feature in customization.md
- Added test cases to verify the functionality
The search filter can now be enabled through configuration without modifying the core code.

Yusin0903:
Hi @Lee-W,

I have tested the search filter functionality using the configuration approach:

[[tool.commitizen.customize.questions]]
type = "list"
name = "change_type"
message = "Select the type of change you are committing"
use_search_filter = true
use_jk_keys = false
It works as expected, so I've:

Removed the modifications to conventional_commits.py
Added documentation for the feature in customization.md
Added test cases to verify the functionality
The search filter can now be enabled through configuration without modifying the core code.

Lee-W:
Hey @Yusin0903 , just noticed the CI is failing. Could you please take a look? Thanks 🙂

commitizen/cz/conventional_commits/conventional_commits.py
Lee-W: but if we do it this way, doesn't that mean we won't be able to use it in other cz rules?

Yusin0903: My understanding is that conventional commits and JIRA were not originally intended to be customized by clients (please correct me if I'm wrong), so wouldn't it be better to just define a fixed version (like a list or select option)?

Lee-W: Yes, that was the design. And I wrongly designed it 🥲 I always want to change `cz_customize` to something like `cz_empty` or `cz_no_default` and try to make everything as customizable to all rules.

woile:
We could also deprecate `cz_customize`, and just use `[tool.commitizen.customize]` to override the existining cz based on the `name`. Example: if you have set `cz_conventional_commits`, then anything in `tool.commitizen.customize` would override `cz_conventional_commits`

commitizen/cz/conventional_commits/conventional_commits.py
Yusin0903: I wanted to ask if cz_empty and cz_no_default imply that JIRA and conventional commits can also be customized? If that's the case, should we address this in the current issue?

Lee-W: > JIRA and conventional commits can also be customized? 

These two are somewhat customizable, but just not at the same level as `cz_customize`.

e.g., they have the fixed set of questions that cannot yet be customized at this moment but some of the behavior can still be changed by our config file

We don't need to do `cz_empty` in this PR. but we might want to pass these 2 options into all cz rules

Lee-W:
> We could also deprecate `cz_customize`, and just use `[tool.commitizen.customize]` to override the existining cz based on the `name`. Example: if you have set `cz_conventional_commits`, then anything in `tool.commitizen.customize` would override `cz_conventional_commits`

Yep, this is a fantastic idea!

commitizen/cz/conventional_commits/conventional_commits.py
Yusin0903: Not sure if this is what you meant,
But I added a question_type field in Settings, where users can choose "select" or "list".
This solution works based on my testing and fulfills the spec.

.cz.toml 
```
[tool.commitizen]
tag_format = "$version"
version_scheme = "semver"
version = "0.0.1"
major_version_zero = true
question_type = "select"
```

```
def __init__(self, config: BaseConfig):
        super().__init__(config)
        self.question_type = self.config.settings.get("question_type", "list")

    def questions(self) -> Questions:
        # Default questions
        questions: Questions = [
            {
                "type": self.question_type,
                "name": "prefix",
                "message": "Select the type of change you are committing",
                ...
         if questions[0]["type"] == "select":
            questions[0].update({"use_search_filter": True, "use_jk_keys": False})
```

Lee-W: Wait...

I thought we only need `use_search_filter` and `use_jk_keys` to config this kind of search on questions that is select or list?

Yusin0903: My thought is that the user only needs to set the type.
Since select type requires additional options like:

```json 
"use_search_filter": true,
"use_jk_keys": false
``` 
but list does not,
I apply these settings in code when the type is select,
so the user doesn't need to configure them manually.

Do you think config these two is a better way ?

Lee-W (approved):
I'm good with this change. We probably want to plan the cz_customize deprecation more thoroughly

woile (approved):

Issue #1352, opened 2025-02-07
cz commit -e crashes on editor exit


naimmas:
### Description

I want to add multi-line description so first I initialize my commit message header and then added the description using vim editor, but on write and quite :wq I received the following error

```
cz commit -e
? Select the type of change you are committing feat: A new feature. Correlates with MINOR in SemVer
? What is the scope of this change? (class or file name): (press [enter] to skip)
 HW timer
? Write a short and imperative summary of the code changes: (lower case and no period)
 feat and fixes
? Provide additional contextual information about the code changes: (press [enter] to skip)
 
? Is this a BREAKING CHANGE? Correlates with MAJOR in SemVer No
? Footer. Information about Breaking Changes and reference issues that this commit closes: (press [enter] to skip)
 
Traceback (most recent call last):
  File "/usr/local/bin/cz", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/local/Cellar/commitizen/4.1.1/libexec/lib/python3.13/site-packages/commitizen/cli.py", line 651, in main
    args.func(conf, arguments)()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/Cellar/commitizen/4.1.1/libexec/lib/python3.13/site-packages/commitizen/commands/commit.py", line 130, in __call__
    m = self.manual_edit(m)
  File "/usr/local/Cellar/commitizen/4.1.1/libexec/lib/python3.13/site-packages/commitizen/commands/commit.py", line 92, in manual_edit
    file.unlink()
    ^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/tempfile.py", line 494, in __getattr__
    a = getattr(file, name)
AttributeError: '_io.TextIOWrapper' object has no attribute 'unlink'
```

### Steps to reproduce

`cz commit -e`
edit the message
exit insert mode 
write & quite `:wq`

### Current behavior

The following error thrown on editor exit and commit message not saved in `.git/COMMIT_EDITMSG`
```
Traceback (most recent call last):
  File "/usr/local/bin/cz", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/local/Cellar/commitizen/4.1.1/libexec/lib/python3.13/site-packages/commitizen/cli.py", line 651, in main
    args.func(conf, arguments)()
[...]

PR #1396, opened 2025-04-29
fix(commit): use os.unlink to remove temp file

alxgu:
NamedTemporaryFile doesn't have a unlink function when delete=False is used

Fix https://github.com/commitizen-tools/commitizen/issues/1352

## Description
Fix the traceback, when `cz c --edit` is used

## Checklist

## Expected behavior

Save the commit message without an exception traceback.

1. Run cz commit -e
```
❯ cz c --edit
? Select the type of change you are committing fix: A bug fix. Correlates with PATCH in SemVer
? What is the scope of this change? (class or file name): (press [enter] to skip)
 test
? Write a short and imperative summary of the code changes: (lower case and no period)
 test
? Provide additional contextual information about the code changes: (press [enter] to skip)
 test
? Is this a BREAKING CHANGE? Correlates with MAJOR in SemVer No
? Footer. Information about Breaking Changes and reference issues that this commit closes: (press [enter] to skip)
 test
  File ".../commitizen/.venv/bin/cz", line 6, in <module>
    sys.exit(main())
  File "...//commitizen/commitizen/cli.py", line 656, in main
    args.func(conf, arguments)()
  File ".../commitizen/commitizen/commands/commit.py", line 130, in __call__
    m = self.manual_edit(m)
  File ".../commitizen/commitizen/commands/commit.py", line 92, in manual_edit
    file.unlink()
  File "/usr/lib/python3.10/tempfile.py", line 633, in __getattr__
    a = getattr(file, name)
AttributeError: '_io.TextIOWrapper' object has no attribute 'unlink'
```

## Additional context
https://github.com/commitizen-tools/commitizen/issues/1352

### DIFF
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -89,7 +89,7 @@ class Commit:
         subprocess.call(argv)
         with open(file_path) as temp_file:
             message = temp_file.read().strip()
-        file.unlink()
+        os.unlink(file.name)
         return message
 
     def __call__(self):
--- a/tests/commands/test_commit_command.py
+++ b/tests/commands/test_commit_command.py
@@ -511,8 +511,6 @@ def test_manual_edit(editor, config, mocker: MockFixture, tmp_path):
 
         assert edited_message == test_message.strip()
 
-        temp_file.unlink()
-
 
 @skip_below_py_3_13
 def test_commit_command_shows_description_when_use_help_option(


commitizen/commands/commit.py
```diff
@@ -89,7 +89,7 @@ def manual_edit(self, message: str) -> str:
         subprocess.call(argv)
         with open(file_path) as temp_file:
             message = temp_file.read().strip()
-        file.unlink()
+        os.unlink(file.name)
```
noirbizarre, line 92: Seems good to me, but maybe it would be cleaner to just move the `subprocess.call` into the context manager and remove the `delete=False` parameter.
It would avoid the temporary file remaining present in case of an exception during the `subprocess.call()` plus, from the [latest documentation](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile), it seems to properly handle some permissions issues that might happen while doing a manual unlink on Windows.

Lee-W: sounds good. let me do the change and get this merged

Lee-W: @noirbizarre after some testing, it doesn't seem to work that way 🤔 
@alxgu 's implementation works fine though

Lee-W (approved):

commitizen/commands/commit.py
Lee-W: I'll approve it and let you decide whether we should merge it

noirbizarre (approved):

commitizen/commands/commit.py
noirbizarre: Good to me 👍🏼

Lee-W: niiice. let's merge it then

alxgu: Thank you for reviewing and merging 😃

PR #1398, opened 2025-05-05
fix(docs): fix the regex and examples in the monorepo documentation

baluyotraf:
## Description
The regex in the monorepo example seems to follow the Python formatting, but the toml text should follow the raw string instead (unless this library internally calls eval?).

Also the example does not follow regex format in the setting.

## Checklist

## Expected behavior

Update the regex pattern and update the example regex.

### DIFF
--- a/docs/tutorials/monorepo_guidance.md
+++ b/docs/tutorials/monorepo_guidance.md
@@ -71,11 +71,11 @@ Example config and commit for `library-b`:
 
 ```toml
 [tool.commitizen.customize]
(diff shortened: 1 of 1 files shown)


noirbizarre (approved):

Lee-W (approved):

Lee-W:
Looks like our `poetry ci` command is not running commitizen correctly. 🤔  I just fixed the commit message

Lee-W:
I think `commitizen-branch` was wrongly implemented. I create a draft PR to keep track of it and hope I can wrap it up some time...

https://github.com/commitizen-tools/commitizen/pull/1400

PR #1400, opened 2025-05-06
Add check against main branch hook

Lee-W:
## Checklist

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import re
 import sys
 from typing import Any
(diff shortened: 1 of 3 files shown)


commitizen/git.py
```diff
@@ -310,3 +310,10 @@ def _get_log_as_str_list(start: str | None, end: str, args: str) -> list[str]:
     if not c.out:
         return []
     return c.out.split(f"{delimiter}\n")
+
+
+def guess_default_branch() -> str:
+    c = cmd.run("git branch --format '%(refname:short)' --list master main")
```
noirbizarre, line 316: This makes a strong hypothesis that the reference branch can only be `master` or `main` so it won't work with `develop` which is common, or any other "exotic choice".

To handle this, I use `git branch -r | grep -Po "HEAD -> .+/\K.+$" | head -n 1` which works with any default branch as long as there is a remote with a default. The `head` is for the case where there are multiple remotes, but makes the hypothesis that they share the same default branch.

Lee-W: I guess I'll have to try those commands 👀 Looks interesting!

I'm thinking of letting users set `default_git_remote` and `default_git_branch_name` (but not yet have time to impletemtn it 🥲)

commitizen/commands/check.py
```diff
@@ -92,7 +90,8 @@ def __call__(self):
             )
         out.success("Commit validation: successful!")
 
-    def _get_commits(self):
+    def _get_commits(self) -> list[git.GitCommit]:
+        # TODO: this method seems to do a few different things. probably would be better if we could split it to smaller functions
```
bearomorphism, line 94: Already done in refactors branch

Lee-W:
close in favor of https://github.com/commitizen-tools/commitizen/pull/1519

PR #1410, opened 2025-05-12
fix: ignore invalid tags

carlossg:
Starting in 4.4.0 due to #1297 when an invalid tag is present an exception is thrown

This commit ignores the invalid tags

Follow up from #1375

## Description

Starting in 4.4.0 due to #1297 when an invalid tag is present an exception is thrown

This commit ignores the invalid tags

Follow up from #1375

## Checklist

## Expected behavior

in https://github.com/carlossg/commitizen-test.git the fix changes from

```
❯ cz bump --dry-run
  File "/opt/homebrew/bin/cz", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Cellar/commitizen/4.7.0/libexec/lib/python3.13/site-packages/commitizen/cli.py", line 656, in main
    args.func(conf, arguments)()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/commitizen/4.7.0/libexec/lib/python3.13/site-packages/commitizen/commands/bump.py", line 147, in __call__
    current_version = self.scheme(provider.get_version())
                                  ~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/commitizen/4.7.0/libexec/lib/python3.13/site-packages/commitizen/providers/scm_provider.py", line 21, in get_version
    versions = sorted(rules.extract_version(t) for t in version_tags)
  File "/opt/homebrew/Cellar/commitizen/4.7.0/libexec/lib/python3.13/site-packages/commitizen/providers/scm_provider.py", line 21, in <genexpr>
    versions = sorted(rules.extract_version(t) for t in version_tags)
                      ~~~~~~~~~~~~~~~~~~~~~^^^
  File "/opt/homebrew/Cellar/commitizen/4.7.0/libexec/lib/python3.13/site-packages/commitizen/tags.py", line 149, in extract_version
    raise InvalidVersion(
        f"Invalid version tag: '{tag.name}' does not match any configured tag format"
    )
packaging.version.InvalidVersion: Invalid version tag: '1.0.0.fb3a1d1' does not match any configured tag format
```

to

```
❯ cz bump --dry-run
bump: version 1.0.0 → 1.1.0
tag to create: 1.1.0
increment detected: MINOR
```

### DIFF
--- a/commitizen/providers/scm_provider.py
+++ b/commitizen/providers/scm_provider.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from packaging.version import InvalidVersion
+
 from commitizen.git import get_tags
 from commitizen.providers.base_provider import VersionProvider
 from commitizen.tags import TagRules
@@ -18,7 +20,13 @@ class ScmProvider(VersionProvider):
         rules = TagRules.from_settings(self.config.settings)
         tags = get_tags(reachable_only=True)
         version_tags = rules.get_version_tags(tags)
-        versions = sorted(rules.extract_version(t) for t in version_tags)
+        versions = []
+        for t in version_tags:
+            try:
+                versions.append(rules.extract_version(t))
+            except InvalidVersion:
+                continue
+        versions = sorted(versions)
         if not versions:
             return "0.0.0"
         return str(versions[-1])
--- a/tests/providers/test_scm_provider.py
+++ b/tests/providers/test_scm_provider.py
@@ -23,6 +23,7 @@ from tests.utils import (
         ("$version", "0.1.0", "0.1.0"),
         ("$version", "v0.1.0", "0.1.0"),
         ("$version", "v-0.1.0", "0.0.0"),
+        ("$version", "1.0.0.xxxx", "0.0.0"),
         # If tag_format is not None or $version, TAG_FORMAT_REGEXS are used, which are
         # much more lenient but require a v prefix.
         ("v$version", "v0.1.0", "0.1.0"),


noirbizarre (changes requested):
I would says silencing exception with `try..except..pass` is never a good solution.
The exception was already raised before the PR (it was actually the reason for the PR as it was producing 1 error line by tag). The fact that it's breaking now is an unexpected side-effect which should be fixed, but silencing and ignoring them totally can't be the solution. It allows to see when you make a mistake on the tag or you forget some or when you tag are legitimately invalid (which seem to be your case)

I think there is 3 acceptable solutions:
- fix the regression and make the exception non-breaking with 1 error line by exception
- make an explicit opt-in option to ignore them
- handle the fact that your tags are not valid by:
	- using the dedicated [`ignored_tag_formats` option](https://commitizen-tools.github.io/commitizen/config/#ignored_tag_formats) to ignore them if you don't want them in your changelog (which given you were intending to silence the exception seem to fit your use case)
	- using a `version_scheme` in which they are valid if you want them in your changelog

I think that fixing the regression is the path to follow (I need to re-read the PR comments to be sure we did not choose to make it breaking given the possibility to explicitly ignore them) but your case is already handled by using `ignored_tag_formats = ["${major}-${minor}-${patch}-*"]`

noirbizarre:
Issue found and should be fixed by #1418 👍🏼

PR #1412, opened 2025-05-14
refactor(customize): improve code readability

bearomorphism:
## Description

Make the code shorter but still easy to read.

## Checklist

## Expected behavior

NA, this is refactor.

## Additional context

NA

### DIFF
--- a/commitizen/cz/customize/customize.py
+++ b/commitizen/cz/customize/customize.py
@@ -33,35 +33,17 @@ class CustomizeCommitsCz(BaseCommitizen):
             raise MissingCzCustomizeConfigError()
         self.custom_settings = self.config.settings["customize"]
 
-        custom_bump_pattern = self.custom_settings.get("bump_pattern")
-        if custom_bump_pattern:
-            self.bump_pattern = custom_bump_pattern
-
-        custom_bump_map = self.custom_settings.get("bump_map")
-        if custom_bump_map:
-            self.bump_map = custom_bump_map
-
-        custom_bump_map_major_version_zero = self.custom_settings.get(
-            "bump_map_major_version_zero"
-        )
-        if custom_bump_map_major_version_zero:
-            self.bump_map_major_version_zero = custom_bump_map_major_version_zero
-
-        custom_change_type_order = self.custom_settings.get("change_type_order")
-        if custom_change_type_order:
-            self.change_type_order = custom_change_type_order
-
-        commit_parser = self.custom_settings.get("commit_parser")
-        if commit_parser:
-            self.commit_parser = commit_parser
-
-        changelog_pattern = self.custom_settings.get("changelog_pattern")
-        if changelog_pattern:
-            self.changelog_pattern = changelog_pattern
-
-        change_type_map = self.custom_settings.get("change_type_map")
-        if change_type_map:
-            self.change_type_map = change_type_map
+        for attr_name in [
+            "bump_pattern",
+            "bump_map",
+            "bump_map_major_version_zero",
+            "change_type_order",
+            "commit_parser",
+            "changelog_pattern",
+            "change_type_map",
+        ]:
+            if value := self.custom_settings.get(attr_name):
+                setattr(self, attr_name, value)
 
     def questions(self) -> Questions:
         return self.custom_settings.get("questions", [{}])
@@ -70,8 +52,7 @@ class CustomizeCommitsCz(BaseCommitizen):
         message_template = Template(self.custom_settings.get("message_template", ""))
         if getattr(Template, "substitute", None):
             return message_template.substitute(**answers)  # type: ignore
-        else:
-            return message_template.render(**answers)
+        return message_template.render(**answers)
 
     def example(self) -> str:
         return self.custom_settings.get("example") or ""
@@ -83,12 +64,7 @@ class CustomizeCommitsCz(BaseCommitizen):
         return self.custom_settings.get("schema") or ""
 
     def info(self) -> str:
-        info_path = self.custom_settings.get("info_path")
-        info = self.custom_settings.get("info")
-        if info_path:
+        if info_path := self.custom_settings.get("info_path"):
             with open(info_path, encoding=self.config.settings["encoding"]) as f:
-                content = f.read()
-            return content
-        elif info:
-            return info
-        return ""
+                return f.read()
+        return self.custom_settings.get("info") or ""


bearomorphism:
I was curious about the difference between `.get("some_key", "")` and `.get("some_key") or ""`

https://chatgpt.com/share/68254494-f83c-8010-8ab3-7f8808682bcf

commitizen/cz/customize/customize.py
```diff
@@ -21,46 +21,32 @@
 
 
 class CustomizeCommitsCz(BaseCommitizen):
-    bump_pattern = defaults.bump_pattern
-    bump_map = defaults.bump_map
-    bump_map_major_version_zero = defaults.bump_map_major_version_zero
-    change_type_order = defaults.change_type_order
-
```
noirbizarre, line 28: I am not sure about this change as those are class variables, this change set them on instantiation.
While I theoretically agree with the pattern change, it is a breaking change as commitizen is meant to be extended and customized and if someone is referencing those class variables from the class and not an instance in its code for any reason, they will have their value changed.

commitizen/cz/customize/customize.py
```diff
@@ -21,46 +21,32 @@
...
             self.commit_parser = commit_parser
 
-        changelog_pattern = self.custom_settings.get("changelog_pattern")
-        if changelog_pattern:
+        if changelog_pattern := self.custom_settings.get("changelog_pattern"):
             self.changelog_pattern = changelog_pattern
 
-        change_type_map = self.custom_settings.get("change_type_map")
-        if change_type_map:
+        if change_type_map := self.custom_settings.get("change_type_map"):
```
noirbizarre, line 49: Given this is n time the same pattern, maybe an helper function would make this even easier to read, something like `self.var = self.settings_or_default("var")`.

commitizen/cz/customize/customize.py
```diff
@@ -83,12 +68,7 @@ def schema(self) -> str:
...
-        if info_path:
+        if info_path := self.custom_settings.get("info_path"):
             with open(info_path, encoding=self.config.settings["encoding"]) as f:
-                content = f.read()
-            return content
-        elif info:
-            return info
-        return ""
+                return f.read()
+        return self.custom_settings.get("info") or ""
```
noirbizarre, line 74: Yes, this one was clearly overly complicated 😅

commitizen/cz/customize/customize.py
bearomorphism: Ok let me revert this part. I haven't used Python for a while and didn't know they belong to the class variables.

noirbizarre:
> I was curious about the difference between `.get("some_key", "")` and `.get("some_key") or ""`

Actually, it matters: if you have `some_key:` in your settings, meaning empty key in yaml, the key exists and the first one returns `None` while the second one returns `""`.
The second approach is most of the time safer is it ensure all falsy values will and with the same type:

| yaml | `.get("key", "")` | `.get("key") or ""` |
|-----------|------------------------|--------------------------|
| Not set   | `""` | `""` |
| `key: ""` | `""` | `""` |
| `key:` | `None` | `""` |
| `key: false` | `False` | `""` |
| `key: 0` | `0` | `""` |
| `key: []` | `[]` | `""` |
| `key: {}` | `{}` | `""` |
| `key: no` (yaml 1.1) | `False` | `""` |

So it really depends on your case, but if there is no default value, and you want to be sure of the type you have in case of falsy value, go for the 2nd.

commitizen/cz/customize/customize.py
bearomorphism: Agree. I'm using a simple for loop to do the same logic in the latest iteration.

bearomorphism:
> > I was curious about the difference between `.get("some_key", "")` and `.get("some_key") or ""`
> 
> Actually, it matters: if you have `some_key:` in your settings, meaning empty key in yaml, the key exists and the first one returns `None` while the second one returns `""`. The second approach is most of the time safer is it ensure all falsy values will and with the same type:
> 
> yaml	`.get("key", "")`	`.get("key") or ""`
> Not set	`""`	`""`
> `key: ""`	`""`	`""`
> `key:`	`None`	`""`
> `key: false`	`False`	`""`
> `key: 0`	`0`	`""`
> `key: []`	`[]`	`""`
> `key: {}`	`{}`	`""`
> `key: no` (yaml 1.1)	`False`	`""`
> So it really depends on your case, but if there is no default value, and you want to be sure of the type you have in case of falsy value, go for the 2nd.

Thank you for explanation. Actually, the thing I'm not sure is if want to fallback to empty string if the value is falsy (current implementation).

noirbizarre:
I think this has been a recurrent discussion on reviews: we have some improvements to do on the "cli flags > user settings > default value" cascading handling:
- lots of repetitions
- sometimes we have the same setting with different behavior depending on the command
- ...
There is room for improvement.
But we have to maintain backward compatibility, so the day we change that, it might be a breaking change.

Lee-W:
> I think this has been a recurrent discussion on reviews: we have some improvements to do on the "cli flags > user settings > default value" cascading handling:
> 
>     * lots of repetitions
> 
>     * sometimes we have the same setting with different behavior depending on the command
> 
>     * ...
> 
> 
> There is room for improvement. But we have to maintain backward compatibility, so the day we change that, it might be a breaking change.

I've also wanted to rewrite the whole setting, default value thing....

bearomorphism:
> > I think this has been a recurrent discussion on reviews: we have some improvements to do on the "cli flags > user settings > default value" cascading handling:
> > ```
> > * lots of repetitions
> > 
> > * sometimes we have the same setting with different behavior depending on the command
> > 
> > * ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There is room for improvement. But we have to maintain backward compatibility, so the day we change that, it might be a breaking change.
> 
> I've also wanted to rewrite the whole setting, default value thing....

I can try this. We can discuss when you have time.

commitizen/cz/customize/customize.py
bearomorphism: I have reverted the change here. Please review.

Lee-W (approved):

Lee-W:
Created an issue https://github.com/commitizen-tools/commitizen/issues/1445 to track and discuss this redesign, probably will go to v5

PR #1404, opened 2025-05-11
docs: improve clarity of docs and fix typo/grammar issues

bearomorphism:
Closes #1405
Closes #1413

## Description

When I first started looking into how to use this tool, I found the documentation really hard to read.
It's such a pity for such a useful tool!

Thanks to powerful AI tools, improving clarity of the documents can be easily done.

- Rewrite README.md
- Rewrite contributing.md
- Fix several typos and grammar mistakes

## Checklist

- [x] Add test cases to all the changes you introduce
- [x] Run `poetry all` locally to ensure this change passes linter check and test
- [x] Test the changes on the local machine manually
- [x] Update the documentation for the changes

## Expected behavior

It should be easier for the readers to understand what Commitizen does after they enter the landing page of Commitizen.

The typo of the cli output should be fixed.

## Steps to Test This Pull Request

NA

## Additional context

NA

PR #1414, opened 2025-05-15
feat(cli): add --tag-format argument to changelog command

jappja:
## Description

This PR adds missing "--tag-format" argument to `cz changelog` command, similar to `cz bump` command.
The changelog command implementation already checks "tag_format" argument but cli support was just missing.
## Checklist

## Expected behavior

Tag format can be given as argument to `cz changelog` command. The changelog command already checks "tag_format" argument: https://github.com/commitizen-tools/commitizen/blob/be02801f92cb948ba86d8c1a9a306be4172e4efe/commitizen/commands/changelog.py#L85

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -450,6 +450,10 @@ data = {
                         "help": "Export the changelog template into this file instead of rendering it",
                     },
                     *deepcopy(tpl_arguments),
+                    {
+                        "name": "--tag-format",
+                        "help": "The format of the tag, wrap around simple quotes",
+                    },
                 ],
             },
             {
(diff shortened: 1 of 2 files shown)


noirbizarre (approved):
I checked and the `changelog` command is actually already handling it but it was missing 👍🏼
Thanks !

woile (approved):

bearomorphism:
> let's wait for other doc changes and then merge this one

If you're referring to #1404, there is no dependencies. It's ok if you merge this one first.

Lee-W:
it's not strickly blocking, wanted to get it released with docs but yep won't really hurt if we merge it first. let's do it 💪

PR #1424, opened 2025-05-17
refactor(init): code cleanup and better test coverage

bearomorphism:
## Checklist

### Code Changes

### Documentation Changes

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -14,7 +14,12 @@ from commitizen.cz import registry
 from commitizen.defaults import CONFIG_FILES, DEFAULT_SETTINGS
 from commitizen.exceptions import InitFailedError, NoAnswersError
 from commitizen.git import get_latest_tag_name, get_tag_names, smart_open
-from commitizen.version_schemes import KNOWN_SCHEMES, Version, get_version_scheme
+from commitizen.version_schemes import (
+    KNOWN_SCHEMES,
+    Version,
+    VersionScheme,
+    get_version_scheme,
+)
 
 
 class ProjectInfo:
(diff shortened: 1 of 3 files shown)


tests/commands/test_init_command.py
```diff
@@ -206,22 +204,22 @@ def test_pre_commit_config_without_cz_hook(_, default_choice, tmpdir, config):
...
+            p.write(yaml.safe_dump({"repos": [CZ_HOOK_CONFIG]}))
 
             commands.Init(config)()
             check_cz_config(default_choice)
             # check that config is not duplicated
-            check_pre_commit_config([cz_hook_config])
+            check_pre_commit_config([CZ_HOOK_CONFIG])
 
 
 class TestNoPreCommitInstalled:
```
bearomorphism, line 225: `test_pre_commit_not_installed` is not working as expected...

tests/commands/test_init_command.py
```diff
@@ -36,14 +37,14 @@ def unsafe_ask(self):
...
-expected_config = (
+EXPECTED_CONFIG = (
     "[tool.commitizen]\n"
     'name = "cz_conventional_commits"\n'
     'tag_format = "$version"\n'
     'version_scheme = "semver"\n'
-    'version = "0.0.1"\n'
     "update_changelog_on_bump = true\n"
     "major_version_zero = true\n"
+    'version = "0.0.1"\n'
```
bearomorphism, line 47: Does the order of these configs matter here?

commitizen/commands/init.py
```diff
@@ -113,194 +110,176 @@ def __call__(self):
...
 
-            if not tag_format:
-                tag_format = default_format
-        return tag_format
+        return tag_format or default_format
 
     def _ask_version_provider(self) -> str:
         """Ask for setting: version_provider"""
 
         OPTS = {
```
woile, line 208: Ideally, the `VersionProvider` ABC class should include `identifier` and `description`. Which would allow us to do something like

```sh
providers = metadata.entry_points(group=PROVIDER_ENTRYPOINT)

{provider.identifier: provider.description for provider in providers} 
```

This way, custom user made providers would also appear in the init functions.

---

Just to be clear, this is optional, I hardcoded them here because it would help 90% of the users and I didn't want to go further.

tests/commands/test_init_command.py
woile: Nop

bearomorphism: Thanks.

PR #1425, opened 2025-05-17
refactor(changelog): cleanup and add test cases

bearomorphism:
## Checklist

### Code Changes

### Documentation Changes

### DIFF
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -74,6 +74,23 @@ def get_commit_tag(commit: GitCommit, tags: list[GitTag]) -> GitTag | None:
     return next((tag for tag in tags if tag.rev == commit.rev), None)
 
 
+def _get_release_info(
+    current_tag_name: str,
+    current_tag_date: str,
+    changes: dict[str | None, list],
+    changelog_release_hook: ChangelogReleaseHook | None,
+    commit_tag: GitTag | None,
+) -> dict[str, Any]:
+    release = {
+        "version": current_tag_name,
+        "date": current_tag_date,
+        "changes": changes,
+    }
+    if changelog_release_hook:
+        return changelog_release_hook(release, commit_tag)
+    return release
+
+
 def generate_tree_from_commits(
     commits: list[GitCommit],
     tags: list[GitTag],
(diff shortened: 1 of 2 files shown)


tests/test_changelog.py
```diff
@@ -1531,10 +1531,46 @@ def test_get_smart_tag_range_returns_an_extra_for_a_range(tags):
...
+def test_get_smart_tag_range_returns_an_empty_list_for_nonexistent_end_tag(tags):
+    start = tags[0]
+    res = changelog.get_smart_tag_range(tags, start.name, "nonexistent")
+    assert len(tags) == len(res)
+
+
+def test_get_smart_tag_range_returns_an_empty_list_for_nonexistent_start_tag(tags):
+    end = tags[0]
+    res = changelog.get_smart_tag_range(tags, "nonexistent", end.name)
+    assert res[0].name == tags[1].name
```
bearomorphism, line 1566: Not sure if this is the expected behavior

commitizen/changelog.py
```diff
@@ -114,7 +111,7 @@ def generate_tree_from_commits(
             and commit_tag not in used_tags
             and rules.include_in_changelog(commit_tag)
         ):
-            used_tags.append(commit_tag)
+            used_tags.add(commit_tag)
```
Lee-W, line 114: Should we add only tags that are not None?

Lee-W: probably need to handle `current_tag` in this case as well 🤔

commitizen/changelog.py
```diff
@@ -170,21 +167,23 @@ def process_commit_message(
     changes: dict[str | None, list],
     change_type_map: dict[str, str] | None = None,
 ):
-    message: dict = {
+    message = {
```
Lee-W, line 170: May I know why we're removing it?

commitizen/git.py
```diff
@@ -41,6 +41,9 @@ def __eq__(self, other) -> bool:
             return False
         return self.rev == other.rev  # type: ignore
 
+    def __hash__(self):
```
Lee-W, line 44: I guess this is for `used_tags`?

commitizen/changelog.py
```diff
@@ -337,17 +349,17 @@ def get_oldest_and_newest_rev(
     if not tags_range:
         raise NoCommitsFoundError("Could not find a valid revision range.")
 
-    oldest_rev: str | None = tags_range[-1].name
+    oldest_rev = tags_range[-1].name
```
Lee-W, line 352: May I know why we want to remove the type here?

commitizen/changelog.py
```diff
@@ -268,40 +268,52 @@ def incremental_build(
...
+    oldest = oldest or newest
+
+    names = set(tag.name for tag in tags)
+    has_newest = newest in names
+    has_oldest = oldest in names
+    if not has_newest and not has_oldest:
+        return []
+
+    if not has_newest:
+        return tags[1:]
```
Lee-W, line 308: I'm a bit confused. Why is the `1` here? I must admit that I'm not that familiar with this part 🤦‍♂️

bearomorphism: I remember I was trying to align the behavior before and after the refactoring

commitizen/changelog.py
bearomorphism: I'm not exactly sure when it's necessary to explicitly annotate variable types in Python, since the type is often inferred from the assigned value.

commitizen/changelog.py
bearomorphism: We can hold off this part and separate it in another PR though. The algorithm here is a bit confusing.

commitizen/changelog.py
bearomorphism: I just added `dict[str, str | list[str] | Any]` here

PR #1444, opened 2025-05-20
refactor(BaseConfig): update function name, upgrade mypy version

bearomorphism:
## Checklist

### Code Changes

### Documentation Changes


commitizen/config/base_config.py
```diff
@@ -30,8 +26,5 @@ def set_key(self, key, value):
     def update(self, data: Settings) -> None:
         self._settings.update(data)
 
-    def add_path(self, path: str | Path) -> None:
-        self._path = Path(path)
-
```
bearomorphism, line 35: I just found this method confusing. It's actually updating the path, right?

Lee-W: I think the main reason we make it this way is that we want to handle both `str` and `Path` implicitly. We probably could use a setter if that's better

bearomorphism: OK I was trying the following now

```python
    @property
    def path(self) -> Path | None:
        return self._path

    @path.setter
    def path(self, path: str | Path) -> None:
        self._path = Path(path)
```

but mypy isn't happy

```
commitizen/config/yaml_config.py:17: error: Incompatible types in assignment (expression has type "Path | str", variable has type "Path | None")  [assignment]
commitizen/config/toml_config.py:17: error: Incompatible types in assignment (expression has type "Path | str", variable has type "Path | None")  [assignment]
commitizen/config/json_config.py:16: error: Incompatible types in assignment (expression has type "Path | str", variable has type "Path | None")  [assignment]
```

I'm taking a look at related issues and PRs. Updating mypy to `1.15.0` does not work for me.

- https://github.com/python/mypy/issues/3004
- https://github.com/python/mypy/pull/18510

commitizen/config/base_config.py
```diff
@@ -19,6 +19,19 @@ def settings(self) -> Settings:
...
+    mypy does not like the following until 1.16
+    See https://github.com/python/mypy/pull/18510
+    TODO: update this when 1.16 is available
+
+    @path.setter
+    def path(self, path: str | Path) -> None:
+        self._path = Path(path)
+    """
+
+    def update_path(self, path: str | Path) -> None:
```
Lee-W, line 32: Or maybe let's use `setter` here with type ignore and update it after 1.16. WDYT?

bearomorphism: sounds good

bearomorphism: fixed.

Lee-W (approved):

PR #1455, opened 2025-05-25
refactor(check): remove unnecessary list construction

bearomorphism:
## Checklist

### Code Changes

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import re
 import sys
+from itertools import islice
 from typing import Any
 
 from commitizen import factory, git, out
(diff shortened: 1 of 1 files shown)


commitizen/commands/check.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, config: BaseConfig, arguments: dict[str, Any], cwd=os.getcwd(
 
         self._valid_command_argument()
 
-        self.config: BaseConfig = config
+        self.config = config
```
Lee-W, line 47: Is there any reason we remove it?

commitizen/commands/check.py
```diff
@@ -126,14 +124,18 @@ def _filter_comments(msg: str) -> str:
...
         """
-
-        lines = []
-        for line in msg.split("\n"):
-            if "# ------------------------ >8 ------------------------" in line:
-                break
-            if not line.startswith("#"):
-                lines.append(line)
-        return "\n".join(lines)
+        msg_lines = msg.split("\n")
```
Lee-W, line 127: Not sure whether we want to do this. Look like we'll loop through lines twice and the logic looks more complicated

Issue #1249, opened 2024-10-06
Commitizen is not compatible with recent versions of ipython because of its dependency on questionary


lsorber:
### Description

Not so recent versions of ipython require `prompt-toolkit>=3.0.41`, but commitizen depends on `questionary<3.0`, the latest version of which depends on `prompt-toolkit<=3.0.36`[^1]. To comply with both requirements, you need to downgrade ipython to a version of end 2022.

[^1]: https://github.com/tmbo/questionary/blob/2.0.1/pyproject.toml#L36

Related: https://github.com/commitizen-tools/commitizen/issues/5

### Possible Solution

1. Ask questionary to release a new version (their master branch relaxed the constraint on prompt-toolkit).
2. Or, replace or remove the dependency on questionary.

bearomorphism:
Closing as **fixed by an upstream relax** in `questionary 2.1.0`.

Background:
- The original incompatibility was `questionary <=2.0.x` pinning `prompt_toolkit<=3.0.36`, which conflicted with ipython's `prompt_toolkit>=3.0.41`.
- `questionary 2.1.0` relaxed the constraint to `prompt_toolkit>=2.0,<4.0`, removing the conflict.
- commitizen master pins `questionary (>=2.0,<3.0)` (`pyproject.toml:16`), which now resolves cleanly with current ipython 8.x and 9.x.
- The repo''s own `uv.lock` resolves `questionary 2.1.1` + `prompt_toolkit 3.0.51` + `ipython 9.12.0` together, demonstrating the scenario the original issue said was impossible.

If anyone hits this again on a newer ipython release, please reopen with the resolved environment so we can pin further.

Surfaced via the round-2 triage in #1965.

Issue #1073, opened 2024-04-18
commitizen v4


noirbizarre:
### Description

Discuss and tracking breaking changes to include in the next major release.

### Possible Solution

- [ ] drop all deprecation:
  - [ ] `--version-type`
- [ ] Review configuration handling (See https://github.com/commitizen-tools/commitizen/pull/996#issuecomment-1987334462)

### Additional context

Should we release a major by breaking change or group them into a single major release ?

Lee-W:
Not sure whether we need to create a `4.0` branch 🤔

Lee-W:
I'm thinking of splitting version_schemes into smaller modules. Make both version_schemes and providers pluggable

woile:
> I'm thinking of splitting version_schemes into smaller modules
good idea, to be in-line with providers, and it would be easier to find them

> Make both version_schemes and providers pluggable
what do you mean? they are already plugins, right?

Lee-W:
> what do you mean? they are already plugins, right?

ah, you're right. I forget it 🤦‍♂️

woile:
We should improve the templates to make it easier to extend

noirbizarre:
Brainstorming a few ideas (doesn't mean that we need everything in the next major):

- **feature**: extend https://github.com/commitizen-tools/commitizen/pull/995 to support (already started):
  - `alt-tag-formats`: to handle tag format changes. Legacy tags can still be parsed and included in the changelog
  - `ignore-tag-formats`: tag that are expected and should be ignored (should not result in a warning)
  - attach the version parts to the version scheme (working on semver2 made me realize that the current PR will only work for semver/pep440)
- **feature**: Make something for all the empty commits or release after prerelease issues:
  - https://github.com/commitizen-tools/commitizen/pull/723
  - https://github.com/commitizen-tools/commitizen/issues/662
  - https://github.com/commitizen-tools/commitizen/issues/1068
  - https://github.com/commitizen-tools/commitizen/issues/753
  - https://github.com/commitizen-tools/commitizen/issues/247
  - https://github.com/commitizen-tools/commitizen/issues/753
- **breaking**: Centralize configurations and defaults. Single source of trust. Why not rely on `pydantic`:
  - https://github.com/commitizen-tools/commitizen/issues/883
  - https://github.com/commitizen-tools/commitizen/pull/996#issuecomment-1987334462
- **feature**: `pre` and `post` bump hooks (entrypoints):
[...]

PR #1467, opened 2025-05-29
fix(question): strict type with pydantic for questions

bearomorphism:
## Description

Enforce strict type with [Pydantic](https://docs.pydantic.dev/latest/) for questions.

## Checklist

### Code Changes

Run `cz c -- --allow-empty` and finish the conventional commits questions.


commitizen/question.py
```diff
@@ -0,0 +1,49 @@
...
+class ListQuestion(QuestionBase):
+    type: Literal["list"]
+    choices: list[Choice]
+    use_shortcuts: Optional[bool] = None
+
+
+class SelectQuestion(QuestionBase):
+    type: Literal["select"]
+    choices: list[Choice]
+    use_search_filter: Optional[bool] = None  # TODO: confirm type
```
bearomorphism, line 31: Need to confirm if this is correct. Ping @Yusin0903 @Lee-W 

#1381

bearomorphism:
Found related pydantic discussion here: #1073

bearomorphism:
Are there still incompatibility issues if we use `pydantic>=1.10,<3`?

commitizen/question.py
Yusin0903: Hi, I checked the questionary.select function signature and found that use_jk_keys, use_search_filter should be bool, not Optional[bool].

Reference: https://github.com/tmbo/questionary/blob/master/questionary/prompts/select.py

woile:
What's the benefit of adding pydantic as a dependency? Couldn't we use dataclasses/typeddict and protocols? The user interaction with commitizen barely includes user input, which is already validated by `questionary`. Therefore, type check at runtime is barely needed 🤔 

We already have issues like this: https://github.com/commitizen-tools/commitizen/issues/1474

woile:
I guess I misinterpreted the ticket. I thought it was creating an issue on a downstream project. 

My point is that a ticket is a burden on (volunteer) maintainers. Because we pin dependencies, if users add commitizen as a dev-dependency, we limit what they can install. pydantic is a widely used dependency, we would have to rush into supporting v3. This PR doesn't make use of much iterfaces from pydantic, but by having it available, we are telling other contributors: yes, you can use the full pydantic featureset. It's a long term issue FMPOV

This would be a clearer issue:
https://github.com/commitizen-tools/commitizen/issues/1249

bearomorphism:
> I guess I misinterpreted the ticket. I thought it was creating an issue on a downstream project.
> 
> My point is that a ticket is a burden on (volunteer) maintainers. Because we pin dependencies, if users add commitizen as a dev-dependency, we limit what they can install. pydantic is a widely used dependency, we would have to rush into supporting v3 (depending on how we pin it). This PR doesn't make use of much iterfaces from pydantic, but by having it available, we are telling other contributors: yes, you can use the full pydantic featureset. It's a long term issue FMPOV
> 
> This would be a clearer issue: #1249

You're right. I intention of this PR was that I just wanted a more strict type for `Questions`. The runtime check is not needed.

Let me close this PR and use `TypedDIct` to address the issue.

Lee-W:
> > I guess I misinterpreted the ticket. I thought it was creating an issue on a downstream project.
> > My point is that a ticket is a burden on (volunteer) maintainers. Because we pin dependencies, if users add commitizen as a dev-dependency, we limit what they can install. pydantic is a widely used dependency, we would have to rush into supporting v3 (depending on how we pin it). This PR doesn't make use of much iterfaces from pydantic, but by having it available, we are telling other contributors: yes, you can use the full pydantic featureset. It's a long term issue FMPOV
> > This would be a clearer issue: #1249
> 
> You're right. I intention of this PR was that I just wanted a more strict type for `Questions`. The runtime check is not needed.
> 
> Let me close this PR and use `TypedDIct` to address the issue.

Maybe we could mention this in our FAQ as well 🤔

PR #1470, opened 2025-05-30
refactor(check): rename variable and compile pattern once

bearomorphism:
## Checklist

### Code Changes

Run `cz check --message "<random strings>"` to check if the error message works like before.

### DIFF
--- a/commitizen/commands/check.py
+++ b/commitizen/commands/check.py
@@ -79,20 +79,19 @@ class Check:
         if not commits:
             raise NoCommitsFoundError(f"No commit found with range: '{self.rev_range}'")
 
-        pattern = self.cz.schema_pattern()
-        displayed_msgs_content = "\n".join(
-            [
-                f'commit "{commit.rev}": "{commit.message}"'
-                for commit in commits
-                if not self.validate_commit_message(commit.message, pattern)
-            ]
+        pattern = re.compile(self.cz.schema_pattern())
+        invalid_msgs_content = "\n".join(
+            f'commit "{commit.rev}": "{commit.message}"'
+            for commit in commits
+            if not self._validate_commit_message(commit.message, pattern)
         )
-        if displayed_msgs_content:
+        if invalid_msgs_content:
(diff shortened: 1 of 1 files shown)


Lee-W (commented):
I'm somewhat worried about this one. I guess there might been users actually use the message to check. Probably could make it a deprecated thing and change in v5

bearomorphism:
> I'm somewhat worried about this one. I guess there might been users actually use the message to check. Probably could make it a deprecated thing and change in v5

Then let me revert the `capitalized` part of this PR.

bearomorphism:
> I'm somewhat worried about this one. I guess there might been users actually use the message to check. Probably could make it a deprecated thing and change in v5

Fixed. You meant the error message, right?

Lee-W (commented):
I'm not sure I get the purpose of this PR after the changes. It seems the pattern only compiles once before this PR. or is there anywhere else I missed?

commitizen/commands/check.py
```diff
@@ -135,14 +134,18 @@ def _filter_comments(msg: str) -> str:
...
             return self.allow_abort
 
         if any(map(commit_msg.startswith, self.allowed_prefixes)):
             return True
+
         if self.max_msg_length:
             msg_len = len(commit_msg.partition("\n")[0].strip())
             if msg_len > self.max_msg_length:
                 return False
-        return bool(re.match(pattern, commit_msg))
```
bearomorphism, line 156: > I'm not sure I get the purpose of this PR after the changes. It seems the pattern only compiles once before this PR. or is there anywhere else I missed?

@Lee-W Before this change, each `self._validate_commit_message` evaluates `re.match(pattern, _)` once, where `re.match` compiles the input string every time.

bearomorphism: Hence `self.cz.schema_pattern()` is compiled `len(commits)` times (assuming no early return)

Lee-W: got it. make sense. let's merge it 🙂

Lee-W (approved):

PR #1479, opened 2025-05-31
Add type hints for cli.py

### LINKED ISSUE #1477: Add type annotations for the arguments of the CLI entry point
### Description

In the CLI entry point (`cli.py`), all properties inside `args` are unknown to the IDE, which makes it harder for new developers to read and understand the code.

This is especially problematic when trying to quickly identify which function `args.func` actually refers to.

### Possible Solution

I suggest adding type hints to clarify which attributes correspond to which functions. Since this change only involves adding type annotations, it will not affect the end-user's runtime behavior. In theory, such a change is safe.

### Additional context

I suggest adding the following content.

```py
class Args(argparse.Namespace):
    config: str | None = None
    debug: bool = False
    name: str | None = None
    no_raise: str | None = None
    report: bool = False
    project: bool = False
    commitizen: bool = False
    verbose: bool = False
    func: type[
        commands.Init  # init
        | commands.Commit  # commit (c)
        | commands.ListCz  # ls
        | commands.Example  # example
        | commands.Info  # info
        | commands.Schema  # schema
        | commands.Bump  # bump
        | commands.Changelog  # changelog (ch)
        | commands.Check  # check
        | commands.Version  # version
    ]
```

### Additional context

_No response_

gbaian10:
## Description

Resolves #1477 
- #1477 

I hope this can be merged into the `refactors` branch instead of `master`.

## Checklist

### Code Changes

### Documentation Changes

  - [ ] Check if there are any broken links in the documentation

>
> ```text
> ```

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -8,7 +8,7 @@ from copy import deepcopy
 from functools import partial
 from pathlib import Path
 from types import TracebackType
-from typing import Any
+from typing import TYPE_CHECKING, Any, cast
 
 import argcomplete
 from decli import cli
@@ -596,8 +596,33 @@ def parse_no_raise(comma_separated_no_raise: str) -> list[int]:
     return no_raise_codes
 
 
+if TYPE_CHECKING:
+
+    class Args(argparse.Namespace):
+        config: str | None = None
+        debug: bool = False
+        name: str | None = None
+        no_raise: str | None = None
+        report: bool = False
+        project: bool = False
+        commitizen: bool = False
+        verbose: bool = False
+        func: type[
+            commands.Init  # init
+            | commands.Commit  # commit (c)
+            | commands.ListCz  # ls
+            | commands.Example  # example
+            | commands.Info  # info
+            | commands.Schema  # schema
+            | commands.Bump  # bump
+            | commands.Changelog  # changelog (ch)
+            | commands.Check  # check
+            | commands.Version  # version
+        ]
+
+
 def main():
-    parser = cli(data)
+    parser: argparse.ArgumentParser = cli(data)
     argcomplete.autocomplete(parser)
     # Show help if no arg provided
     if len(sys.argv) == 1:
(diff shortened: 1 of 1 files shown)


commitizen/cli.py
```diff
@@ -612,7 +637,7 @@ def main():
         # https://github.com/commitizen-tools/commitizen/issues/429
         # argparse raises TypeError when non exist command is provided on Python < 3.9
         # but raise SystemExit with exit code == 2 on Python 3.9
-        if isinstance(e, TypeError) or (isinstance(e, SystemExit) and e.code == 2):
```
bearomorphism, line 615: Is the deleted isinstance expression always true? Just want to doubly confirm

bearomorphism (commented):
👍👍👍

commitizen/cli.py
```diff
@@ -597,8 +597,33 @@ def parse_no_raise(comma_separated_no_raise: str) -> list[int]:
     return no_raise_codes
 
 
+if TYPE_CHECKING:
+
+    class Args(argparse.Namespace):
+        config: str | None = None
+        debug: bool = False
+        name: str | None = None
+        no_raise: str | None = None
```
Lee-W, line 606: We could probably add a comment mentioning that it will later be parsed as `list[int]`. A bit confusing at the first glance

commitizen/cli.py
```diff
@@ -597,8 +597,33 @@ def parse_no_raise(comma_separated_no_raise: str) -> list[int]:
...
+        report: bool = False
+        project: bool = False
+        commitizen: bool = False
+        verbose: bool = False
+        func: type[
```
Lee-W, line 611: This is nice!

commitizen/cli.py
Lee-W: According to the comments above, it seems we can remove `TypeError`. I just traced the code a bit. https://github.com/python/cpython/blob/558e27abf1f1e0f87262238bc1d9e84cca7743c6/Lib/argparse.py#L2582 it seems to be the case after 3.9

gbaian10: Does this mean changing it to this way? Only catching `SystemExit` to avoid a silent exit?

```py
try:
    args, unknown_args = parser.parse_known_args()
except SystemExit as e:
    # ... comments ...
    if e.code == 2:
        raise NoCommandFoundError()
    raise e
```

gbaian10: Fixed. I also removed the comments.

Since the project requires `Python >=3.9`, argparse only raises `SystemExit` when non-existent commands are provided, making `TypeError` handling unnecessary.

Lee-W (approved):

PR #1431, opened 2025-05-17
refactor(bump_rule): add bump_rule interface, DefaultBumpRule and its…

bearomorphism:
… tests

## Checklist

- [ ] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/)

### Code Changes

- [ ] Add test cases to all the changes you introduce
- [ ] Run `poetry all` locally to ensure this change passes linter check and tests
- [ ] Manually test the changes:
  - [ ] Verify the feature/bug fix works as expected in real-world scenarios
  - [ ] Test edge cases and error conditions
  - [ ] Ensure backward compatibility is maintained
  - [ ] Document any manual testing steps performed
- [ ] Update the documentation for the changes

### Documentation Changes

- [ ] Run `poetry doc` locally to ensure the documentation pages renders correctly

Issue #129, opened 2020-01-29
[feature/refactor] Use named capture group in bump_pattern to enable stricter check


Lee-W:
**Goal**
make regular expression pattern stricter so that we won't accidentally match things we don't need

**Description**
In [commitizen/cz/conventional_commits/conventional_commits.py#L33](https://github.com/Woile/commitizen/blob/command-changelog/commitizen/cz/conventional_commits/conventional_commits.py#L33) on `command-changelog` branch, I use named capture group so that we could use a stricter regular expression like `.*\n\nBREAKING CHANGE`. The benefit of it is that we don't have to break the whole commit message into lines like [commitizem/bump.py#L31](https://github.com/Woile/commitizen/blob/v1.16.0/commitizen/bump.py#L31). It can also avoid bump or generate changelog based on commit message like `fix ---  it does not follow the rule but still match the pattern`.
Another thought on this topic is that we probably merge the `bump_map` and `bump_pattern` into one some data class to store `name`(e.g., `break`), `pattern`(e.g., `.*\n\nBREAKING CHANGE`), `behavior`(e.g., `PATCH`).

Woile:
Not sure how to follow with this one. do we still need it?

Would the dataclass be internal or it should be provided for templating?

Lee-W:
The basic idea of the dataclass part is to make moving from `dict` to `dataclass` which might be clear.

The main idea of this issue is to use named capture group to make the regular expression stricter like how you implement `commit_parser`.

```python
# now
bump_pattern = r"^(BREAKING[\-\ ]CHANGE|feat|fix|refactor|perf)(\(.+\))?(!)?"

# named capture group
bump_pattern = r"(?P<MAJOR>^.*\n\nBREAKING[-]CHANGE.*|)|(?P<MINOR>^feat.*)|?(P<PATCH>^fix.*|^perf.*|^refactor.*)"
```

Woile:
Oh, I see, the `find_increment` would have to be completely refactored. 

Still I see some complications, for conventional commits how would you capture `BREAKING CHANGE` and `!` as breaking chagnes with a named group? I've tried a while ago with little success haha

Regarding the dataclasses I'd need an example to understand it better, for me a dict is usually clearer than anything, and can be easily converted to configuration if it's kept simple.

Lee-W:
> Still I see some complications, for conventional commits how would you capture `BREAKING CHANGE` and `!` as breaking chagnes with a named group? I've tried a while ago with little success haha

[...]

PR #1518, opened 2025-06-09
feat(bump_rule): add BumpRule, VersionIncrement enum, ConventionalCommitBumpRule

bearomorphism:
Related issue: #129
Original PR: #1431
Previous iteration of this PR (force-pushed; commit `9644e1a1`) covered a `Prerelease` enum that has now been dropped — see "Out of scope" below.

## Description

Refactor commitizen's bump-rule machinery to be strict, explicit, and pluggable, while staying backward-compatible with existing `cz` plugins and configurations.

### What changes

1. **New `BumpRule` protocol** (`commitizen/bump_rule.py`)
   - `BumpRule.extract_increment(commit_message, major_version_zero) -> VersionIncrement` — single-message → single-increment.
   - `ConventionalCommitBumpRule` — handles the conventional-commits spec (`feat`, `fix`, `perf`, `refactor`, `BREAKING CHANGE`/`BREAKING-CHANGE`, `feat!:` bang).
   - `CustomBumpRule` — drop-in replacement for the legacy `bump_pattern` + `bump_map` flow. Tries the modern named-group form first, then falls back to the legacy regex-key form so existing user configurations keep working.
   - `BaseCommitizen.bump_rule` is a new `cached_property` that returns `_bump_rule` if a plugin sets one, otherwise builds a `CustomBumpRule` from `bump_pattern`/`bump_map`/`bump_map_major_version_zero`.
   - `ConventionalCommitsCz._bump_rule = ConventionalCommitBumpRule()` opts the built-in into the new path.

2. **Consolidated `VersionIncrement` enum** (`commitizen/version_increment.py`)
   - Reuses the `IntEnum` already added in #1724 (`NONE < PATCH < MINOR < MAJOR`).
   - New helpers required by the bump-rule plumbing:
     - `safe_cast` (alias of existing `from_value`) — `str | object → VersionIncrement`, returns `NONE` for unknown.
     - `safe_cast_dict` — bulk-cast a `Mapping[str, object]` for `bump_map` conversion.
     - `get_highest_by_messages(messages, extract_increment)` — finds the highest increment across many commit messages, splitting multi-line messages and skipping `NONE`.
   - `__str__` continues to return `self.name` (`"MAJOR"` / `"MINOR"` / `"PATCH"`) so log lines and hook env vars are unchanged.

3. **`BaseVersion.bump()` accepts `VersionIncrement` directly**
   - Replaces the `Increment: Literal["MAJOR","MINOR","PATCH"]` string-typed parameter with the enum.
   - `Increment` alias is removed from `commitizen.version_schemes`.
   - `commitizen.commands.version.Version.__call__` is simplified (no more 9-line `if/elif` mapping enum back to string before calling `bump()`).

4. **`commands/bump.py` uses the new `BumpRule` flow**
   - `Bump._find_increment` now does:
     ```python
     return VersionIncrement.get_highest_by_messages(
         (commit.message for commit in commits),
         lambda x: self.cz.bump_rule.extract_increment(x, is_major_version_zero),
     )
     ```
     instead of building a `bump_map` and calling `bump.find_increment`.
   - Pre/post-bump hooks still receive `CZ_PRE_INCREMENT` / `CZ_POST_INCREMENT` as `"MAJOR"` / `"MINOR"` / `"PATCH"` strings (or empty when no bump). The `VersionIncrement.NONE` case is mapped back to `None` at the call site so env-var output matches master byte-for-byte.

5. **`commitizen.bump.find_increment` is removed.** Only callers were the deleted `tests/test_bump_find_increment.py`. Coverage is now provided by the per-rule tests in `tests/test_bump_rule.py` and the bump-flow tests in `tests/commands/test_bump_command.py`.

6. **`commitizen.defaults.MAJOR` / `MINOR` / `PATCH`** are now exposed only via PEP-562 `__getattr__` with a `DeprecationWarning` directing users to `commitizen.version_increment.VersionIncrement`. The string values (`"MAJOR"`, `"MINOR"`, `"PATCH"`) are unchanged. **Plus a regression test** in `tests/test_deprecated.py` to lock the contract in (a previous iteration shipped a buggy non-tuple form that crashed on access).

7. **Docs.** `docs/commands/bump.md` gains a "Custom bump" section describing `bump_pattern` / `bump_map` / `bump_map_major_version_zero` with both the new named-group form and the legacy regex-key form, all in valid Python `dict` syntax.

### Out of scope (explicitly dropped from a previous iteration of this PR)

- **`Prerelease` is NOT converted to an Enum.** It stays as `Prerelease: TypeAlias = Literal["alpha", "beta", "rc"]`. Reasoning:
  - The CLI already restricts `--prerelease` to `{"alpha", "beta", "rc"}` via argparse `choices`, and the `Literal` already pins the type-level contract.
  - No internal code path passes other strings.
  - Converting to an `Enum` would break external Python callers of `BaseVersion.bump(prerelease="alpha")` for negligible gain inside the codebase.
  - Could be revisited later as its own PR (e.g. `class Prerelease(str, Enum)` with a backward-compatible mixin).

### Backward compatibility

- All third-party `cz` plugins that define `bump_pattern` + `bump_map` keep working: the `CustomBumpRule` fallback path matches the legacy `bump.find_increment` semantics for both named-group and regex-key bump maps.
- **`bump_map_major_version_zero` is optional.** Plugins that only define `bump_pattern` and `bump_map` (without `bump_map_major_version_zero`) continue to work — `CustomBumpRule` validates this lazily and only raises `NoPatternMapError` when an actual bump is attempted with `major_version_zero=True`. This mirrors master's `commands/bump.py::_find_increment` behaviour exactly.
- **Iteration order is preserved** through the `bump_map` → `safe_cast_dict` → `CustomBumpRule` pipeline. Order matters for the legacy fallback path (it returns the first matching pattern), and tests now lock the order-sensitivity in.
- `commitizen.defaults.MAJOR` / `MINOR` / `PATCH` keep working via `__getattr__` (with a deprecation warning).
- Hook env vars (`CZ_PRE_INCREMENT`, `CZ_POST_INCREMENT`) and `cz bump` stdout output are byte-for-byte identical.
- Default `BUMP_MAP` values still serialize as `"MAJOR"` / `"MINOR"` / `"PATCH"` strings via `str(VersionIncrement.X)`.

### Diff at a glance

```
 19 files changed, 1548 insertions(+), 518 deletions(-)
```

## Checklist

### Code Changes

- [x] Add test cases to all the changes you introduce — `tests/test_bump_rule.py` (635 new lines), expanded `tests/test_version_increment.py`, regression entry in `tests/test_deprecated.py`.
  - `uv run ruff check` — clean
  - `uv run ruff format --check` — clean
  - `uv run mypy commitizen tests` — clean
  - `uv run pytest` — 1322 passed, 3 skipped, 4 GPG-deselected on Windows
  - [x] Test edge cases and error conditions (`major_version_zero`, multi-line messages, custom `bump_pattern` fallback path)
  - [x] Document any manual testing steps performed (existing test suite covers them)

### Documentation Changes

- [x] Run `uv run poe doc` locally to ensure the documentation pages render correctly

## Additional Context

- The original branch had been sitting in conflict for ~11 months on top of `v4.10.0`. This iteration was rebuilt fresh on top of current `master` (`4d99415a`).
- Maintainer feedback from August 2025 (Lee-W): *"we can likely add some of the logic incrementally while keeping compatibility for a few minor versions"* — this iteration explicitly trades the more invasive `Prerelease` enum change for incrementality.
- A self code-review pass found and fixed three issues before publishing:
  1. `commitizen.defaults.MAJOR/MINOR/PATCH` were stored as bare strings instead of `(value, replacement)` tuples — `from commitizen.defaults import MAJOR` would have crashed with `ValueError`. Fixed with regression test.
  2. `bump_map_major_version_zero` was incorrectly required at construction time, breaking plugins that only defined `bump_pattern` + `bump_map`. Fixed by making it optional and validating lazily, matching master's behaviour exactly. Regression tests added.
  3. The legacy fallback path's order-dependence on dict iteration is now explicitly tested and documented.


commitizen/bump_rule.py
```diff
@@ -0,0 +1,253 @@
...
+            get_increment(line)
+            for message in commit_messages
+            for line in message.split("\n")
+        )
+
+    @staticmethod
+    def get_highest(
+        increments: Iterable[VersionIncrement | None],
+    ) -> VersionIncrement | None:
+        return max(filter(None, increments), default=None)
```
Lee-W: I'm thinking of extending `__gt__` and `__lt__` to remove this filter. WDYT? Not sure whether making `None` the minimum value makes sense or not 🤔

commitizen/bump_rule.py
```diff
@@ -0,0 +1,253 @@
...
+
+    This protocol specifies the contract that all version bump rule implementations must follow.
+    It defines how commit messages should be analyzed to determine the appropriate semantic
+    version increment.
+
+    The protocol is used to ensure consistent behavior across different bump rule implementations,
+    such as conventional commits or custom rules.
+    """
+
+    def get_increment(
```
Lee-W: `extract_increment` seems to be more accurate

commitizen/bump_rule.py
```diff
@@ -0,0 +1,253 @@
...
+        self, commit_message: str, major_version_zero: bool
+    ) -> VersionIncrement | None:
+        """Determine the version increment based on a commit message.
+
+        This method analyzes a commit message to determine what kind of version increment
+        is needed according to the Conventional Commits specification. It handles special
```
Lee-W: Not sure we should mention Conventional commits here. We should be able to support other commit spec

commitizen/bump_rule.py
```diff
@@ -0,0 +1,253 @@
...
+                - MINOR: For breaking changes when major_version_zero is True, or for new features
+                - PATCH: For bug fixes, performance improvements, or refactors
+                - None: For commits that don't require a version bump (docs, style, etc.)
+        """
+
+
+class ConventionalCommitBumpRule(BumpRule):
+    _BREAKING_CHANGE_TYPES = set(["BREAKING CHANGE", "BREAKING-CHANGE"])
+    _MINOR_CHANGE_TYPES = set(["feat"])
+    _PATCH_CHANGE_TYPES = set(["fix", "perf", "refactor"])
```
Lee-W:
```suggestion
    _MINOR_CHANGE_TYPES = {"feat"}
    _PATCH_CHANGE_TYPES = {"fix", "perf", "refactor"}
```

commitizen/bump_rule.py
```diff
@@ -0,0 +1,253 @@
...
+            VersionIncrement | None: The type of version increment needed:
+                - MAJOR: For breaking changes when major_version_zero is False
+                - MINOR: For breaking changes when major_version_zero is True, or for new features
+                - PATCH: For bug fixes, performance improvements, or refactors
+                - None: For commits that don't require a version bump (docs, style, etc.)
+        """
+
+
+class ConventionalCommitBumpRule(BumpRule):
+    _BREAKING_CHANGE_TYPES = set(["BREAKING CHANGE", "BREAKING-CHANGE"])
```
Lee-W:
```suggestion
    _BREAKING_CHANGE_TYPES = {"BREAKING CHANGE", "BREAKING-CHANGE"}
```

commitizen/commands/bump.py
```diff
@@ -144,21 +145,14 @@ def _is_initial_tag(
...
+        is_major_version_zero = self.bump_settings["major_version_zero"]
 
-        if not bump_map or not bump_pattern:
-            raise NoPatternMapError(
-                f"'{self.config.settings['name']}' rule does not support bump"
-            )
-        return bump.find_increment(commits, regex=bump_pattern, increments_map=bump_map)
+        return VersionIncrement.get_highest_by_messages(
+            (commit.message for commit in commits),
+            lambda x: self.cz.bump_rule.get_increment(x, is_major_version_zero),
```
Lee-W: this might be a breaking change for other non-standard `cz`s

Lee-W: We'll need a way to keep backward compatibility or move it to 5.0.0 instead (but that would be a pity since the PR looks os good)

commitizen/defaults.py
```diff
@@ -113,31 +114,27 @@ class Settings(TypedDict, total=False):
...
 
-MAJOR = "MAJOR"
-MINOR = "MINOR"
-PATCH = "PATCH"
-
 CHANGELOG_FORMAT = "markdown"
 
 BUMP_PATTERN = r"^((BREAKING[\-\ ]CHANGE|\w+)(\(.+\))?!?):"
-BUMP_MAP = OrderedDict(
+BUMP_MAP = dict(
```
Lee-W, line 127: I guess we'll be able to remove them in the next major version?

Lee-W (commented):
This is definitely a great pull request that cleans up a lot of legacy code. However, we need to be cautious about maintaining backward compatibility. We can likely add some of the logic incrementally while keeping compatibility for a few minor versions. After that, we can bump the major version and remove the compatibility handling code.

commitizen/bump_rule.py
bearomorphism: Let me check when I have a moment. This PR has been a while

bearomorphism: I believe this does not work because comparison between `None` and `None` is invalid.

```
Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 17.0.0 (clang-1700.0.13.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> None < None
  File "<python-input-0>", line 1, in <module>
    None < None
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
>>> max([None, None])
  File "<python-input-1>", line 1, in <module>
    max([None, None])
    ~~~^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
>>> 
```

bearomorphism: Or we can just add `NONE` in as part of the class

commitizen/bump_rule.py
bearomorphism: Let me move it to `ConventionalCommitBumpRule`

bearomorphism:
Can we do squash rebase on this PR when it is ready for merge?

commitizen/bump_rule.py
Lee-W: Thanks for verifying! I think we should keep it as it is.

Lee-W:
> Can we do squash rebase on this PR when it is ready for merge?

yep, I can do that

commitizen/bump_rule.py
bearomorphism: I added `NONE` as part of the class `VersionIncrement` so we can get rid of `VersionIncrement | None` type.

commitizen/defaults.py
bearomorphism: probably yes

commitizen/bump_rule.py
bearomorphism: Done.

commitizen/commands/bump.py
bearomorphism: Could you provide more details about why this might be a breaking change for non-standard `cz` s? I am not familiar with non-standard cz. Thanks

bearomorphism:
What are the next steps for this PR? If there are backward compatibility issues, I can adjust it so that this PR can be checked in. Thanks

commitizen/defaults.py
bearomorphism: Added comments about removing them in v5

commitizen/bump_rule.py
bearomorphism, line 1: Core feature change

bearomorphism:
We can get back to this PR after #1598

bearomorphism:
@Lee-W It would be great if we can complete this big PR. I can fix those backward-compatibility issues, but I need more details so I can address the issues.

bearomorphism:
(I've resolved merge conflicts for several times for this PR)

Lee-W:
This would fall into `4.11.0`. Let's foucs on merging the non-features PRs first and then get back to this one.

bearomorphism:
A lot of conflicts... I will probably create another branch and do the change again...

bearomorphism:
will do a human review later...

Issue #1498, opened 2025-06-03
`_valid_command_argument`


bearomorphism:
### Description

I just think it's not a good design to validate and reassign things in a function. The function name itself is confusing.

https://github.com/commitizen-tools/commitizen/blob/47beadc563d8272b64f88ddbf3e9a37351b8f535/commitizen/commands/check.py#L51-L62

bearomorphism:
Closing this since #1519 is merged into master branch.

PR #1472, opened 2025-05-30
refactor(cz): better typing and refactor message method

bearomorphism:
## Checklist

- [ ] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/)

### Code Changes

- [ ] Add test cases to all the changes you introduce
- [ ] Run `poetry all` locally to ensure this change passes linter check and tests
- [ ] Manually test the changes:
  - [ ] Verify the feature/bug fix works as expected in real-world scenarios
  - [ ] Test edge cases and error conditions
  - [ ] Ensure backward compatibility is maintained
  - [ ] Document any manual testing steps performed
- [ ] Update the documentation for the changes

### Documentation Changes

- [ ] Run `poetry doc` locally to ensure the documentation pages renders correctly
  - [ ] Check if there are any broken links in the documentation

> When running `poetry doc`, any broken internal documentation links will be reported in the console output like this:
>
> ```text
> INFO    -  Doc file 'config.md' contains a link 'commands/bump.md#-post_bump_hooks', but the doc 'commands/bump.md' does not contain an anchor '#-post_bump_hooks'.
> ```

PR #1457, opened 2025-05-25
Refactors

Lee-W:
## Checklist

- [ ] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/)

### Code Changes

- [ ] Add test cases to all the changes you introduce
- [ ] Run `poetry all` locally to ensure this change passes linter check and tests
- [ ] Manually test the changes:
  - [ ] Verify the feature/bug fix works as expected in real-world scenarios
  - [ ] Test edge cases and error conditions
  - [ ] Ensure backward compatibility is maintained
  - [ ] Document any manual testing steps performed
- [ ] Update the documentation for the changes

### Documentation Changes

- [ ] Run `poetry doc` locally to ensure the documentation pages renders correctly

PR #1519, opened 2025-06-09
feat: add check against default branch

bearomorphism:
## Description

Related PR: #1400 
Wait for #1472 and #1457

Related issue: #1498 

- Added `cz check --default-range`.

### Some TODOs

- [x] Add tests
- [x] Documentation

## Checklist

### DIFF
--- a/commitizen/cli.py
+++ b/commitizen/cli.py
@@ -472,6 +472,13 @@ data = {
                         "exclusive_group": "group1",
                     },
                     {
+                        "name": ["-d", "--use-default-range"],
+                        "action": "store_true",
+                        "default": False,
+                        "help": "check from the default branch to HEAD. e.g, refs/remotes/origin/master..HEAD",
+                        "exclusive_group": "group1",
+                    },
+                    {
                         "name": ["-m", "--message"],
                         "help": "commit message that needs to be checked",
                         "exclusive_group": "group1",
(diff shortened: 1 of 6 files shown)


commitizen/cli.py
```diff
@@ -474,6 +474,13 @@ def __call__(
                         "help": "a range of git rev to check. e.g, master..HEAD",
                         "exclusive_group": "group1",
                     },
+                    {
+                        "name": ["-d", "--default-range"],
```
Lee-W: `--use-default-range` or `--check-default-range` might sound more like a bool.

not sure whehter adding a `default_branch` config to the `project.toml` would make it better. could be next pr though

Lee-W (commented):
other than the naming issue, everything looks great. a wonderful feature we should have long ago

commitizen/cli.py
bearomorphism: use `--use-default-range`

tests/test_git.py
```diff
@@ -480,3 +480,22 @@ def test_create_commit_cmd_string(mocker, os_name, committer_date, expected_cmd)
     mocker.patch("os.name", os_name)
     result = git._create_commit_cmd_string("", committer_date, "temp.txt")
     assert result == expected_cmd
+
+
+def test_get_default_branch(mocker: MockFixture):
```
Lee-W: This should be separate as 2 test cases instead

bearomorphism: done, please review

PR #1539, opened 2025-06-12
refactor(changelog): simplify logic for get_oldest_and_newest_rev

bearomorphism:
## Description

- Refactor `get_oldest_and_newest_rev`
- Since we only use the latest element of the result of `get_smart_range`, I believe we can replace it with a simpler function.
- The implementation of the function `get_next_tag_name_after_version` is based on an assumption that "all GitTags are unique". If the oldest tag's name is the input `version`, then it returns `None`. The behavior aligns with "Return `None` for  `oldest_rev` if the oldest tag is the last tag in the list and matches the requested oldest tag"

## Checklist

### Code Changes


commitizen/changelog.py
```diff
@@ -281,6 +281,18 @@ def incremental_build(
...
+        if tags[idx].name == version:
+            if idx == len(tags) - 1:
+                return None
+            return tags[idx + 1].name
+        idx -= 1
+    raise NoCommitsFoundError(f"Could not find a valid revision range. {version=}")
+
+
+# TODO: unused, deprecate this?
 def get_smart_tag_range(
```
bearomorphism, line 294: Not sure if removing the function is a breaking change.

commitizen/changelog.py
```diff
@@ -281,6 +282,16 @@ def incremental_build(
     return output_lines
 
 
+def get_next_tag_name_after_version(tags: Iterable[GitTag], version: str) -> str | None:
+    a, b = tee(chain((tag.name for tag in tags), [None]))
+    next(b, None)
+    try:
+        return next(y for x, y in zip(a, b) if x == version)
+    except StopIteration:
+        raise NoCommitsFoundError(f"Could not find a valid revision range. {version=}")
```
bearomorphism: Thanks @gbaian10 for reminding me here we don't have to use a while loop.

commitizen/changelog.py
Lee-W: let's mark it as deprecated and remove in the next version

Lee-W: let's add a deprecation warning

tests/test_changelog.py
```diff
@@ -1535,6 +1535,24 @@ def test_get_smart_tag_range_returns_an_extra_for_a_single_tag(tags):
     assert 2 == len(res)
 
 
+def test_get_next_tag_name_after_version(tags):
+    # Test finding next tag after a version
+    next_tag_name = changelog.get_next_tag_name_after_version(tags, "v1.2.0")
```
Lee-W, line 1540: It looks like previous tag instead of next tag 🤔

bearomorphism: Should I rename it?

commitizen/changelog.py
bearomorphism: done

tests/test_changelog.py
Lee-W: if you don't have the issue with the name previous_verison than let's do it :)

bearomorphism: k

Lee-W (approved):

PR #1540, opened 2025-06-12
refactor(changelog): shorten generate_tree_from_commits and use set to check used tags

bearomorphism:
## Checklist

### Code Changes

### DIFF
--- a/commitizen/changelog.py
+++ b/commitizen/changelog.py
@@ -32,6 +32,7 @@ from collections import OrderedDict, defaultdict
 from collections.abc import Generator, Iterable, Mapping, Sequence
 from dataclasses import dataclass
 from datetime import date
+from itertools import chain
 from typing import TYPE_CHECKING, Any
 
 from jinja2 import (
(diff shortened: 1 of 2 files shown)


commitizen/changelog.py
```diff
@@ -88,33 +89,32 @@ def generate_tree_from_commits(
...
+    )
+
+    used_tags: set[GitTag] = set()
+    if current_tag:
+        used_tags.add(current_tag)
+        if current_tag.name:
+            current_tag_name = current_tag.name
+            current_tag_date = current_tag.date
+
+    commit_tag: GitTag | None = None
```
Lee-W, line 109: this line is probably not needed?

bearomorphism: This variable is also used at the end of this function. If this line is removed, `commit_tag` at the end of this function becomes unbounded. If `commits` is empty, then `commit_tag` is not initialized.

Lee-W: got it. let's merge it then 🚀

PR #1545, opened 2025-06-12
fix(ExitCode): add from_str in ExitCode and replace parse_no_raise with it, warn if the error code is not in range

bearomorphism:
## Description

- Now we validate if the error code exists given a decimal string

## Checklist

### Code Changes

### DIFF
--- a/commitizen/exceptions.py
+++ b/commitizen/exceptions.py
@@ -1,10 +1,12 @@
-import enum
+from __future__ import annotations
+
+from enum import IntEnum
 from typing import Any
 
 from commitizen import out
 
 
-class ExitCode(enum.IntEnum):
+class ExitCode(IntEnum):
     EXPECTED_EXIT = 0
     NO_COMMITIZEN_FOUND = 1
     NOT_A_GIT_PROJECT = 2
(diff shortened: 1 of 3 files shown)


commitizen/exceptions.py
```diff
@@ -39,6 +41,23 @@ class ExitCode(enum.IntEnum):
     CONFIG_FILE_IS_EMPTY = 31
     COMMIT_MESSAGE_LENGTH_LIMIT_EXCEEDED = 32
 
+    @classmethod
+    def from_str(cls, value: str) -> ExitCode | None:
+        if value.isdecimal():
+            try:
+                return cls(int(value))
+            except ValueError:
```
Lee-W: instead of returning `None`, I think we should raise error and handles it in `commitizen/cli.py` instead

Lee-W: I know it might make the code longer, but makes more sense IMO

bearomorphism: k

bearomorphism: fixed.

bearomorphism: oh I misunderstood. I'll fix it

bearomorphism: fixed.

Lee-W (approved):

PR #1547, opened 2025-06-12
refactor(git): remove redundant if branch

bearomorphism:
## Checklist

### Code Changes

### DIFF
--- a/commitizen/git.py
+++ b/commitizen/git.py
@@ -328,6 +328,4 @@ def _get_log_as_str_list(start: str | None, end: str, args: str) -> list[str]:
     c = cmd.run(command)
     if c.return_code != 0:
         raise GitCommandError(c.err)
-    if not c.out:
-        return []
     return c.out.split(f"{delimiter}\n")


commitizen/git.py
```diff
@@ -328,6 +328,4 @@ def _get_log_as_str_list(start: str | None, end: str, args: str) -> list[str]:
     c = cmd.run(command)
     if c.return_code != 0:
         raise GitCommandError(c.err)
-    if not c.out:
```
Lee-W, line 331: I'm not following. Could you please share why this is not needed? Thanks

bearomorphism: 1. `c.out` is `str`
2. `x` is `str` and `x` is falsy => `x == ""`
3. `"".split(sep) == []` for all `sep` if `sep` is a string.

If `c.out` is falsy, then `c.out.split(f"{delimiter}\n")` is already `[]`. With or without the `if` does not affect the output of the function.

Lee-W: right, I tried to trace the source code a bit and verify it will always be str. Let's merge it.

Lee-W (approved):

Issue #1571, opened 2025-08-07
Add config option for line length warning (`-l/--message-length-limit`)


mattgebert:
### Description

In issue #191, it was suggested to be able to add command line arguments to commitizen to ensure the correct line length, implemented in #1076. This was further updated in #1178  to allow the same argument to be used in `check` in addition to `commit`. 

Examples:

    cz commit -l/--message-length-limit N

or

    cz check -l N -m MESSAGE

However, the is not yet support in [config](https://commitizen-tools.github.io/commitizen/config/) or [customisation](https://commitizen-tools.github.io/commitizen/customization/) to add such a default argument without requiring the command line specification.

### Possible Solution

An additional config option should exist in the configuration file to optionally define a commit line length.
This would then be utilised by both commitizen pre-commit [commit-msg, pre-push] hooks and command line invocations on the root directory of a project (i.e. with the relevant pyproject.toml for a python package).

    pyproject.toml
    
    [tool.commitizen]
    name = "cz_customize"
    
    [tool.commitizen.customize]
      ... 
      message_length_limit = 72  # or l=72?
    
Then allowing the default parameter to propagate as follows:

    cz commit
    >>> fails / succeeds depending on message length

    cz check -m MESSAGE
    >>> fails / succeeds depending on message length

Lee-W:
looks like a valid feature. Thanks for suggestion!

Narwhal-fish:
Hi @Lee-W , I'd be happy to help with this one. May I take it?

Narwhal-fish:
Hi @Lee-W, I'd be happy to help with this one. May I take it?

Lee-W:
yep, sure :)

bearomorphism:
I'm closing this issue as completed. Thanks @Narwhal-fish :)

PR #1574, opened 2025-08-12
feat: add config option for line length warning

Narwhal-fish:
## Description
This PR implements configuration file support for the `message_length_limit` option, addressing issue #1571 . Previously, users could only set commit message length limits via command-line arguments (`-l/--message-length-limit`). Now they can define a default limit in their configuration file.

## Checklist

### Code Changes

### Documentation Changes

## Expected Behavior
Users can now set default commit message length limits in their configuration file:

```toml
[tool.commitizen]
message_length_limit = 72
```

1. Test Configuration File Support
```bash
# Add the following to your pyproject.toml:
[tool.commitizen]
message_length_limit = 30

# Test with long message (should fail)
cz check -m "fix: this is a very long commit message that exceeds 30 characters"

# Test with short message (should pass)  
cz check -m "fix: short message"
```
2. Test CLI Override
```bash
# CLI argument should override config setting
cz check -m "fix: this message is longer than 30 chars but shorter than 60" -l 60

# Disable limit with CLI
cz check -m "fix: very long message that would normally fail config limit" -l 0
```


Narwhal-fish:
Sorry, I accidentally closed this pull request.

Lee-W:
let's rebase from `v4-9-test`

Narwhal-fish:
I have rebased my branch onto v4-9-0.
I’m not very familiar with Git, so please let me know if I missed anything.

Lee-W:
> I have rebased my branch onto v4-9-0. I’m not very familiar with Git, so please let me know if I missed anything.

You've done it perfectly 🙂 I'll take a look when bandwidth allow

commitizen/commands/check.py
```diff
@@ -40,7 +41,13 @@ def __init__(self, config: BaseConfig, arguments: CheckArgs, *args: object) -> N
...
             arguments.get("allow_abort", config.settings["allow_abort"])
         )
-        self.max_msg_length = arguments.get("message_length_limit", 0)
+
+        # Use command line argument if provided, otherwise use config setting
+        cmd_length_limit = arguments.get("message_length_limit")
+        if cmd_length_limit is None:
+            self.max_msg_length = config.settings.get("message_length_limit", 0)
+        else:
+            self.max_msg_length = cmd_length_limit
```
bearomorphism:
```suggestion
        self.max_msg_length = arguments.get("message_length_limit", config.settings.get("message_length_limit", 0))
```

Same for `commit.py`

commitizen/commands/check.py
```diff
@@ -154,6 +161,11 @@ def _validate_commit_message(
         if self.max_msg_length:
             msg_len = len(commit_msg.partition("\n")[0].strip())
             if msg_len > self.max_msg_length:
-                return False
+                raise CommitMessageLengthExceededError(
+                    f"commit validation: failed!\n"
+                    f"commit message length exceeds the limit.\n"
+                    f'commit "": "{commit_msg}"\n'
```
bearomorphism: what does `""` mean here

commitizen/commands/commit.py
```diff
@@ -81,7 +81,11 @@ def _prompt_commit_questions(self) -> str:
 
         message = cz.message(answers)
         message_len = len(message.partition("\n")[0].strip())
-        message_length_limit = self.arguments.get("message_length_limit", 0)
+
+        message_length_limit = self.arguments.get("message_length_limit")
+        if message_length_limit is None:
+            message_length_limit = self.config.settings.get("message_length_limit", 0)
```
bearomorphism: same

commitizen/defaults.py
```diff
@@ -108,6 +109,7 @@ class Settings(TypedDict, total=False):
     "always_signoff": False,
     "template": None,  # default provided by plugin
     "extras": {},
+    "message_length_limit": 0,  # 0 for no limit
```
bearomorphism: Why don't we just use `None` for no limit?

commitizen/commands/check.py
bearomorphism: I guess you forgot to put something like commit hash here in the error message

commitizen/commands/check.py
Narwhal-fish: noted!

commitizen/commands/commit.py
Narwhal-fish: noted!

commitizen/commands/check.py
Narwhal-fish: You’re right. I referenced the error handling for invalid messages and forgot to add the commit hash. I’ll make the correction.

commitizen/defaults.py
Narwhal-fish: Initially, I set it to 0 simply because I thought commit messages wouldn't have a 0-character scenario, but your suggestion of using None is more clear. I've made this change accordingly.

Lee-W:
we might need to resolve the conflict first. Thanks!

bearomorphism:
Hi @Narwhal-fish , could you help to rebase your branch when you have a moment?

You can run

```
git fetch upstream
git rebase upstream master -i
```

and then drop all commits but yours.

commitizen/defaults.py
bearomorphism: I see why you used 0 for no limit. It is because you wanted to align with the original design.

I will probably be a breaking change if we change the default value of this option. wdyt @Lee-W ?

Lee-W: what would be the behavior or 0 now? if 0 still works fine, it's probably ok

commitizen/commands/commit.py
```diff
@@ -81,8 +81,13 @@ def _prompt_commit_questions(self) -> str:
...
         message_len = len(message.partition("\n")[0].strip())
-        message_length_limit = self.arguments.get("message_length_limit", 0)
-        if 0 < message_length_limit < message_len:
+
+        message_length_limit = self.arguments.get(
+            "message_length_limit",
+            self.config.settings.get("message_length_limit", None),
+        )
+
+        if message_length_limit is not None and message_len > message_length_limit:
```
Lee-W:
```suggestion
        if message_length_limit and message_len > message_length_limit:
```

this can check both 0 and None

Lee-W (approved):

commitizen/defaults.py
Lee-W: I think 0 and None should both be no limit

PR #1582, opened 2025-09-04
fix: cargo workspaces

### LINKED ISSUE #1581: Wrong handling of cargo workspaces
### Description

The pull request #1402 added the version update of Cargo.lock, which is indeed necessary.
However, the implementation of the workspace feature of cargo is wrong and can lead to crash when bumping the version.

### Steps to reproduce

1. Create a rust project (`cargo init`)
2. Replace the `[package]` by a workspace (rewrite `Cargo.toml`) :
```toml
[workspace]
members = ["whatever"]
resolver = "2"

[worspace.package]
description = ""
edition = "2024"
license = ""
version = "0.1.0"
```
3. Create a `whatever` crate (`cargo new whatever`)
4. `cz init`
5. `git add . && git commit -m "feat: first commit"`
6. `cz bump --yes`

### Current behavior

Right now, the code expects the workspace to have a "name" key, [which is not a standard workspace key](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table). A crash occurs if the `Cargo.toml` doesn't have a `[package]` but has a `[workspace.package]` without a "name" key.

### Desired behavior

The code should not expect any "name" key. Instead, when a workspace is used, it should :
1. read members
2. resolve potential glob
3.  for each workspace member, check the Cargo.toml. If it contains the "version.workspace = true" line, this means the version is inherited from the workspace version and should be changed in the Cargo.lock

### Screenshots

_No response_

### Environment

Commitizen Version: 4.8.3
Python Version: 3.12.3
Operating System: Linux

vemilano:
## Description
Closes #1581 

## Checklist

### Code Changes

### Documentation Changes

### DIFF
--- a/commitizen/providers/cargo_provider.py
+++ b/commitizen/providers/cargo_provider.py
@@ -73,20 +73,19 @@ class CargoProvider(TomlProvider):
 
             for member in workspace_members:
                 for path in glob.glob(member, recursive=True):
-                    if not matches_exclude(path, excluded_workspace_members):
-                        cargo_file = Path(path) / "Cargo.toml"
-                        cargo_toml_content = tomlkit.parse(cargo_file.read_text())
-                        try:
-                            version_workspace = cargo_toml_content["package"][
-                                "version"
-                            ][  # type: ignore[index]
-                                "workspace"
-                            ]
-                            if version_workspace is True:
-                                package_name = cargo_toml_content["package"]["name"]  # type: ignore[index]
-                                members_inheriting.append(package_name)
-                        except tomlkit.exceptions.NonExistentKey:
-                            continue
+                    if matches_exclude(path, excluded_workspace_members):
+                        continue
+                    cargo_file = Path(path) / "Cargo.toml"
+                    cargo_toml_content = tomlkit.parse(cargo_file.read_text())
+                    try:
+                        version_workspace = cargo_toml_content["package"]["version"][  # type: ignore[index]
+                            "workspace"
+                        ]
+                        if version_workspace is True:
+                            package_name = cargo_toml_content["package"]["name"]  # type: ignore[index]
+                            members_inheriting.append(package_name)
+                    except tomlkit.exceptions.NonExistentKey:
+                        continue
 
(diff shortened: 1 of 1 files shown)


woile:
Thanks for the PR! It's great, although I think it's a bit more complex. 
I would like to see a test covering this example: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

```toml
[workspace]
members = ["member1", "path/to/member2", "crates/*"]
exclude = ["crates/foo", "path/to/other"]
```

tests/providers/test_cargo_provider.py
```diff
@@ -23,13 +24,19 @@
 
 CARGO_WORKSPACE_TOML = """\
 [workspace.package]
-name = "whatever"
+members = ["whatever"]
```
woile: I think memebers shouldn't be inside `[workspace.package]` but instead of `[workspace]`

vemilano: Good catch !

commitizen/providers/cargo_provider.py
```diff
@@ -55,27 +63,30 @@ def set_lock_version(self, version: str) -> None:
...
-                        version_workspace = cargo_toml_content["package"]["version"][  # type: ignore[index]
-                            "workspace"
-                        ]
-                        if version_workspace is True:
-                            package_name = cargo_toml_content["package"]["name"]  # type: ignore[index]
-                            members_inheriting.append(package_name)
-                    except tomlkit.exceptions.NonExistentKey:
-                        continue
+                for path in glob.glob(member, recursive=True):
+                    if not matches_exclude(path, excluded_workspace_members):
```
woile: To avoid too much indentation, maybe we can do this?
```suggestion
                    if matches_exclude(path, excluded_workspace_members):
                        continue
```

Lee-W:
I'll try to take a look next week but don't need to wait for me. Feel free to just merge it :)

woile:
Ok, let's give it a go then. Can you rebase @vemilano ? 🙏🏻

Issue #1580, opened 2025-09-02
`_bump_with_regex` function changes dependencies if they have same version number


be8312e4b036:
https://github.com/commitizen-tools/commitizen/blob/75a97e56bddac011e0d0eca3018fc0f873edbe4c/commitizen/bump.py#L131C4-L142C5

In a `pyproject.toml` of the form:
```
[project]
name = "myproject"
version = "1.2.3"
dependencies = [
    "package-a==5.0.0",
    "package-b==1.2.3"
]

[tool.commitizen]
name = "cz_conventional_commits"
version_files = ["pyproject.toml"]
tag_format = "v$version"
version_scheme = "pep440"
version_provider = "pep621"

[tool.uv.sources]
mygitpackage = { git = "...", rev = "v1.2.3" }
```
the packages `package-b` and `mygitpackage` get bumped as well as `myproject`  (when running `cz bump`) because the regex is just matching on the version number. These shouldn't be bumped and leads to errors when they are.

bearomorphism:
checking this one

bearomorphism:
I think we need to change the mechanism here. A regex contains too little information.

bearomorphism:
I want to do some cleanup before moving on. See #1594.

be8312e4b036:
@bearomorphism Just seen that #1594 has been merged now. Where is this fix in the timeline please?

bearomorphism:
@Lee-W I just found this file and you are the only author.

https://github.com/commitizen-tools/commitizen/blob/v4-10-1/tests/data/repeated_version_number.json

Shouldn't the case described in this issue be tested 🤔

bearomorphism:
@be8312e4b036 This documentation may solve your issue:
https://commitizen-tools.github.io/commitizen/commands/bump/#version_files

In your case, you can try adding `"pyproject.toml:version"` to your version files setting.

I agree the source code is a bit misleading...

bearomorphism:
Hi @be8312e4b036, is this still an issue?

Thanks!

mronidk:
> Hi [@be8312e4b036](https://github.com/be8312e4b036), is this still an issue?
> 
> Thanks!

I have experienced this issue as well. Dependencies with the same version as the pyproject.toml gets bumped. 

**EDIT:** Specifying the version in version files seems to resolve the issue, as you suggested @bearomorphism

bearomorphism:
Np

Lee-W:
[...]

PR #1594, opened 2025-09-11
refactor(bump): cleanup related to update_version_file

bearomorphism:
## Description

Try to address confusing variable names and function names, also reduced the number of function parameters for better readability.

Relate #1580 

Better test coverage for `update_version_in_file`:

- Covered return value
- Covered `check_consistency=True`
- Covered other `encoding` values

## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```

### DIFF
--- a/tests/test_bump_update_version_in_files.py
+++ b/tests/test_bump_update_version_in_files.py
@@ -2,6 +2,7 @@ from pathlib import Path
 from shutil import copyfile
 
 import pytest
+from _pytest.fixtures import FixtureRequest
 
 from commitizen import bump
 from commitizen.exceptions import CurrentVersionNotFoundError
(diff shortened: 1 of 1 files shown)


commitizen/bump.py
```diff
@@ -75,36 +75,46 @@ def update_version_in_files(
 
     Returns the list of updated files.
     """
-    # TODO: separate check step and write step
```
bearomorphism, line 78: I believe this TODO is already addressed

commitizen/bump.py
```diff
@@ -75,32 +75,42 @@ def update_version_in_files(
...
+                bump_line,
+                on_current_version_not_found,
             )
 
         # Write the file out again
         with smart_open(path, "w", encoding=encoding) as file:
-            file.write(version_file)
+            file.write(bumped_version_file_content)
         updated.append(path)
     return updated
```
bearomorphism: btw the return value is not covered in the unit tests for this function. I'm not sure if it's covered anywhere else.

commitizen/bump.py
bearomorphism: @Lee-W you left this TODO according to the git history, could you help to confirm this? Thanks.

commitizen/bump.py
```diff
@@ -115,31 +126,7 @@ def _files_and_regexes(patterns: Iterable[str], version: str) -> list[tuple[str,
         for file in iglob(filepath):
             out.add((file, regex))
 
-    return sorted(out)
-
-
-def _bump_with_regex(
```
bearomorphism, line 121: I believe we don't need this extra layer of abstraction.

commitizen/bump.py
```diff
@@ -64,8 +64,8 @@ def update_version_in_files(
     new_version: str,
     files: Iterable[str],
     *,
-    check_consistency: bool = False,
-    encoding: str = ENCODING,
+    check_consistency: bool,
+    encoding: str,
```
bearomorphism, line 68: I believe we don't need this default argument. These only introduce unnecessary complexity.

Lee-W: since it's only used in one place and we actually pass value into it. I'm good with this change

commitizen/bump.py
Lee-W: Doesn't look like so, but we probably don't need it either 🤔 Can't recall why this TODO was added

tests/test_bump_update_version_in_files.py
```diff
@@ -220,8 +236,158 @@ def test_update_version_in_globbed_files(commitizen_config_file, file_regression
...
+        encoding="utf-8",
+    )
+
+    # Verify return value is a list
+    assert isinstance(updated_files, list)
+
+    # Verify all files in the input are in the returned list
+    assert len(updated_files) == len(version_files)
+    for file_path in version_files:
+        assert file_path in updated_files
```
Lee-W:
```suggestion
    assert set(version_files).issubset(updated_files)
```

tests/test_bump_update_version_in_files.py
```diff
@@ -220,8 +236,158 @@ def test_update_version_in_globbed_files(commitizen_config_file, file_regression
...
+
+    # Verify the returned paths are strings
+    for file_path in updated_files:
+        assert isinstance(file_path, str)
```
Lee-W:
```suggestion
    assert all(isinstance(file_path, str) for file_path in updated_files)
```

Lee-W (approved):

PR #1596, opened 2025-09-12
refactor(ChangelogFormat): refactor get_metadata_from_file for better readability and type

bearomorphism:
… readability and better type

## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```

### DIFF
--- a/commitizen/changelog_formats/base.py
+++ b/commitizen/changelog_formats/base.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from abc import ABCMeta
-from typing import IO, Any, ClassVar
+from typing import IO, ClassVar
 
 from commitizen.changelog import Metadata
 from commitizen.config.base_config import BaseConfig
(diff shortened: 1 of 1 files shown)


commitizen/changelog_formats/base.py
```diff
@@ -40,34 +40,40 @@ def get_metadata(self, filepath: str) -> Metadata:
...
         unreleased_level: int | None = None
-        for index, line in enumerate(file):
-            line = line.strip().lower()
 
-            unreleased: int | None = None
-            if "unreleased" in line:
-                unreleased = self.parse_title_level(line)
+        lines = file.readlines()
+        for index, line in enumerate(line.strip().lower() for line in lines):
+            parsed_unreleased_level = self.parse_title_level(line)
```
bearomorphism, line 49: I think we can assume that `self.parse_title_level` is always a pure function? Then we don't have to call it twice here.

commitizen/changelog_formats/base.py
```diff
@@ -40,34 +40,40 @@ def get_metadata(self, filepath: str) -> Metadata:
...
-                meta.latest_version = parsed.version
-                meta.latest_version_tag = parsed.tag
-                meta.latest_version_position = index
+            if parsed_version_tag := self.parse_version_from_title(line):
+                out_metadata.latest_version = parsed_version_tag.version
+                out_metadata.latest_version_tag = parsed_version_tag.tag
+                out_metadata.latest_version_position = index
                 break  # there's no need for more info
-        if meta.unreleased_start is not None and meta.unreleased_end is None:
-            meta.unreleased_end = index
```
bearomorphism, line 71: This `index` is unbounded if the file is empty, although this never happens for all code paths

commitizen/changelog_formats/base.py
Lee-W: If we want to make this assumption, we'll need to make this a static method instead, which makes sense but could be seen as a breaking change.

commitizen/changelog_formats/base.py
```diff
@@ -40,34 +40,40 @@ def get_metadata(self, filepath: str) -> Metadata:
...
+    def get_metadata_from_file(self, file: IO[str]) -> Metadata:
+        out_metadata = Metadata()
         unreleased_level: int | None = None
-        for index, line in enumerate(file):
-            line = line.strip().lower()
 
-            unreleased: int | None = None
-            if "unreleased" in line:
-                unreleased = self.parse_title_level(line)
+        lines = [line.strip().lower() for line in file.readlines()]
```
Lee-W, line 47: loop through it twice, probably not worth it?

commitizen/changelog_formats/base.py
Lee-W: but making that a static method is a good idea

PR #1598, opened 2025-09-13
test(SemVer): refine tests and remove unnecessary tests which test th…

bearomorphism:
…e behavior of library

## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```


tests/test_version_scheme_semver2.py
```diff
@@ -136,76 +121,3 @@ def test_semver_scheme_property():
 
 def test_semver_implement_version_protocol():
     assert isinstance(SemVer2("0.0.1"), VersionProtocol)
-
-
-def test_semver_sortable():
```
bearomorphism, line 141: I'm not sure why we need to test this. The `__lt__` operator is defined in `packaging` library and we are not overriding it.

tests/utils.py
```diff
@@ -21,6 +23,15 @@
 )
 
 
+class VersionSchemeTestArgs(NamedTuple):
+    current_version: str
+    increment: Increment | None
+    prerelease: Prerelease | None
+    prerelease_offset: int
+    devrelease: int | None
+    expected_version: str
```
bearomorphism: Main change

tests/test_version_scheme_semver2.py
```diff
@@ -1,131 +1,539 @@
...
-]
+from tests.utils import VersionSchemeTestArgs
 
 
 @pytest.mark.parametrize(
-    "test_input, expected",
-    itertools.chain(tdd_cases, weird_cases, simple_flow, linear_prerelease_cases),
+    "args",
+    [
+        # test driven development
```
Lee-W:
```suggestion
```

I think we can remove this line. This does not provide much information

tests/test_version_scheme_semver2.py
```diff
@@ -1,131 +1,539 @@
...
+        VersionSchemeTestArgs(
+            current_version="0.1.1",
+            increment="PATCH",
+            prerelease=None,
+            prerelease_offset=0,
+            devrelease=None,
+            expected_version="0.1.2",
+        ),
```
Lee-W: I feel like we should make it something like this

"version_args, expected_version"

```suggestion
        (
            VersionSchemeTestArgs(
                current_version="0.1.1",
                increment="PATCH",
                prerelease=None,
                prerelease_offset=0,
                devrelease=None,,
            ),
        "0.1.2"
```

The expected version probably should be viewed as a different argument

bearomorphism: sounds good

tests/test_version_scheme_semver2.py
bearomorphism: yep

removed

PR #1599, opened 2025-09-13
refactor(cargo_provider): cleanup and get rid of potential type errors

bearomorphism:
## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```


commitizen/providers/cargo_provider.py
```diff
@@ -30,65 +25,71 @@ class CargoProvider(TomlProvider):
...
-        # If there is a root package, change its version (but not the workspace version)
-        try:
-            return document["package"]["version"]  # type: ignore[index,return-value]
-        # Else, bump the workspace version
-        except tomlkit.exceptions.NonExistentKey:
-            ...
-        return document["workspace"]["package"]["version"]  # type: ignore[index,return-value]
+    def get(self, document: TOMLDocument) -> str:
+        out = _try_get_workspace(document)["package"]["version"]
+        assert isinstance(out, str)
```
Lee-W: I'm surprised our linting rule does not block this. I'll create another PR for that. But no, I don't think we should allow assert in production code. If we want to make mypy happy

```python
if TYPE_CHCKING:
    assert ...
```

should be used insetad

bearomorphism: sounds good

bearomorphism:
Now all `assert` are wrapped in `if TYPE_CHECKING:`

PR #1601, opened 2025-09-14
refactor(BaseConfig): update docstring, extract factory method and remove unnecessary variable assignment

bearomorphism:
## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```

### DIFF
--- a/commitizen/config/base_config.py
+++ b/commitizen/config/base_config.py
@@ -17,8 +17,8 @@ if TYPE_CHECKING:
 
 class BaseConfig:
     def __init__(self) -> None:
+        self.is_empty_config = False
         self._settings: Settings = DEFAULT_SETTINGS.copy()
-        self.encoding = self.settings["encoding"]
         self._path: Path | None = None
 
     @property
@@ -30,7 +30,7 @@ class BaseConfig:
         return self._path  # type: ignore[return-value]
 
     @path.setter
-    def path(self, path: str | Path) -> None:
+    def path(self, path: Path) -> None:
         self._path = Path(path)
 
     def set_key(self, key: str, value: Any) -> Self:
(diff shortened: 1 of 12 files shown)


commitizen/config/json_config.py
```diff
@@ -22,7 +22,6 @@
 class JsonConfig(BaseConfig):
     def __init__(self, *, data: bytes | str, path: Path | str) -> None:
```
bearomorphism: I think we can remove the `str` of the parameter `path` in the future? I'm not sure if that will be a breaking change

commitizen/config/base_config.py
```diff
@@ -17,8 +17,8 @@
 
 class BaseConfig:
     def __init__(self) -> None:
+        self.is_empty_config = False
         self._settings: Settings = DEFAULT_SETTINGS.copy()
-        self.encoding = self.settings["encoding"]
```
bearomorphism, line 21: I'm not sure why this member variable exists. It will be stale if `self._settings` is updated somewhere.

commitizen/config/base_config.py
```diff
@@ -17,8 +17,8 @@
 
 class BaseConfig:
```
bearomorphism, line 18: Should this be an abstract class?

woile: Ideally, a `Protocol`, but it doesn't hurt like this

commitizen/config/json_config.py
woile: I think it's fine to remove the `str`, at most, it would generate a `mypy` error

commitizen/config/base_config.py
bearomorphism: Thanks. I found it odd that we can create `BaseConfig` objects and it's sometimes confusing. I attempted to refactor it on another branch (not published yet) but many tests would be affected.

commitizen/config/json_config.py
bearomorphism: removed `| str`

commitizen/config/base_config.py
Lee-W: Let's keep it as it is now and improve later

Lee-W (approved):

PR #1605, opened 2025-09-15
refactor(Init): make project_info a module and remove self.project_info

bearomorphism:
## Description

Make `project_info` a module as `ProjectInfo` only contains stateless helpers.


bearomorphism:
Thanks for the quick review!

commitizen/commands/init.py
```diff
@@ -195,14 +141,10 @@ def __call__(self) -> None:
...
 
     def _ask_config_path(self) -> str:
-        default_path = (
-            "pyproject.toml" if self.project_info.has_pyproject else ".cz.toml"
-        )
-
         name: str = questionary.select(
             "Please choose a supported config file: ",
             choices=CONFIG_FILES,
-            default=default_path,
```
bearomorphism, line 218: The suffix `path` can sometimes refer to `Path` objects. I consider it less confusing with the suffix `filename` to indicate that it is a string.

tests/test_project_info.py
```diff
@@ -0,0 +1,287 @@
...
+
+import pytest
+
+from commitizen import project_info
+
+
+class TestFileDetection:
+    """Test file detection functions."""
+
+    def test_has_pyproject_when_file_exists(self, chdir):
```
Lee-W: i think we can use parameterize for these

bearomorphism: Could you provide some details? I'm not sure how to make it more maintainable with parameterize.
Thanks 🙏

tests/test_project_info.py
```diff
@@ -0,0 +1,287 @@
...
+        assert project_info.has_pyproject() is True
+
+    def test_has_pyproject_when_file_not_exists(self, chdir):
+        """Test has_pyproject returns False when pyproject.toml doesn't exist."""
+        assert project_info.has_pyproject() is False
+
+    def test_has_uv_lock_when_file_exists(self, chdir):
+        """Test has_uv_lock returns True when uv.lock exists."""
+        Path("uv.lock").touch()
+        assert project_info.has_uv_lock() is True
```
Lee-W:
```suggestion
    @pytest.mark.parameterzie("file_name, func_name", [("uv.lock", "has_uv_lock"), ...])
    def test_return_true_when_file_exists(self, chdir, file_name: str, func_name: str):
        """Test has_uv_lock returns True when uv.lock exists."""
        Path(file_name).touch()
        assert getattr(project_info, "has_uv_lock")() is True
```

probably can try something like it

bearomorphism: Maybe it would be better if we just remove these simple `has_XXX` functions.

Lee-W: I think they're quite straightforward? Why do you think we should remove them

bearomorphism: `Path(...).is_file()` is also straightforward.

bearomorphism: I will update a commit later. I am still testing locally.

bearomorphism: Now both `project_info.py` and its test file are shorter.

PR #1612, opened 2025-09-16
refactor(BaseFormat): merge ChangelogFormat into BaseFormat

bearomorphism:
## Description

`BaseFormat` already looks like a base class for changelog format, but it has a parent class protocol `ChangelogFormat`. It is confusing to me.

Additionally, the protocol `ChangelogFormat` declares implementations of methods. I consider it confusing too.

This change addresses the unnecessary complexity and remove `ChangelogFormat` completely from the codebase. The alternative now is `BaseFormat`.

I am not sure if this is a breaking change. Please help to review.

### DIFF
--- a/commitizen/changelog_formats/restructuredtext.py
+++ b/commitizen/changelog_formats/restructuredtext.py
@@ -5,6 +5,7 @@ from itertools import zip_longest
 from typing import IO, TYPE_CHECKING, Any, Union
 
 from commitizen.changelog import Metadata
+from commitizen.tags import VersionTag
 
 from .base import BaseFormat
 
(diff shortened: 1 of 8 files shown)


commitizen/changelog_formats/restructuredtext.py
```diff
@@ -90,3 +91,9 @@ def is_underlined_title(self, first: str, second: str) -> bool:
             and not second.isalnum()
             and all(char == second[0] for char in second[1:])
         )
+
+    def parse_version_from_title(self, line: str) -> VersionTag | None:
+        raise NotImplementedError("parse_version_from_title is not implemented")
+
+    def parse_title_level(self, line: str) -> int | None:
+        raise NotImplementedError("parse_title_level is not implemented")
```
bearomorphism, line 99: Not sure if we're going to implement these.

bearomorphism: I think not. Those methods look like protected and is only used in `get_metadata_from_file` but this class overrides the function `get_metadata_from_file`

Lee-W:
I think @noirbizarre might be a better person to take a look at this one. My guts feeling tells me `ChangelogFormat` is mostly for typing

PR #1619, opened 2025-09-20
test: rename the fixture config to mock_config for better code search

bearomorphism:
## Description

There are too many things called `config` in this repo (we have module, parameter key, etc).

Added a prefix `mock_` so we can distinguish them more easily.

### DIFF
--- a/tests/commands/conftest.py
+++ b/tests/commands/conftest.py
@@ -7,7 +7,7 @@ from commitizen.config import BaseConfig, JsonConfig
 
 
 @pytest.fixture()
-def config():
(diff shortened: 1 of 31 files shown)


tests/commands/conftest.py
```diff
@@ -7,7 +7,7 @@
 
 
 @pytest.fixture()
-def config():
+def mock_config():
```
Lee-W, line 10: We're not mocking a config here. We're providing a sample config instead. `mock_config` is inaccurate. I probably would try `sample_config` or something similiar

PR #1621, opened 2025-09-20
test(utils): do not create file for simplicity

bearomorphism:


Lee-W:
I'm not sure I understand the benefit we gain from this. Could you please share more insights? Thanks!

bearomorphism:
probably no. I remember I was thinking not creating file makes the tests faster.

PR #1626, opened 2025-09-21
refactor(ConventionalCommitsCz): rewrite message method to make the p…

bearomorphism:
…attern more clear

## Description

I believe this makes the code read more clearly

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -155,15 +155,16 @@ class ConventionalCommitsCz(BaseCommitizen):
         is_breaking_change = answers["is_breaking_change"]
 
         if scope:
-            scope = f"({scope})"
-        if body:
-            body = f"\n\n{body}"
+            # example: "fix(users): email pattern corrected"
+            first_line = f"{prefix}({scope}): {subject}"
+        else:
+            # example: "fix: email pattern corrected"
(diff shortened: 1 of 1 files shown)


commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -155,15 +155,16 @@ def message(self, answers: ConventionalCommitsAnswers) -> str:  # type: ignore[o
...
+            # example: "fix: email pattern corrected"
+            first_line = f"{prefix}: {subject}"
+
         if is_breaking_change:
             footer = f"BREAKING CHANGE: {footer}"
-        if footer:
-            footer = f"\n\n{footer}"
 
-        return f"{prefix}{scope}: {subject}{body}{footer}"
+        return "\n\n".join(s for s in (first_line, body, footer) if s)
```
Lee-W, line 167: I guess `filter` might work better in this case?

bearomorphism: I don't have a preference here. Both look good to me

bearomorphism: It seems that we are not using `filter` in this project?

I remembered that there used to be some, but they are replaced by list comprehensions.

Lee-W: ok, let's keep it as it is then

PR #1627, opened 2025-09-21
docs: add url to contact links

bearomorphism:
## Description

1. **(Will do in a separate PR)** Sometimes I still need a blank issue form... for example, the idea might be a performance improvement or test coverage improvement, which is neither documentation change, feature request or bug
2. There should be a url field for each entry in `contact_links`. I added @woile 's github profile link.

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -3,4 +3,5 @@
 blank_issues_enabled: false
 contact_links:
   - name: Security Contact
+    url: https://github.com/woile
     about: Please report security vulnerabilities to santiwilly@gmail.com


Lee-W:
or maybe let's add those types?

bearomorphism:
> or maybe let's add those types?

sounds good, will do later

bearomorphism:
Let me add those type of forms in a separate PR. We can fix the contact link first.

bearomorphism:
Pipeline failed... that's strange

Lee-W:
not sure. just triggered the run

Lee-W (approved):

bearomorphism:
image

Just a note that this change does work. We didn't have the 4th option before this PR.

PR #1632, opened 2025-09-28
refactor(Commit): refactor _prompt_commit_questions and fix some type…

bearomorphism:
… hint

## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```


commitizen/commands/commit.py
```diff
@@ -64,31 +64,35 @@ def _read_backup_message(self) -> str | None:
...
+    def _validate_subject_length(self, message: str) -> None:
+        # By the contract, message_length_limit is set to 0 for no limit
+        subject = message.partition("\n")[0].strip()
+        limit = self.arguments.get("message_length_limit", 0)
+        if limit == 0:
+            return
+        if len(subject) > limit:
             raise CommitMessageLengthExceededError(
-                f"Length of commit message exceeds limit ({message_len}/{message_length_limit})"
+                f"Length of commit message exceeds limit ({len(subject)}/{limit}), subject: '{subject}'"
```
bearomorphism: I believe appending more details is not a breaking change?

commitizen/commands/commit.py
```diff
@@ -64,31 +64,35 @@ def _read_backup_message(self) -> str | None:
...
-        message = cz.message(answers)
-        message_len = len(message.partition("\n")[0].strip())
-        message_length_limit = self.arguments.get("message_length_limit", 0)
-        if 0 < message_length_limit < message_len:
+        message = self.cz.message(answers)
+        self._validate_subject_length(message)
+        return message
+
+    def _validate_subject_length(self, message: str) -> None:
+        # By the contract, message_length_limit is set to 0 for no limit
```
bearomorphism, line 88: Not sure if it still works if we use `None` for no limit

bearomorphism: Related to a comment in #1574

commitizen/commands/commit.py
Lee-W: Yep, it should be fine

commitizen/commands/commit.py
Lee-W: If we use None, it's most likely to break, but should not be too hard to handle that

commitizen/commands/commit.py
```diff
@@ -62,33 +62,37 @@ def _read_backup_message(self) -> str | None:
...
+        subject = message.partition("\n")[0].strip()
+        limit = self.arguments.get("message_length_limit", 0)
+        if limit == 0:
+            return
```
Lee-W: This can be checked before running the `_validate_subject_length`. If there's nothing to validate, then we don't do validation

commitizen/commands/commit.py
Lee-W: will take a look at that PR later. But I'll get this merged

commitizen/commands/commit.py
Lee-W: Easy fix. I'll do it and push it. Thanks!

Lee-W (approved):

Issue #1640, opened 2025-11-05
Not possible to use `--get-next` when changelog is enabled


woile:
### Description

It's not possible to properly use `--get-next` when setting `update_changelog_on_bump = true` (is true).

### Steps to reproduce

1. Create a `.cz.toml` with 
```sh
version = "0.0.1"
version_scheme = "semver2"
update_changelog_on_bump = true
```
2. Create a commit and bump normally `cz bump --yes`
3. Add a new commit:
```sh
touch foo.txt
git add foo.txt
git commit -m "fix: foobar"
```
4. Try to get next version `cz bump --get-next`
5. Observe error

```sh
--changelog or --changelog-to-stdout is not allowed with --get-next
```

This would be fine, if there was a way to say `--changelog=false` or `--no-changelog`. But there's not. So this requires the user to "temporarly" set the flag in the `.cz.toml` to false, which in a CI it's painful. 

### Current behavior

There's no way to get the next version if the user has enabled `update_changelog_on_bump` in their settings

### Desired behavior

I see 2 options, to discuss:

1. Do not block on this setting when it's enabled
2. Add a new flag `--no-changelog` that forcefully marks changelog=false

What do you think @Lee-W @noirbizarre ?

### Environment

Commitizen Version: 3.31.0
Python Version: 3.12.11 (main, Jun  3 2025, 15:41:47) [GCC 14.3.0]
Operating System: Linux

woile:
I think this should be just a warning instead of an exception. @Lee-W @bearomorphism ?

bearomorphism:
This week is busy week. I'll take a look when I get bandwidth.

bearomorphism:
I took a quick glance and found this PR #1195 that initially added `--get-next` option.

Maybe need to investigate further why we treat such combination of parameter as `NotAllowed` exception.

https://github.com/commitizen-tools/commitizen/pull/1195/files#diff-920733205b591ccb6aded23a60aba1c20600036d38b5c224e27da47cf09ffb60

bearomorphism:
I think it should just be a warning. I am working on a patch PR now.

bearomorphism:
The patch PR #1644 is ready. Please take a look

bearomorphism:
Please see #1645

Lee-W:
agree a warning would be better 👍

bearomorphism:
We can also consider redesigning the option. For example, make `--get-next` to be a value of `--dry-run`

```
[...]

PR #1644, opened 2025-11-10
fix(bump): remove `NotAllowed` related to `--get-next` option, and other related refactoring

bearomorphism:
# Changes

1. Remove unreachable logic, such as setting `dry_run = True` while we already have `get_next = True`. The latter condition raises a `GetNextExit()` exception in the original code, so no effect if we set `dry_run` to `True`. Moreover, the original code is `self.dry_run = True` which is ineffective since we never use the member variable.
2. Simplify things.
3. Rename the variable `get_next` to `next_version_to_stdout`. The variable name `get_next` is very unclear and hard to guess what it will do. I think we can consider renaming the option in the next major release.
4. (relate #1640) Replace `NotAllowed` related to `--get-next` option with warnings.
5. Remove `GetNextExit` exception because we already have `DryRunExit`. `--get-next` is just `--dry-run` but doing less things IMO.

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -16,7 +16,6 @@ from commitizen.exceptions import (
     BumpTagFailedError,
     DryRunExit,
     ExpectedExit,
-    GetNextExit,
     InvalidManualVersion,
     NoCommitsFoundError,
     NoneIncrementExit,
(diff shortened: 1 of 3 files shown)


Lee-W:
ah... didn't notice this one. I thought I moved all the PRs out of `v4-10-0`

bearomorphism:
It's ok this was a very new PR

PR #1645, opened 2025-11-10
fix(bump): remove `NotAllowed` related to `--get-next` option, and other related refactoring

bearomorphism:
# Changes

1. Remove unreachable logic, such as setting `dry_run = True` while we already have `get_next = True`. The latter condition raises a `GetNextExit()` exception in the original code, so no effect if we set `dry_run` to `True`. Moreover, the original code is `self.dry_run = True` which is ineffective since we never use the member variable.
2. Simplify things.
3. Rename the variable `get_next` to `next_version_to_stdout`. The variable name `get_next` is very unclear and hard to guess what it will do. I think we can consider renaming the option in the next major release.
4. (relate #1640) Replace `NotAllowed` related to `--get-next` option with warnings.
5. Remove `GetNextExit` exception because we already have `DryRunExit`. `--get-next` is just `--dry-run` but doing less things IMO.

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -16,7 +16,6 @@ from commitizen.exceptions import (
     BumpTagFailedError,
     DryRunExit,
     ExpectedExit,
-    GetNextExit,
     InvalidManualVersion,
     NoCommitsFoundError,
     NoneIncrementExit,
@@ -47,7 +46,7 @@ class BumpArgs(Settings, total=False):
     dry_run: bool
     file_name: str
     files_only: bool | None
-    get_next: bool
+    get_next: bool  # TODO: maybe rename to `next_version_to_stdout`
     git_output_to_stderr: bool
     increment_mode: str
     increment: Increment | None
@@ -94,7 +93,6 @@ class Bump:
         )
         self.cz = factory.committer_factory(self.config)
         self.changelog_flag = arguments["changelog"]
-        self.changelog_config = self.config.settings.get("update_changelog_on_bump")
         self.changelog_to_stdout = arguments["changelog_to_stdout"]
         self.git_output_to_stderr = arguments["git_output_to_stderr"]
         self.no_verify = arguments["no_verify"]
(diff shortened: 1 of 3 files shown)


bearomorphism:
image

Is the branck `v4-10-1` broken?

Lee-W:
I just update the `v4-10-1` branch. Could you please rebase from it again? We might need to come up with another release strategy, but this is what we have for the time being

bearomorphism:
@Lee-W @noirbizarre please review

bearomorphism:
> We might need to come up with another release strategy

We may prioritize this. Clarifying which commit new branches should be based on would make it much easier for new contributors, and it should also help reduce merge conflicts and PRs being accidentally closed due to test branch deletion.

commitizen/commands/bump.py
```diff
@@ -270,20 +268,18 @@ def __call__(self) -> None:
...
-            current_version, new_version, self.bump_settings["bump_message"]
-        )
-
-        if get_next:
+        if next_version_to_stdout:
             if increment is None and new_tag_version == current_tag_version:
                 raise NoneIncrementExit(
                     "[NO_COMMITS_TO_BUMP]\n"
                     "The commits found are not eligible to be bumped"
                 )
```
Lee-W, line 276:
```suggestion
        if (
            next_version_to_stdout
            and increment is None
            and new_tag_version == current_tag_version
        ):
            raise NoneIncrementExit(
                "[NO_COMMITS_TO_BUMP]\n"
                "The commits found are not eligible to be bumped"
            )
```

Lee-W (approved):
One nit. Close to merge

Lee-W:
> > We might need to come up with another release strategy
> 
> We may prioritize this. Clarifying which commit new branches should be based on would make it much easier for new contributors, and it should also help reduce merge conflicts and PRs being accidentally closed due to test branch deletion.

Generally `feat` goes to next minor release and others goes to next patch release. docs, tests can go the master directly

commitizen/commands/bump.py
bearomorphism: This looks not equivalent to the original logic

Lee-W: Let me check again later👀

bearomorphism: There are still things that need to be done after exiting `if increment is None and new_tag_version == current_tag_version:` block

Lee-W: yep, missed that

Issue #1643, opened 2025-11-10
[refactor] separate option validation in `bump`


bearomorphism:
Probably not required in this PR, but I feel we could move these validations out as a separate function 🤔

_Originally posted by @Lee-W in https://github.com/commitizen-tools/commitizen/pull/1195#discussion_r1709774497_

bearomorphism:
Created PR #1646

bearomorphism:
Closing via #1646.

PR #1646, opened 2025-11-11
fix(bump): extract option validation and new version resolution

bearomorphism:
# Changes

Relate #1643 

- Refactors
- Collect all `NotAllowed` error messages and raise it at the end of option validation process

### DIFF
--- a/commitizen/commands/bump.py
+++ b/commitizen/commands/bump.py
@@ -29,6 +29,7 @@ from commitizen.version_schemes import (
     Increment,
     InvalidVersion,
     Prerelease,
+    VersionProtocol,
     get_version_scheme,
 )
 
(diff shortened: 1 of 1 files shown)


commitizen/commands/bump.py
```diff
@@ -158,41 +159,91 @@ def _find_increment(self, commits: list[git.GitCommit]) -> Increment | None:
...
+                (self.arguments["increment"], "--increment"),
+                (self.arguments["prerelease"], "--prerelease"),
+                (self.arguments["devrelease"] is not None, "--devrelease"),
+                (self.arguments["local_version"], "--local-version"),
+                (self.arguments["build_metadata"], "--build-metadata"),
+                (self.arguments["major_version_zero"], "--major-version-zero"),
             ):
                 if val:
-                    raise NotAllowed(f"{option} cannot be combined with MANUAL_VERSION")
+                    errors.append(f"{option} cannot be combined with MANUAL_VERSION")
```
Lee-W, line 172: Should we make it `option1, option2 ... cannot be combined with MANUAL_VERSION` instead

commitizen/commands/bump.py
```diff
@@ -158,41 +159,91 @@ def _find_increment(self, commits: list[git.GitCommit]) -> Increment | None:
...
 
-        if major_version_zero and current_version.release[0]:
-            raise NotAllowed(
+        if self.arguments["major_version_zero"] and current_version.release[0]:
+            errors.append(
                 f"--major-version-zero is meaningless for current version {current_version}"
             )
+        if self.arguments["build_metadata"] and self.arguments["local_version"]:
+            errors.append("--local-version cannot be combined with --build-metadata")
```
Lee-W, line 179: The format here is not the same as the manual version one. The manual version one kinda confusing 🤔 `cannot be combined with --manual-version` might be better. How is it used 👀

bearomorphism: `MANUAL_VERSION` is a positional argument. I remember there is no `--manual-version`

commitizen/commands/bump.py
bearomorphism: I don't have a preference here. If the user can easily know what is wrong through the error message, then it is a good error message.

commitizen/commands/bump.py
bearomorphism: the output of `cz bump --help`:

```
...
positional arguments:
  MANUAL_VERSION        bump to the given version (e.g: 1.5.3)
...
```

commitizen/commands/bump.py
Lee-W: ok then, let's keep it as it is still someone complains

commitizen/commands/bump.py
Lee-W: got you. Thanks for confirming!

commitizen/commands/bump.py
```diff
@@ -158,41 +159,91 @@ def _find_increment(self, commits: list[git.GitCommit]) -> Increment | None:
...
+        ):
+            raise NoCommitsFoundError(
+                "[NO_COMMITS_FOUND]\n"
+                "No commits found to generate a pre-release.\n"
+                "To avoid this error, manually specify the type of increment with `--increment`"
+            )
 
-        if build_metadata and is_local_version:
-            raise NotAllowed("--local-version cannot be combined with --build-metadata")
+        # we create an empty PATCH increment for empty tag
```
Lee-W, line 224: Not sure whehter we documented this. If not, we should add it 🤔 but we could do that in the future

tests/commands/test_bump_command.py
```diff
@@ -1508,17 +1508,6 @@ def test_bump_get_next__changelog_to_stdout_is_not_allowed(mocker: MockFixture):
         cli.main()
 
 
-@pytest.mark.usefixtures("tmp_commitizen_project")
-def test_bump_get_next__manual_version_is_not_allowed(mocker: MockFixture):
```
Lee-W: Why is this removed?

Lee-W (commented):
mostly good. one minor question

tests/commands/test_bump_command.py
bearomorphism: There are some conflicts between this and #1645

Maybe we wait for #1645 get merged

Lee-W: one nit on that one. and than I think we're good to merge both

commitizen/commands/bump.py
bearomorphism: Found this PR: https://github.com/commitizen-tools/commitizen/pull/723

bearomorphism: Not sure what "empty tag" means here

I think it should be "no increment"

bearomorphism: #1656

PR #1654, opened 2025-11-19
test: replace try with pytest.raises

bearomorphism:


tests/test_git.py
```diff
@@ -111,11 +112,8 @@ def test_git_message_with_empty_body():
 @pytest.mark.usefixtures("tmp_commitizen_project")
 def test_get_log_as_str_list_empty():
     """ensure an exception or empty list in an empty project"""
```
bearomorphism: Not sure why there are more than one possible behavior in the original test code. That's strange.

Lee-W: Took a look at the history. Looks like it varies based on the git version.

Lee-W: We probably need to take a look at which git version changed this

bearomorphism: Sounds good

Lee-W: We can either revert this test or add a skipif based on git version. We can merge this one after one of these two actions has been done. Thanks!

bearomorphism: will revert later

bearomorphism:
ok now it is ready to be merged.
@Lee-W could you help to take a look

Lee-W (approved):

PR #1664, opened 2025-11-20
docs(third-party): separate providers and template docs

bearomorphism:
### Documentation Changes


Lee-W (approved):

woile:
Can we call them differently? I don't think they are "templates", and templates already fill another roll. 
What about "conventions", "rules", or something else?

bearomorphism:
I'd prefer calling them "rules" or "plugins" to distinguish them from changelog templates.

These third party projects contain data about commit validation rules, commit generation process, bump rules, etc. I agree the name "template" is a bit misleading.

We can add a historical note to point out that such things are called "templates" before this PR.

bearomorphism:
image

https://github.com/commitizen-tools/commitizen/blob/d037977a3939e50b95e88dce9d8fc8a34aa6e0fb/docs/config.md?plain=1#L10

In this page, it is also called "rules"

woile:
If we want to be specific, I think the best would be to start calling them: "conventions".

Teams, people or organizations develop "commit convetions", I don't think "rules", nor "templates" are adequate.
At least, that's what makes sense to me. I would search for "conventions", not "templates".
And yes, they are installed as "plugins" [[0]](https://github.com/commitizen-tools/commitizen/blob/master/pyproject.toml#L61)

Keep in mind, that this split, doesn't take into account the fact that you can publish a "convention", a "provider", a "changelog_format" and a "scheme" all under the same package. `commitizen` itself is an example of this [[1]](https://github.com/commitizen-tools/commitizen/blob/master/pyproject.toml#L61-L85)

Maybe it would make more sense to have:
"Third party plugins"
And for each plugging, we document what do they provide:
- a convention
- a scheme
- a provider
- a changelog_format

or all of them

bearomorphism:
Sounds good.

bearomorphism:
> And for each plugging, we document what do they provide:

For this part, I added some TODOs and we can have future contributors to address the missing content issue. (I would not like to look into what features the plugin provides for each plugin. If I do so, there will be likely inaccurate information which is worse than missing content, and it is also hard for reviewers to spot the misinformation in this PR)

Now this PR is ready for review.

docs/third-party-plugins/about.md
```diff
@@ -0,0 +1,74 @@
...
+
+Read and write version from data sources.
+
+### Changelog format
+
+Generate changelog in customized formats.
+
+## How do I create a new plugin?
+
+<!-- TODO -->
```
Lee-W: Let's add a WIP warning or something like that. Otherwise, someone read it would be confused

docs/third-party-plugins/about.md
```diff
@@ -0,0 +1,74 @@
...
+
+    ## Usage
+
+    ```sh
+    cz --name package-name commit
+    ```
+
+    ## Example
+
+    <!-- Example usage of the plugin. -->
```
Lee-W, line 73: Let's dedent it and use ```markdown ``` code block

mkdocs.yml
```diff
@@ -56,7 +56,17 @@ nav:
...
+      - About: "third-party-plugins/about.md"
+      - "third-party-plugins/github-jira-conventional.md"
+      - "third-party-plugins/conventional-jira.md"
+      - "third-party-plugins/cz-emoji.md"
+      - "third-party-plugins/cz-conventional-gitmoji.md"
+      - "third-party-plugins/cz-legacy.md"
+      - "third-party-plugins/cz-path.md"
+      - "third-party-plugins/commitizen-deno-provider.md"
+      - "third-party-plugins/cz-ai.md"
+      - "third-party-plugins/commitizen-emoji.md"
```
Lee-W: let's sort by alphabet?

docs/third-party-plugins/about.md
bearomorphism: I guess that does not work (?

image

mkdocs.yml
bearomorphism: done

docs/third-party-plugins/about.md
bearomorphism: Commented the whole section and added some simple actions about the TODO

docs/third-party-plugins/about.md
Lee-W: hmmm... ok.. it's not as smart as it should be

Lee-W (approved):

PR #1670, opened 2025-11-22
refactor(cli): separate parser data and main function logic

bearomorphism:
The main function is getting a bit large. I believe it is better to separate the core main function logic with cli configuration data for maintainability.

## Checklist

### Code Changes


woile:
I'm not sure I like splitting the cli into 2 files. I expect to find the args of a cli in the `cli.py`

Lee-W:
I don't have a strong opinion on this one 🤔 either way works for me. so I'll let you both discuss whether we want to make change like this one

bearomorphism:
> I'm not sure I like splitting the cli into 2 files. I expect to find the args of a cli in the `cli.py`

I don't expect a main function to be so large, but I also see your point. Maybe we group cli related thing into a file, and other things, such as excepthook and configuration file related things, into another file?

bearomorphism:
For example

- `cli.py`: CLI related logic and data, will export to `main.py`
- `main.py`: entry point of `cz`, with configuration file resolution

woile:
That'd be fine by me

PR #1673, opened 2025-11-22
fix(version): fix the behavior of `cz version --major`

bearomorphism:
## Description

Fixed the behavior of `--major` and `--minor`. Added tests.

### Before this PR

```bash
cz version --major
# 4.10.0
```

### After this PR

```bash
cz version --major
# Output errors about wrong usage of `cz version`
```

Also added the expected usage for `--major` and `--minor` options.

## Checklist

### Code Changes

### DIFF
--- a/commitizen/commands/version.py
+++ b/commitizen/commands/version.py
@@ -29,20 +29,20 @@ class Version:
 
     def __init__(self, config: BaseConfig, arguments: VersionArgs) -> None:
         self.config: BaseConfig = config
-        self.parameter = arguments
+        self.arguments = arguments
 
     def __call__(self) -> None:
-        if self.parameter.get("report"):
+        if self.arguments.get("report"):
             out.write(f"Commitizen Version: {__version__}")
             out.write(f"Python Version: {sys.version}")
             out.write(f"Operating System: {platform.system()}")
             return
 
-        if self.parameter.get("verbose"):
+        if self.arguments.get("verbose"):
             out.write(f"Installed Commitizen Version: {__version__}")
 
-        if not self.parameter.get("commitizen") and (
-            self.parameter.get("project") or self.parameter.get("verbose")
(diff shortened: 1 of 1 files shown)


woile:
Shouldn't this scenario fail? `cz version` gives the version of `cz` itself not the project.
I was not sure what to do myself.

But the way I was picturing it was:

> how to work with the project's version?

which translates to `cz version --project` or `cz version -p`.

I think users would want to work withe the project's version, if you want your project's next version: `cz version -p --next`, or your project's current version: `cz version -p --major`. 

By allowing `cz version --major` I think we may be creating confusion for the users. It becomes easier to shoot yourself in the foot, don't you think? I don't see a use case for `cz version --major` really. 
I'm inclined for failing, until we have a concrete reason for `cz version --major` 

I agree that right now it's a bogus scenario:
`cz version --major` returns a wrong value

bearomorphism:
Sounds good, then I will make it fail in the next iteration.
I created this patch because I was playing with cz version and I found the behavior of cz version --major strange, at least I thought it should not output the full version of commitizen itself.

Failing the command sounds better

bearomorphism:
How about showing the help of `cz version` if there are unexpected combination of parameters? I am not sure if there is a clean way to do it.

bearomorphism:
I moved the no argument `cz version` logic back to the end of `__call__`... reordering those was not a smart decision, that makes us hard to augment the command

Lee-W:
I just rebase the `v4-10-1` with the latest changes. Could you please rebase from `v4-10-1`? Thanks!

bearomorphism:
Also updated PR description.

Lee-W:
Once this is rebased, reviewed and merged, I'm thinking of releasing 4.10.1 and defer the rest of things to 4.11.0. WDYT?

bearomorphism:
Sounds good

PR #1682, opened 2025-12-03
perf: add TYPE_CHECKING to CzQuestion imports

### LINKED ISSUE #1515: Add `TYPE_CHECKING` to `CzQuestion`
we might be able to move it to type_checking. but this can be next pr or let ruff handle it in the future

_Originally posted by @Lee-W in https://github.com/commitizen-tools/commitizen/pull/1485#discussion_r2134506265_

bearomorphism:
Closes #1515

## Checklist

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -1,10 +1,14 @@
+from __future__ import annotations
+
 import os
-from typing import TypedDict
+from typing import TYPE_CHECKING, TypedDict
 
 from commitizen import defaults
 from commitizen.cz.base import BaseCommitizen
 from commitizen.cz.utils import multiple_line_breaker, required_validator
-from commitizen.question import CzQuestion
+
+if TYPE_CHECKING:
+    from commitizen.question import CzQuestion
 
 __all__ = ["ConventionalCommitsCz"]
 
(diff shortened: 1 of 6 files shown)


commitizen/cz/conventional_commits/conventional_commits.py
```diff
@@ -1,10 +1,14 @@
...
 import os
-from typing import TypedDict
+from typing import TYPE_CHECKING, TypedDict
 
 from commitizen import defaults
 from commitizen.cz.base import BaseCommitizen
 from commitizen.cz.utils import multiple_line_breaker, required_validator
-from commitizen.question import CzQuestion
+
+if TYPE_CHECKING:
```
Lee-W, line 10: There's a ruff rule doing something like this. Probably worth enabling it and let ruff do it for us in the future

Lee-W (approved):

commitizen/cz/conventional_commits/conventional_commits.py
bearomorphism: Do you mean TC004?

https://docs.astral.sh/ruff/rules/runtime-import-in-type-checking-block/

bearomorphism: I remember there is no ruff rule that can automate this like what TypeScript's `import type ...` do

bearomorphism: Ok I see TC003

https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/

ChatGPT is liar

PR #1684, opened 2025-12-03
docs(contributing): fix broken link

bearomorphism:


woile:
Maybe we should automate check for broken links (and run with a cron) using https://github.com/lycheeverse/lychee

bearomorphism:
> Maybe we should automate check for broken links (and run with a cron) using https://github.com/lycheeverse/lychee

sounds good if it's not expensive

Issue #1704, opened 2025-12-10
Monotonic Versioning Schema


YazdanRa:
### Description

Add a new single number (monotonic) versioning schema.

For example: 

v1, v2, v3, ...

### Possible Solution

Adding a new schema. It won't affect the rest of the code.

I have already implemented this feature for our own use. Wanted to share if it will be useful for others.

### Additional context

Some services are using monotonic versioning (e.g. AWS SageMaker Models) for their artifacts, this will enable the users to keep the code and registry in sync.

bearomorphism:
Hi @YazdanRa , thanks for your contribution!

I wonder how monotonic versioning works. It would be great if you could provide some references.

The only reference I could find using the keyword "monotonic versioning" is [monotonic-versioning-manifesto](https://github.com/benkehoe/monotonic-versioning-manifesto), but it does not work as you mentioned in the issue description.

YazdanRa:
Hi @bearomorphism, Thanks for prompt response!

I can see the confusion. I picked the name "monotonic" purely based on what it means and honestly didn't really dig into the specifications.

For what I meant this to be, version only has 1 section, a single number that only can increases. In my [PR](https://github.com/commitizen-tools/commitizen/pull/1705) I treated it as the major section of the version to don't change a lot from the code.

```python
class MonotonicVersion(BaseVersion):
    """
    Monotonic versioning scheme

    Any increment bump simply increases the single numeric component.
    """

    parser: ClassVar[re.Pattern] = re.compile(
        r"v?(?P<version>([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z.]+)?(\w+)?)"
    )

    def increment_base(self, increment: Increment | None = None) -> str:
        return f"{self.major + 1}"

    def _get_increment_base(
        self, increment: Increment | None, exact_increment: bool
    ) -> str:
        return self.increment_base(increment)

```

I probably should change the naming so it's not confusing. Any suggestions on what it should be called instead?

YazdanRa:
Maybe "Sequential Version", "Integer Version", "Number Version", "Single Number Version".

bearomorphism:
I would ask AI to provide naming suggestions. Naming is hard.

"Monotonic" has nothing to do with "single". Probably "Single number version" or "Monic version"? (I haven't checked if the names are already used.)

[...]

PR #1705, opened 2025-12-10
feat(tags): enable version schemes with less than 3 components

YazdanRa:
## Description

Upadted version unpacking in `tags.py` to enable support for version schemes with less than 3 components.

## Checklist

- [X] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/)

### Code Changes

- [X] Run `poetry all` locally to ensure this change passes linter check and tests

### Documentation Changes

## Expected Behavior

If version in pyproject.toml or cli argument is less than 3 components (e.g. `1`, `1.0`) we won't get `ValueError: not enough values to unpack (expected 3, got N)` error.

Addded new unit tests
```shell
poetry all
```

## Additional Context
#1704

### DIFF
--- a/docs/commands/bump.md
+++ b/docs/commands/bump.md
@@ -65,6 +65,19 @@ You can also set this in the configuration file with `version_scheme = "semver"`
     | Devrelease     | `0.1.1.dev1`   | `0.1.1-dev1`    |
     | Dev and pre    | `1.0.0a3.dev1` | `1.0.0-a3-dev1` |
 
+
+!!! note "Incomplete Version Handling"
+    Commitizen treats a three-part version (major.minor.patch) as complete.
+    If your configured version is incomplete (for example, `1` or `1.2`), Commitizen pads missing parts with zeros when it needs `major/minor/patch` for tag formatting.
+    The tag output depends on your `tag_format`: formats using `${version}` keep `1`/`1.2`, while formats using `${major}.${minor}.${patch}` will render `1.0.0`/`1.2.0`.
+
+    When bumping from an incomplete version, Commitizen looks for the latest existing tag that matches the provided release prefix.
+    For example, if the current version is `1.2` and the latest `1.2.x` tag is `1.2.3`, then a patch bump yields `1.2.4` and a minor bump yields `1.3.0`.
+
+!!! tip
+    To control the behaviour of bumping and version parsing, you may implement your own `version_scheme` by inheriting from `commitizen.version_schemes.BaseVersion` or use an existing plugin package.
+
+
 ### PEP440 Version Examples
 
 Commitizen supports the [PEP 440][pep440] version format, which includes several version types. Here are examples of each:


bearomorphism:
As I mentioned in #1704, please publish the new version scheme as a third-party plugin.

You may either reuse this branch or create another branch to update the third-party plugin list. Thanks!

tests/commands/test_bump_command.py
```diff
@@ -966,6 +968,48 @@ def test_bump_manual_version_disallows_major_version_zero(mocker):
...
+
+    tag_exists = git.tag_exist("1.3.0")
+    assert tag_exists is True
+
+    for version_file in [
+        tmp_commitizen_project.join("__version__.py"),
+        tmp_commitizen_project.join("pyproject.toml"),
+    ]:
+        with open(version_file) as f:
+            assert "1.3.0" in f.read()
```
bearomorphism: Use `@pytest.mark.parametrize` to merge these two test cases

YazdanRa: It has been addressed. Thanks.

bearomorphism (commented):
Code changes LGTM, but I am not sure if this is a `feat` change.

YazdanRa:
What should it be?

bearomorphism:
@Lee-W wdyt

YazdanRa:
Please let me know if any action is needed on my end ツ

Lee-W (commented):
This could be a feature, but we'll need to document this behavior somewhere. Overall, looks good to me. Once doc is ready, we can merge it to `v4-11-0` branch

Lee-W:
@woile based on our design, I think 1.2 is essentially 1.2.0. Do you think there would be exception that these 2 are different?

YazdanRa:
Will send another PR to add the plugin documentation after the next release.

Thanks and happy holidays!

woile:
@Lee-W I usually understand 1.2 as the full range of versions: 1.2.0, 1.2.1, 1.2.3... etc. This in something like npm.

In the case that we are not talking about semver, the semantics could change and what we know would no longer apply. It would depend on how the version scheme is designed. For example, python itself doesn't use semver nor pep440. A bump like 3.12 to 3.13 introduces breaking changes, and even though they have patches, what people use is the major.minor to reference a version

noirbizarre:
Agreed with @Lee-W it needs documentation to be merged.

To me it also need more testing as I agree with @woile, to me `1.2` is the latest `1.2.x`.

I fear this can have unexpected effects on repository maintaining Docker-like or github actions like tagging (which is my case on some repository):

- latest version is tagged `x.y.z`
- an action will automatically move the rolling tags `x.y` and `x` to match 

So in the case of a bump 1.2.3 from to 1.2.4, there will be a `1.2` and `1` tag on the same commit. 

This case is relatively common and need to be tested (to avoid errors like duplicate tags).

YazdanRa:
Just following up, please let me know if you need any action from my side ツ

Lee-W:
Agree with @woile and @noirbizarre , I guess that means we're not going to tackle this issue in this way. Probably make `1.2.x` point to the latest version? I feel that would be a good v5 feature

YazdanRa:
> Agree with @woile and @noirbizarre , I guess that means we're not going to tackle this issue in this way. Probably make `1.2.x` point to the latest version? I feel that would be a good v5 feature

Got it, I started the work for it. Although, just want to mention that the change here (normalize_tag) will still be needed.
That line in normalize_tag is about formatting tags, not resolving “latest.” When a user uses $major.$minor.$patch in tag_format, we still need a stable 3‑tuple even if the version is 1 or 1.2. Padding to 1.0.0 / 1.2.0 there is still correct and keeps tag rendering consistent.

For the change you mentioned, I will need to update the following function:

```python
def find_tag_for(
    self, tags: Iterable[GitTag], version: Version | str
) -> GitTag | None:
```

Lee-W:
What would happen without this fix? Specifically, ` we still need a stable 3‑tuple even if the version is 1 or 1.2. Padding to 1.0.0 / 1.2.0 there is still correct and keeps tag rendering consistent.`

YazdanRa:
> What would happen without this fix? Specifically, ` we still need a stable 3‑tuple even if the version is 1 or 1.2. Padding to 1.0.0 / 1.2.0 there is still correct and keeps tag rendering consistent.`

The `normalize_tag` function is about reading the tags and extracting the version from them.
If we have the `v12.3` tag and it tries to extract the version (Major, Minor, Patch) we get `ValueError: not enough values to unpack (expected 3, got 2)`. (similarly for `v5`)

For the rest of the code to work, we need to default Patch (and Minor) to a value. In my change, I defaulted them to 0 since the value of that function is not being used to set any value. I could also default to `None` but since the expected type is not `None` it would have been too many changes across the codebase.

Lee-W:
> > What would happen without this fix? Specifically, ` we still need a stable 3‑tuple even if the version is 1 or 1.2. Padding to 1.0.0 / 1.2.0 there is still correct and keeps tag rendering consistent.`
> 
> The `normalize_tag` function is about reading the tags and extracting the version from them. If we have the `v12.3` tag and it tries to extract the version (Major, Minor, Patch) we get `ValueError: not enough values to unpack (expected 3, got 2)`. (similarly for `v5`)
> 
> For the rest of the code to work, we need to default Patch (and Minor) to a value. In my change, I defaulted them to 0 since the value of that function is not being used to set any value. I could also default to `None` but since the expected type is not `None` it would have been too many changes across the codebase.

I'm good with this change. but maybe let's add a warning here? like this is missed we fallback to 0 but in the future version (e.g., v5), we're going to change it. WDYT?

YazdanRa:
> > > What would happen without this fix? Specifically, ` we still need a stable 3‑tuple even if the version is 1 or 1.2. Padding to 1.0.0 / 1.2.0 there is still correct and keeps tag rendering consistent.`
> > 
> > 
> > The `normalize_tag` function is about reading the tags and extracting the version from them. If we have the `v12.3` tag and it tries to extract the version (Major, Minor, Patch) we get `ValueError: not enough values to unpack (expected 3, got 2)`. (similarly for `v5`)
> > For the rest of the code to work, we need to default Patch (and Minor) to a value. In my change, I defaulted them to 0 since the value of that function is not being used to set any value. I could also default to `None` but since the expected type is not `None` it would have been too many changes across the codebase.
> 
> I'm good with this change. but maybe let's add a warning here? like this is missed we fallback to 0 but in the future version (e.g., v5), we're going to change it. WDYT?

Sure thing, Shall I still keep the other change for resolving the latest version with the given prefix in `find_tag_for` function as well?

Lee-W:
> > > > What would happen without this fix? Specifically, ` we still need a stable 3‑tuple even if the version is 1 or 1.2. Padding to 1.0.0 / 1.2.0 there is still correct and keeps tag rendering consistent.`
> > > 
> > > 
> > > The `normalize_tag` function is about reading the tags and extracting the version from them. If we have the `v12.3` tag and it tries to extract the version (Major, Minor, Patch) we get `ValueError: not enough values to unpack (expected 3, got 2)`. (similarly for `v5`)
> > > For the rest of the code to work, we need to default Patch (and Minor) to a value. In my change, I defaulted them to 0 since the value of that function is not being used to set any value. I could also default to `None` but since the expected type is not `None` it would have been too many changes across the codebase.
> > 
> > 
> > I'm good with this change. but maybe let's add a warning here? like this is missed we fallback to 0 but in the future version (e.g., v5), we're going to change it. WDYT?
> 
> Sure thing, Shall I still keep the other change for resolving the latest version with the given prefix in `find_tag_for` function as well?

I don't get it 🤔 If we already have another function that can check the latest version (e.g., 1.2 -> 1.2.5 ), then we should use that here. If not, I would need you help to elaborate a bit. Thanks!

YazdanRa:
@Lee-W there are two changes in this PR:

- Update the `normalize_tag` function to handle tag versions with less than 3 components by masking with 0.0.0.

This is a fix to `ValueError: not enough values to unpack (expected 3, got 2)` when the git tags and version in `pyproject.toml` file has less than 3 components.

- Update the `find_tag_for` function to return the latest version matching with the given version prefix. 

This is a new feature as we discussed in [here](https://github.com/commitizen-tools/commitizen/pull/1705#issuecomment-3664101259). If we're referencing an uncompleted version (e.g. 1.2), it will retrieve the latest patch for that version. (e.g. 1.2.3)

YazdanRa:
I added some documentation for this behaviour.

image

Please let me know if there is anything needed from me here.

PR #1715, opened 2025-12-10
fix: workspace member with fixed version

vemilano:
## Description
Fix 2 things with cargo provider : 
- use same order to get / set a version for package or workspace
- there is no `workspace.version = "0.1.0"` in spec to fix the version of a specific workspace member. It must be  `version = "0.1.0"`, even if it is a workspace member

## Checklist

### Code Changes

### Documentation Changes

>
> ```text
> ```

### DIFF
--- a/commitizen/providers/cargo_provider.py
+++ b/commitizen/providers/cargo_provider.py
@@ -5,6 +5,7 @@ import glob
 from pathlib import Path
 
 import tomlkit
+from tomlkit.items import Table
 
 from commitizen.providers.base_provider import TomlProvider
 
@@ -41,11 +42,11 @@ class CargoProvider(TomlProvider):
 
     def set(self, document: tomlkit.TOMLDocument, version: str) -> None:
         try:
-            document["workspace"]["package"]["version"] = version  # type: ignore[index]
+            document["package"]["version"] = version  # type: ignore[index]
             return
         except tomlkit.exceptions.NonExistentKey:
             ...
-        document["package"]["version"] = version  # type: ignore[index]
+        document["workspace"]["package"]["version"] = version  # type: ignore[index]
 
     def set_version(self, version: str) -> None:
         super().set_version(version)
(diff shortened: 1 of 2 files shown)


bearomorphism:
Hi @vemilano , 

We are planning to merge the branch v4-10-1, which contains changes in `cargo_provider.py`, into master branch.

You can resolve the potential conflict on this PR after we release 4.10.1.

cc @Lee-W

bearomorphism:
nit: you have typo in your commit message

vemilano:
I reviewed the commit in 4.10.1 and it should fix both issues. However, it changes one behavior: if there are both a workspace and a root package, it will bump the version of the workspace, and not the root package.

This might be unwanted behavior, as if there is a root package, this is probably the main project the author is working on.

We may want to get the version of the root package first (if it exists), and fallback on the version of the workspace.

What do you think of it @bearomorphism ?

bearomorphism:
Actually, I am not familiar with cargo files.

@Lee-W wdyt? Which behavior do you think is correct?

woile:
Hi @vemilano could you rebase? The current `cargo_provider` is quite different as of now. However, the behavior is the same. `workspace` is evaluated first. This has been done for simplicity and lack of "opinions" (we test workspace, and on failure we check project root). I usually either have a root version, or a shared workspace version, so it hasn't been an issue for me. Could you elaborate on why updating the root version first makes sense when they differ?  thanks

For rebasing:

```sh
git remote add upstream https://github.com/commitizen-tools/commitizen.git
git fetch upstream
git rebase upstream/master
# fix conflicts
git add commitizen/providers/cargo_provider.py
git rebase --continue
git push origin fix/cargo-workspace-version -f
```

vemilano:
Hi @woile , apologies for the delay.

I checked 4.10.1and it fixes the issues I wanted to fix with this pull request, so rebasing should not be necessary. However, as you said, this version changes the priority of "version reading": firstly workspace version, then package version. It's ok in most cases, except (maybe) when we have both a package and a workspace in the Cargo.toml
With 4.10.1, in this case, it will bump the workspace version rather than the package version.

If a project have both package and workspace in the Cargo.toml, we might want to increase the version of the package, because it's often the main binary/library the author is working on. In this case, we should try to get package version first, then workspace version.

It's open to debate, I just wanted to mention this case so that we're sure of the final choice.

woile:
Sounds good to me, can you rebase and resolve conflicts?

PR #1718, opened 2025-12-11
build: add missing dark mode back

bearomorphism:
The original config seems to be incorrect. The dark mode toggle button does not show.

Reference: https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/


Lee-W (approved):

Lee-W:
feels more like `build` as this is configuring the setting to build doc

PR #1722, opened 2025-12-11
docs(faq): categorize faq into "how to", "feature request graveyard",…

bearomorphism:
… and "history"

## Why

Now faq page focuses on "how to do XXX", and "feature request graveyard" is about decisions.

Better name suggestions are welcome.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -1,0 +1,23 @@
+## Is this project affiliated with the [cz-cli][cz-cli] project?
+
+**It is not affiliated.**
+
+Both are used for similar purposes, parsing commits, generating changelog and version we presume.
+Our Commitizen project is written in python to make integration easier for python projects, whereas [cz-cli][cz-cli] is written in JavaScript and serves the JS packages.
+
+<!-- TODO: Add more details about the differences between Commitizen and cz-cli -->
+
+They differ a bit in design, not sure if cz-cli does any of this, but these are some things you can do with our Commitizen:
+
+- create custom rules, version bumps and changelog generation. By default, we use the popular conventional commits (I think cz-cli allows this).
+- single package, install one thing and it will work. cz-cli is a monorepo, but you have to install different dependencies as far as I know.
+- pre-commit integration
+- works on any language project, as long as you create the `.cz.toml` or `cz.toml` file.
+
+Where do they cross paths?
+
+If you are using conventional commits in your git history, then you could swap one with the other in theory.
+
+Regarding the name, [cz-cli][cz-cli] came first, they used the word Commitizen first. When this project was created originally, the creator read "be a good commitizen", and thought it was just a cool word that made sense, and this would be a package that helps you be a good "commit citizen".
+
+[cz-cli]: https://github.com/commitizen/cz-cli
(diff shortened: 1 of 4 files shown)


docs/history.md
```diff
@@ -0,0 +1,24 @@
+
```
Lee-W:
```suggestion
```

docs/history.md
```diff
@@ -0,0 +1,24 @@
...
+## Is this project affiliated with the [Commitizen JS][cz-js] project?
```
Lee-W:
```suggestion
## Is this project affiliated with the [cz-cli][cz-js] project?
```

docs/history.md
```diff
@@ -0,0 +1,24 @@
...
+
+**It is not affiliated.**
+
+Both are used for similar purposes, parsing commits, generating changelog and version we presume.
+This one is written in python to make integration easier for python projects and the other serves the JS packages.
+
+<!-- TODO: Add more details about the differences between Commitizen and Commitizen JS -->
+
+They differ a bit in design, not sure if cz-js does any of this, but these are some things you can do with our Commitizen:
```
Lee-W:
```suggestion
They differ a bit in design, not sure if [cz-cli] does any of this, but these are some things you can do with our Commitizen:
```

docs/history.md
```diff
@@ -0,0 +1,24 @@
...
+
+- create custom rules, version bumps and changelog generation. By default, we use the popular conventional commits (I think cz-js allows this).
+- single package, install one thing and it will work. cz-js is a monorepo, but you have to install different dependencies as far as I know.
+- pre-commit integration
+- works on any language project, as long as you create the `.cz.toml` or `cz.toml` file.
```
Lee-W:
```suggestion
- create custom rules, version bumps and changelog generation. By default, we use the popular conventional commits (I think cz-cli allows this).
- single package, install one thing and it will work. cz-cli is a monorepo, but you have to install different dependencies as far as I know.
- pre-commit integration
- works on any language project, as long as you create the `.cz.toml` or `cz.toml` file.
```

docs/history.md
```diff
@@ -0,0 +1,24 @@
...
+
+Where do they cross paths?
+
+If you are using conventional commits in your git history, then you could swap one with the other in theory.
+
+Regarding the name, [cz-js][cz-js] came first, they used the word Commitizen first. When this project was created originally, the creator read "be a good commitizen", and thought it was just a cool word that made sense, and this would be a package that helps you be a good "commit citizen".
```
Lee-W:
```suggestion
Regarding the name, [cz-cli][cz-js] came first, they used the word Commitizen first. When this project was created originally, the creator read "be a good commitizen", and thought it was just a cool word that made sense, and this would be a package that helps you be a good "commit citizen".
```

Lee-W: Let's use their repo name but we could mention it's written in javascript

mkdocs.yml
```diff
@@ -81,6 +81,7 @@ nav:
       - "third-party-plugins/github-jira-conventional.md"
   - Contributing: "contributing.md"
   - Contributing TL;DR: "contributing_tldr.md"
+  - "history.md"
```
Lee-W, line 86: Don't we need a key here?

bearomorphism: image

It will render like this if there is no key.

docs/history.md
bearomorphism: Fixed

docs/history.md
bearomorphism: Maybe we can add a formatter for markdown?

Issue #1678, opened 2025-11-30
Introduce `--next` flag on `cz version --project`


woile:
### Description

returns the next version based on the project's config. 

It could potentially support forcing the increment type. Example: --next=MAJOR or --next=PATCH

bearomorphism:
Cross-checked against `master` (v4.15.1) by an `claude-opus-4.7` validation pass while triaging #1976.

**Status**: partially fixed.

| Form | Status |
| --- | --- |
| `cz version --project --next MAJOR\|MINOR\|PATCH` | implemented (`commitizen/cli.py:577-587`, `commitizen/commands/version.py:93-103`); covered by `tests/commands/test_version_command.py::test_next_version` |
| `cz version --project --next` (no value, derive increment from commit history) | **not implemented** — explicit TODO at `commitizen/commands/version.py:85-91` returns `"--next USE_GIT_COMMITS is not implemented yet."`, locked in by `tests/commands/test_version_command.py:289-292` |

The default `--next` form (no value) is the headline behaviour the issue asks for, so this issue should stay open until the bump-increment derivation in `commitizen/commands/bump.py` is refactored so `version.py` can call it and replace the `USE_GIT_COMMITS` TODO.

(Surfaced via the round-3 triage in #1976.)

Issue #1677, opened 2025-11-30
Add new capabilities to `cz version`


woile:
### Description

The idea of this ticket is to track the "graduation" of `cz version --project` into a powerful command, capable of fetching information about the project.

This includes:
- [ ] Introduce `--next` flag on `cz version --project`. Which returns the next version based on the project's config. It could potentially support forcing the increment type. Example: `--next=MAJOR` or `--next=PATCH` #1678 
- [ ] Allow users to play with the configured `version_scheme`, by letting them provide any version, `cz version <version_arg>`, this would override the commitizen version or the given `--project` parameter. The user would then be able to "learn" about the behavior, by doing `cz version 0.1.0 --next=MAJOR` #1679
- Add `--tag`: to actually retrieve the tag of the version with the template applied. If your tags looks like this: `v0.1.0`. You have to manually add v to your version right now. This would be useful to interact with git. If you want to retrieve the tagged commit, you could do: git show-ref $(cz version -p --tag). This is useful for auditing. And, if you have an old CI (e.g: jenkins), it would give an easy way for teams to check if the tag exists already before trying to bump. This parameter is the one I'm least sure about, so maybe we can work on the other features first, and if we think it would be useful, we add it.
- Do we want a `--hash` to get the version hash for auditing purposes?

#1640

Clockwork-Muse:
> Add --tag

I have a use for this.
Due to some weaknesses in the current published action, I've had to write my own action, and I'm having to do manual tag formatting.

Additionally, I would probably have a use for a `--prev` flag, to get the most recent previous revision.  This is potentially for regenerating incremental changelogs when running on a tag; the tag would be the "current" version, so the version to start from the previous one.

Lee-W:
if we have `--next`, `--prev` is indeed something we can consider IMO 👀

woile:
> Additionally, I would probably have a use for a --prev flag, to get the most recent previous revision. This is potentially for regenerating incremental changelogs when running on a tag; the tag would be the "current" version, so the version to start from the previous one.

I'm not sure I full understand this, what about this:

```sh
current=cz version -p
cz bump --yes
next_version=cz version -p 
cz changelog $next_version --dry-run > incremental-changelog.md
```

And by the way, I've introduced a new, more flexible github action, maybe it's useful to you
https://github.com/commitizen-tools/setup-cz

Clockwork-Muse:
> I'm not sure I full understand this, what about this:

[...]

PR #1724, opened 2025-12-12
feat(version): add MANUAL_VERSION, --next and --patch to version command

### LINKED ISSUE #1679: Add support for `cz version <version_arg>`
### Description

Allow users to play with the configured version_scheme, by letting them provide any version, `cz version <version_arg>`, this would override the commitizen version or the given --project parameter. The user would then be able to "learn" about the behavior, by doing cz version 0.1.0 --next=MAJOR

bearomorphism:
Closes #1679

Manually tested and added test cases, the result LGTM.

### DIFF
--- a/commitizen/commands/version.py
+++ b/commitizen/commands/version.py
@@ -2,22 +2,32 @@ import platform
 import sys
 from typing import TypedDict
 
+from packaging.version import InvalidVersion
+
 from commitizen import out
 from commitizen.__version__ import __version__
 from commitizen.config import BaseConfig
 from commitizen.exceptions import NoVersionSpecifiedError, VersionSchemeUnknown
 from commitizen.providers import get_provider
 from commitizen.tags import TagRules
+from commitizen.version_increment import VersionIncrement
 from commitizen.version_schemes import get_version_scheme
 
 
 class VersionArgs(TypedDict, total=False):
+    manual_version: str | None
+    next: str | None
+
+    # Exclusive groups 1
     commitizen: bool
     report: bool
     project: bool
     verbose: bool
+
+    # Exclusive groups 2
     major: bool
     minor: bool
+    patch: bool
     tag: bool
 
 
@@ -43,40 +53,74 @@ class Version:
         if self.arguments.get("verbose"):
             out.write(f"Installed Commitizen Version: {__version__}")
 
-        if not self.arguments.get("commitizen") and (
-            self.arguments.get("project") or self.arguments.get("verbose")
+        if self.arguments.get("commitizen"):
+            out.write(__version__)
+            return
+
+        if (
+            self.arguments.get("project")
+            or self.arguments.get("verbose")
+            or self.arguments.get("next")
+            or self.arguments.get("manual_version")
         ):
+            version_str = self.arguments.get("manual_version")
+            if version_str is None:
+                try:
+                    version_str = get_provider(self.config).get_version()
+                except NoVersionSpecifiedError:
+                    out.error("No project information in this project.")
+                    return
             try:
-                version = get_provider(self.config).get_version()
-            except NoVersionSpecifiedError:
-                out.error("No project information in this project.")
-                return
-            try:
-                version_scheme = get_version_scheme(self.config.settings)(version)
+                scheme_factory = get_version_scheme(self.config.settings)
             except VersionSchemeUnknown:
                 out.error("Unknown version scheme.")
                 return
 
+            try:
+                version = scheme_factory(version_str)
+            except InvalidVersion:
+                out.error(f"Invalid version: '{version_str}'")
+                return
+
+            if next_increment_str := self.arguments.get("next"):
+                if next_increment_str == "USE_GIT_COMMITS":
+                    # TODO: implement this
+                    raise NotImplementedError("USE_GIT_COMMITS is not implemented")
+
+                next_increment = VersionIncrement.safe_cast(next_increment_str)
+                # TODO: modify the interface of bump to accept VersionIncrement
+                version = version.bump(increment=str(next_increment))  # type: ignore[arg-type]
+
             if self.arguments.get("major"):
-                version = f"{version_scheme.major}"
-            elif self.arguments.get("minor"):
-                version = f"{version_scheme.minor}"
-            elif self.arguments.get("tag"):
+                out.write(version.major)
+                return
+            if self.arguments.get("minor"):
+                out.write(version.minor)
+                return
+            if self.arguments.get("patch"):
+                out.write(version.micro)
+                return
+
+            display_version: str
+            if self.arguments.get("tag"):
                 tag_rules = TagRules.from_settings(self.config.settings)
-                version = tag_rules.normalize_tag(version_scheme)
+                display_version = tag_rules.normalize_tag(version)
+            else:
+                display_version = str(version)
 
             out.write(
-                f"Project Version: {version}"
+                f"Project Version: {display_version}"
                 if self.arguments.get("verbose")
-                else version
+                else display_version
             )
             return
 
-        if self.arguments.get("major") or self.arguments.get("minor"):
-            out.error(
-                "Major or minor version can only be used with --project or --verbose."
-            )
-            return
+        for argument in ("major", "minor", "patch"):
+            if self.arguments.get(argument):
+                out.error(
+                    f"{argument} can only be used with MANUAL_VERSION, --project or --verbose."
+                )
+                return
 
         if self.arguments.get("tag"):
             out.error("Tag can only be used with --project or --verbose.")
--- a/commitizen/commands/version.py
+++ b/commitizen/commands/version.py
@@ -11,7 +11,7 @@ from commitizen.exceptions import NoVersionSpecifiedError, VersionSchemeUnknown
 from commitizen.providers import get_provider
 from commitizen.tags import TagRules
 from commitizen.version_increment import VersionIncrement
-from commitizen.version_schemes import get_version_scheme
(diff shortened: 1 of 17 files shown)


bearomorphism:
@Lee-W I already ran `pytest tests/commands/test_version_command.py -n auto --regen-all` to regenerate the test files, but the pipeline still fails. (The tests passed on my machine)

Do you have any ideas why this happen

commitizen/commands/version.py
```diff
@@ -41,24 +51,49 @@ def __call__(self) -> None:
...
+                # TODO: modify the interface of bump to accept VersionIncrement
+                version = version.bump(increment=str(next_increment))  # type: ignore[arg-type]
+
             if self.arguments.get("major"):
-                version = f"{version_scheme.major}"
-            elif self.arguments.get("minor"):
-                version = f"{version_scheme.minor}"
+                out.write(version.major)
+                return
+            if self.arguments.get("minor"):
```
woile, line 108: I realize now that this creates problems with the (not)monotonic kind of versions (and possible non-semver). I'm not sure what to do about it.

I think for now it's fine that if you diverge *too* much from semver in your custom version scheme, then you won't get the full range of features.

woile:
There was a way to run all the pre-commits, but I don't remember how it works 😅

tests/commands/test_version_command.py
```diff
@@ -185,6 +185,56 @@ def test_version_just_major_error_no_project(config, capsys, argument: str):
...
     assert not captured.out
     assert (
-        "Major or minor version can only be used with --project or --verbose."
-        in captured.err
+        "can only be used with MANUAL_VERSION, --project or --verbose." in captured.err
     )
+
+
+@pytest.mark.parametrize(
+    "next_version, current_version, expected_version",
```
woile: maybe rename `next_version` to `next_increment`?
In the user facing cli, it makes sense semantically: `--next=MAJOR` (give me the next major).

tests/commands/test_version_command.py
```diff
@@ -185,6 +185,56 @@ def test_version_just_major_error_no_project(config, capsys, argument: str):
...
+        {
+            "manual_version": current_version,
+            "next": next_version,
+        },
+    )()
+    captured = capsys.readouterr()
+    assert expected_version in captured.out
+
+
+def test_next_version_invalid_version(config, capsys):
```
woile, line 244: what about having a parametrize of the combinations that should fail:
```py
@pytest.mark.parametrize(
    "args",
    [
        # incomplete list:
        {"next"},
        # ...
        {"--verbose", "next"}
    ]
)
```

commitizen/commands/version.py
```diff
@@ -41,24 +51,49 @@ def __call__(self) -> None:
...
                 out.error("Unknown version scheme.")
                 return
 
+            try:
+                version = version_scheme(version_str)
+            except InvalidVersion:
+                out.error("Invalid version.")
+                return
+
+            if next_str := self.arguments.get("next"):
```
woile: what would happen here if I run `cz version --project --next`? I would expect that to work

woile:
Nice to see this going forward 🎉

commitizen/commands/version.py
bearomorphism: That also works.

bearomorphism: The behavior is the same as `cz version --next=*`

bearomorphism:
> @Lee-W I already ran `pytest tests/commands/test_version_command.py -n auto --regen-all` to regenerate the test files, but the pipeline still fails. (The tests passed on my machine)
> 
> Do you have any ideas why this happen

I see why the test failure happens. Please see my other PR #1726 .

bearomorphism:
I will rebase this branch after #1726 is merged. The test failure should be resolved then.

commitizen/commands/version.py
woile: But what would be the output?

bearomorphism: Oh I misunderstood your question. The output should depend on the commit history, right?
I forgot to implement that.

bearomorphism: I will do #1678 in another PR since it's a bit complicated (we need to consider `major_version_zero`, `bump_map`, etc. for this feature and I prefer not to just copy & paste code, need some refactoring)

woile: Indeed, that logic should be core and it should be possible to re-use it across commands.

commitizen/version_schemes.py
```diff
@@ -140,9 +140,7 @@ def bump(
         """
 
 
-# With PEP 440 and SemVer semantic, Scheme is the type, Version is an instance
-Version: TypeAlias = VersionProtocol
-VersionScheme: TypeAlias = type[VersionProtocol]
+# Note: version schemes are classes that initialize a VersionProtocol instance
```
noirbizarre: No, this introduces a non-necessary (as out of the scope of this PR) breaking change: users with importing those `TypeAlias` for their own version scheme will have an import error.
Also those type aliases gives semantics and make the code easier to read so I would not remove them.
And this change make the PR more verbose than it should.

commitizen/version_increment.py
```diff
@@ -0,0 +1,28 @@
...
+    NONE = 0
+    PATCH = 1
+    MINOR = 2
+    MAJOR = 3
+
+    def __str__(self) -> str:
+        return self.name
+
+    @classmethod
+    def safe_cast(cls, value: object) -> "VersionIncrement":
```
noirbizarre: Add `from __future__ import annotations` and remove the quotes to stay consistent with the existing code base.

Also it's not a cast but an instantiation, so maybe rename this in `from_value()`

commitizen/version_increment.py
```diff
@@ -0,0 +1,28 @@
+from enum import IntEnum
+
+
+class VersionIncrement(IntEnum):
```
noirbizarre, line 6: Why not a `StrEnum` with `auto()` fields?

noirbizarre (dismissed):
That's an interesting feature, one that I would totally use!

Also, I never realized that the `version` command was doing totally different things depending on the flags 😅 
So maybe it's the occasion to have a dedicated `--version` flag and have the `version` command focusing on the the project versioning and not  the Commitizen version. It would be cleaner, easier to understand and more standard. WDYT ?

> [!NOTE]
> I would not remove the Type aliases as it makes the PR bigger than it should and it introduce a breaking change (and reading the PR and issues, we know that it has been used by some users for some custom version schemes).

noirbizarre:
Also, the enumn part seems to overlap with #1518, so maybe #1518 should go first so this PR can reuse the Increment enum

bearomorphism:
> Also, the enumn part seems to overlap with https://github.com/commitizen-tools/commitizen/pull/1518, so maybe https://github.com/commitizen-tools/commitizen/pull/1518 should go first so this PR can reuse the Increment enum

I would prefer to merge this before #1518 because it is a bit larger than this PR.

> So maybe it's the occasion to have a dedicated --version flag and have the version command focusing on the the project versioning and not the Commitizen version. It would be cleaner, easier to understand and more standard. WDYT ?

+1. As a user, I often run `cz --version` and end up seeing an error. We can create another work item about adding the dedicated flag and then deprecate `cz version --commitizen` and other removable flags in the next major version.

commitizen/version_increment.py
bearomorphism: iirc there is a ruff rule for this. will check later when I have a moment

commitizen/version_schemes.py
bearomorphism: will revert, thanks

commitizen/version_increment.py
bearomorphism: Thanks, I didn't know `StrEnum`.

However, I want to make those `VersionIncrement` objects comparable with the order NONE < PATCH < MINOR < MAJOR in future PRs.

woile:
We had a `--version` in the past, but it was replaced by `version`, do you remember why @Lee-W ? 
It would be a breaking change if we go that route. If agreed, I can add it as a different task on the parent issue tracking this work #1677

commitizen/version_increment.py
bearomorphism: https://docs.astral.sh/ruff/rules/quoted-annotation/

tests/commands/test_version_command.py
bearomorphism: Sounds good.

bearomorphism:
@woile @noirbizarre @Lee-W 
Could you review and merge this PR, so we can move on to other `cz version` related task? Thanks!

woile:
There are some open conversations still

tests/commands/test_version_command.py
bearomorphism: fixed

commitizen/commands/version.py
```diff
@@ -43,40 +53,82 @@ def __call__(self) -> None:
...
 
+            try:
+                version = scheme_factory(version_str)
+            except InvalidVersion:
+                out.error(f"Invalid version: '{version_str}'")
+                return
+
+            if next_increment_str := self.arguments.get("next"):
+                if next_increment_str == "USE_GIT_COMMITS":
+                    out.error("--next USE_GIT_COMMITS is not implemented yet.")
```
woile, line 87: this is what we have to refactor later, right? move the logic out of `bump` so it can be reused

woile (approved):
Looks good! Maybe @Lee-W do you have any thoughts?

commitizen/commands/version.py
bearomorphism: future work

woile (approved):
LGTM

PR #1726, opened 2025-12-13
test(commands): centralize all --help tests in a file to dedup code

bearomorphism:
# Test improvements

1. Addressed the issues about missing `skip_below_py_3_13` and missing forced consistent environment variables to reduce the chance of unnecessary file regression test failures.
2. Extracted duplicated logics.


tests/utils.py
```diff
@@ -1,29 +1,17 @@
...
 
-skip_below_py_3_10 = pytest.mark.skipif(
-    sys.version_info < (3, 10),
-    reason="The output message of argparse is different between Python 3.10 and lower than Python 3.10",
-)
-
-skip_below_py_3_13 = pytest.mark.skipif(
-    sys.version_info < (3, 13),
-    reason="The output message of argparse is different between Python 3.13 and lower than Python 3.13",
-)
```
bearomorphism, line 25: The variable name itself does not reflect why we want to set this variable... this may make it hard for future contributors to understand the code correctly

tests/commands/test_common_command.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Note: If the command description changes, please run `pytest tests/commands/test_common_command.py --regen-all` to regenerate the test files.
+    """
+    # Force consistent terminal output
+    monkeypatch.setenv("COLUMNS", "80")
+    monkeypatch.setenv("TERM", "dumb")
+    monkeypatch.setenv("LC_ALL", "C")
+    monkeypatch.setenv("LANG", "C")
+    monkeypatch.setenv("NO_COLOR", "1")
+    monkeypatch.setenv("PAGER", "cat")
```
bearomorphism, line 46: Copied from AI output. I know COLUMNS is necessary, but not sure for the rest

bearomorphism:
@woile @Lee-W could you review this PR when you have a moment? This PR contains the resolution of error when someone is trying to update CLI `--help` description. Some other PRs updating CLI options have dependencies on this PR.

Thanks!

woile:
I don't understand what this is supposed to fix

bearomorphism:
> I don't understand what this is supposed to fix

The behavior of `--help` of argparse is different in some cases between python 3.13 and below 3.13 as the message of original `skip_below_py_3_13` says. Some of `--help` tests still uses `skip_below_py_3_10` before this PR, and those tests still pass in the pipeline luckily.

When we try to add new options to the CLI, the help text changes, and usually we run `pytest --regen-all` to regenerate the file_regression test file, but the regenerated content may be different depends on the local environment (monkeypatch in this PR addresses this issue).

bearomorphism:
The following pipeline result is a false alarm test failure caused by forgetting to skip the test below python 3.13.

https://github.com/commitizen-tools/commitizen/actions/runs/20207591917/job/58008470452?pr=1724

Test passes for python version 3.13, 3.14 but not on 3.10

This PR can fix it.

tests/commands/test_common_command.py
```diff
@@ -0,0 +1,69 @@
...
+from pytest_mock import MockFixture
+
+from commitizen import cli
+from commitizen.commands import Example, Info, ListCz, Schema
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 13),
+    reason="The output message of argparse is different between Python 3.13 and lower than Python 3.13",
+)
```
noirbizarre, line 13: Can't we have different snapshots for Python <3.13 and >=3.13 ?
It seems weird to have some tests skipped for some Python version for things that are existing on those Python versions

tests/commands/test_common_command.py
noirbizarre, line 46: I think this could be a fixture like `fixed_term` for any test relying on output.
Basically, what it does:
- `COLUMNS` fix the width of the term
- `TERM=dumb` deactivate reactive output on libs supporting this (like spinners, progress bars...), better for static snapshots.
- `LC_ALL=C` and `LANG=C` disable localization and utf-8 chars in most CLI tools, it ensure test snapshot doesn't fail on non-english setups 
- `NO_COLOR` is the now "standard" way of disabling ANSI color support (cf. https://no-color.org/)
- `PAGER=cat` will avoid user defined pagers like `bat` to change the output.

noirbizarre (approved):
Good to me, but I added some comment for possible later improvements

tests/commands/test_common_command.py
bearomorphism: Sounds good. I will check how to do it later and will do in another PR.

tests/commands/test_common_command.py
bearomorphism: will do in another PR

PR #1733, opened 2025-12-23
fix(cargo_provider): support workspace virtual manifests

### LINKED ISSUE #1732: cz bump crashes with version_provider=cargo on workspace virtual manifest (missing [package] / TypeError on version.workspace)
### Description

Commitizen crashes when using `version_provider = "cargo"` on a Cargo **workspace virtual manifest** (root `Cargo.toml` contains `[workspace]` and `[workspace.package]` but **no** `[package]`). Our workspace members inherit version via `version.workspace = true`. Running `cz bump` fails inside `commitizen/providers/cargo_provider.py` while trying to update `Cargo.lock`.

This prevents using Commitizen’s cargo provider for monorepo-style Rust workspaces where the version is centralized in `[workspace.package].version`.

### Steps to reproduce

1. Create a Cargo workspace that uses a **virtual manifest** at the root (no `[package]`), with a centralized version:

   ```toml
   # ./Cargo.toml (workspace root)
   [workspace]
   members = ["crates/a"]

   [workspace.package]
   version = "0.1.0"
   edition = "2024"
   ```

2. Create a member crate that inherits the version via `version.workspace = true`:

   ```toml
   # ./crates/a/Cargo.toml
   [package]
   name = "a"
   version.workspace = true
   edition.workspace = true
   ```

3. Configure Commitizen to use the cargo provider:

   ```toml
   # ./cz.toml
   [tool.commitizen]
   version_scheme = "semver"
   version_provider = "cargo"
   tag_format = "v$version"
   ```

4. Run:

   ```bash
   cz bump --yes
   ```

5. Observe the crash inside `commitizen/providers/cargo_provider.py` (during manifest parsing / Cargo.lock update).

### Current behavior

`cz bump` crashes with exceptions from `cargo_provider.py`:

* `tomlkit.exceptions.NonExistentKey: 'Key "package" does not exist.'` (root manifest has no `[package]`)
* then `TypeError: string indices must be integers, not 'str'` (when accessing `package_content["version"]["workspace"]` for members)

The command terminates without completing the bump/tag operation.

### Desired behavior

`cz bump` should support Cargo workspaces with a **virtual manifest** (root `Cargo.toml` has `[workspace]` / `[workspace.package]` but no `[package]`) and members using `version.workspace = true`.

Concretely:

* Read and bump the canonical version from `[workspace.package].version`.
* Do not assume `[package]` exists in the workspace root manifest.
* When scanning/updating member manifests, correctly handle `version.workspace = true` (a nested table / boolean) without type errors.

### Screenshots

_No response_

### Environment

(computation) juncheng@localhost:~/projects/computation$ cz version --report
Commitizen Version: 4.10.1
Python Version: 3.11.12 (main, Apr  9 2025, 04:04:00) [Clang 20.1.0 ]
Operating System: Linux

jcfangc:
## Description

Fix `cargo_provider` so `cz bump` works with Cargo **workspace virtual manifests** (root `Cargo.toml` containing `[workspace]` / `[workspace.package]` but no `[package]`) when workspace members inherit the version via `package.version.workspace = true`.

Previously, `cz bump` could crash in `commitizen/providers/cargo_provider.py` with:

* `NonExistentKey: Key "package" does not exist` (virtual manifest has no `[package]`)
* `TypeError: expected root version to be a string` / unsafe assumptions around `package.version` shape

This change makes the provider:

* Read/write the version from `[workspace.package].version` when present (fallback to `[package].version` for non-workspace manifests).
* When updating `Cargo.lock`:

  * For non-workspace manifests, update the root package entry.
  * For workspace virtual manifests, scan `workspace.members` (respecting `exclude`) and update lock entries for crates that declare `package.version.workspace = true`.

Fixes #1732

---

## Checklist

* [x] I have read the contributing guidelines
* [x] Added test coverage for the introduced behavior and edge cases
* [x] Ran the test suite locally
* [x] Manually tested the behavior (workspace virtual manifest + non-workspace crate)
* [ ] Updated documentation (not needed for this change)

> Note: the project uses `uv` / `poe` tasks (not Poetry). I ran the equivalent `poe` / `uv`-based test commands.

---

## Expected Behavior

`cz bump` with `version_provider = "cargo"` should succeed for Cargo workspaces using a virtual manifest and centralized versioning in `[workspace.package].version`, without crashing. The workspace version should be updated, and `Cargo.lock` entries for crates inheriting the workspace version (`package.version.workspace = true`) should be updated safely.

Non-workspace projects (root `[package]`) keep existing behavior.

---

### Manual test (workspace virtual manifest)

1. Create a workspace root `Cargo.toml`:

   ```toml
   [workspace]
   members = ["crates/*"]

   [workspace.package]
   version = "1.0.0"
   edition = "2024"
   rust-version = "1.92"
   repository = "https://example.com/repo.git"
   ```

2. Add a member crate `crates/common-core/Cargo.toml`:

   ```toml
   [package]
   name = "common-core"
   version.workspace = true
   edition.workspace = true
   rust-version.workspace = true
   repository.workspace = true
   ```

3. Generate `Cargo.lock` once:

   ```bash
   cargo build
   ```

4. Configure Commitizen:

   ```toml
   [tool.commitizen]
   tag_format = "v$version"
   version_scheme = "semver"
   version_provider = "cargo"
   version_files = ["Cargo.toml:workspace.package.version"]
   ```

5. Run:

   ```bash
   cz bump --yes
   ```

6. Verify:

   * `[workspace.package].version` is updated (e.g., `1.0.0 -> 1.1.0`)
   * `Cargo.lock` entries for workspace crates inheriting the version are updated
   * no exceptions are raised

### Manual test (non-workspace manifest)

1. Use a standard crate with root `[package]` and an existing `Cargo.lock`.
2. Run `cz bump --yes` and verify the root package version in `Cargo.lock` is updated as before.

---

## Additional Context

* Reported in #1732.
* Targets workspace virtual manifests (no root `[package]`) and avoids unsafe assumptions about the TOML shape of version fields.

### DIFF
--- a/commitizen/providers/cargo_provider.py
+++ b/commitizen/providers/cargo_provider.py
@@ -45,48 +45,72 @@ class CargoProvider(TomlProvider):
             self.set_lock_version(version)
 
     def set_lock_version(self, version: str) -> None:
-        cargo_toml = parse(self.file.read_text())
-        cargo_lock = parse(self.lock_file.read_text())
-        packages = cargo_lock["package"]
+        cargo_toml_content = parse(self.file.read_text())
+        cargo_lock_content = parse(self.lock_file.read_text())
+        packages = cargo_lock_content["package"]
+
         if TYPE_CHECKING:
             assert isinstance(packages, AoT)
 
-        root_pkg = _table_get(cargo_toml, "package")
+        root_pkg = _table_get(cargo_toml_content, "package")
         if root_pkg is not None:
             name = root_pkg.get("name")
             if isinstance(name, str):
                 _lock_set_versions(packages, {name}, version)
-            self.lock_file.write_text(dumps(cargo_lock))
+            self.lock_file.write_text(dumps(cargo_lock_content))
             return
 
-        ws = _table_get(cargo_toml, "workspace") or {}
-        members = cast("list[str]", ws.get("members", []) or [])
-        excludes = cast("list[str]", ws.get("exclude", []) or [])
-        inheriting = _workspace_inheriting_member_names(members, excludes)
+        ws = _table_get(cargo_toml_content, "workspace") or {}
+        member_globs = cast("list[str]", ws.get("members", []) or [])
+        exclude_globs = cast("list[str]", ws.get("exclude", []) or [])
+        inheriting = _workspace_inheriting_member_names(member_globs, exclude_globs)
         _lock_set_versions(packages, inheriting, version)
-        self.lock_file.write_text(dumps(cargo_lock))
+        self.lock_file.write_text(dumps(cargo_lock_content))
 
 
 def _table_get(doc: TOMLDocument, key: str) -> DictLike | None:
-    """Return a dict-like table for `key` if present, else None (type-safe for Pylance)."""
+    """Get a TOML table by key as a dict-like object.
+
+    Returns:
+        The value at `doc[key]` cast to a dict-like table (supports `.get`) if it
+        exists and is table/container-like; otherwise returns None.
+
+    Rationale:
+        tomlkit returns loosely-typed Container/Table objects; using a small
+        helper keeps call sites readable and makes type-checkers happier.
+    """
     try:
-        v = doc[key]  # tomlkit returns Container/Table-like; typing is loose
+        value = doc[key]
     except NonExistentKey:
         return None
-    return cast("DictLike", v) if hasattr(v, "get") else None
+    return cast("DictLike", value) if hasattr(value, "get") else None
 
 
 def _root_version_table(doc: TOMLDocument) -> DictLike:
-    """Prefer [workspace.package]; fallback to [package]."""
-    ws = _table_get(doc, "workspace")
-    if ws is not None:
-        pkg = ws.get("package")
-        if hasattr(pkg, "get"):
-            return cast("DictLike", pkg)
-    pkg = _table_get(doc, "package")
-    if pkg is None:
+    """Return the table that owns the "root" version field.
+
+    This provider supports two layouts:
+
+    1) Workspace virtual manifests:
+            [workspace.package]
+            version = "x.y.z"
+
+    2) Regular crate（non-workspace root manifest）:
+            [package]
+            version = "x.y.z"
+
+    The selected table is where `get()` reads from and `set()` writes to.
+    """
+    workspace_table = _table_get(doc, "workspace")
+    if workspace_table is not None:
+        workspace_package_table = workspace_table.get("package")
+        if hasattr(workspace_package_table, "get"):
+            return cast("DictLike", workspace_package_table)
+
+    package_table = _table_get(doc, "package")
+    if package_table is None:
         raise NonExistentKey("expected either [workspace.package] or [package]")
-    return pkg
+    return package_table
 
 
 def _is_workspace_inherited_version(v: Any) -> bool:
@@ -94,19 +118,38 @@ def _is_workspace_inherited_version(v: Any) -> bool:
 
 
 def _iter_member_dirs(
-    members: Iterable[str], excludes: Iterable[str]
+    member_globs: Iterable[str], exclude_globs: Iterable[str]
 ) -> Iterable[Path]:
-    for pat in members:
-        for p in glob.glob(pat, recursive=True):
-            if any(fnmatch.fnmatch(p, ex) for ex in excludes):
+    """Yield workspace member directories matched by `member_globs`, excluding `exclude_globs`.
+
+    Cargo workspaces define members/exclude as glob patterns (e.g. "crates/*").
+    This helper expands those patterns and yields the corresponding directories
+    as `Path` objects, skipping any matches that satisfy an exclude glob.
+
+    Kept as a helper to make call sites read as domain logic ("iterate member dirs")
+    rather than glob/filter plumbing.
+    """
+    for member_glob in member_globs:
+        for match in glob.glob(member_glob, recursive=True):
+            if any(fnmatch.fnmatch(match, ex) for ex in exclude_globs):
                 continue
-            yield Path(p)
+            yield Path(match)
 
 
 def _workspace_inheriting_member_names(
     members: Iterable[str], excludes: Iterable[str]
 ) -> set[str]:
-    out: set[str] = set()
+    """Return workspace member crate names that inherit the workspace version.
+
+    A member is considered "inheriting" when its Cargo.toml has:
+        [package]
+        version.workspace = true
+
+    This scans `members` globs (respecting `excludes`) and returns the set of
+    `[package].name` values for matching crates. Missing/invalid Cargo.toml files
+    are ignored.
+    """
+    inheriting_member_names: set[str] = set()
     for d in _iter_member_dirs(members, excludes):
         cargo_file = d / "Cargo.toml"
         if not cargo_file.exists():
@@ -118,13 +161,25 @@ def _workspace_inheriting_member_names(
         if _is_workspace_inherited_version(pkgd.get("version")):
             name = pkgd.get("name")
             if isinstance(name, str):
-                out.add(name)
-    return out
+                inheriting_member_names.add(name)
+    return inheriting_member_names
 
 
-def _lock_set_versions(packages: Any, names: set[str], version: str) -> None:
-    if not names:
+def _lock_set_versions(packages: Any, package_names: set[str], version: str) -> None:
+    """Update Cargo.lock package entries in-place.
+
+    Args:
+        packages: `Cargo.lock` parsed TOML "package" array (AoT-like). Mutated in-place.
+        package_names: Set of package names whose `version` field should be updated.
+        version: New version string to write.
+
+    Notes:
+        We use `enumerate` + index assignment because tomlkit AoT entries may be
+        Container-like and direct mutation patterns vary; indexed assignment is
+        reliable for updating the underlying document.
+    """
+    if not package_names:
         return
-    for i, p in enumerate(packages):
-        if getattr(p, "get", None) and p.get("name") in names:
+    for i, pkg_entry in enumerate(packages):
+        if getattr(pkg_entry, "get", None) and pkg_entry.get("name") in package_names:
             packages[i]["version"] = version
--- a/commitizen/providers/cargo_provider.py
+++ b/commitizen/providers/cargo_provider.py
@@ -1,14 +1,18 @@
 from __future__ import annotations
 
-import fnmatch, glob
+import fnmatch
+import glob
 from pathlib import Path
-from typing import TYPE_CHECKING, Iterable, Any, cast
+from typing import TYPE_CHECKING, Any, cast
 
 from tomlkit import TOMLDocument, dumps, parse
 from tomlkit.exceptions import NonExistentKey
+
 from commitizen.providers.base_provider import TomlProvider
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from tomlkit.items import AoT
 
 
@@ -56,8 +60,8 @@ class CargoProvider(TomlProvider):
             return
 
         ws = _table_get(cargo_toml, "workspace") or {}
-        members = cast(list[str], ws.get("members", []) or [])
-        excludes = cast(list[str], ws.get("exclude", []) or [])
+        members = cast("list[str]", ws.get("members", []) or [])
+        excludes = cast("list[str]", ws.get("exclude", []) or [])
         inheriting = _workspace_inheriting_member_names(members, excludes)
         _lock_set_versions(packages, inheriting, version)
         self.lock_file.write_text(dumps(cargo_lock))
@@ -69,7 +73,7 @@ def _table_get(doc: TOMLDocument, key: str) -> DictLike | None:
         v = doc[key]  # tomlkit returns Container/Table-like; typing is loose
     except NonExistentKey:
         return None
-    return cast(DictLike, v) if hasattr(v, "get") else None
+    return cast("DictLike", v) if hasattr(v, "get") else None
 
 
 def _root_version_table(doc: TOMLDocument) -> DictLike:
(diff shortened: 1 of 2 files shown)


bearomorphism:
Hi @jcfangc ,
Thank you for your contribution! Please follow the contribution guideline. You forget to run `poetry format` on the latest iteration.

jcfangc:
> Hi @jcfangc , Thank you for your contribution! Please follow the contribution guideline. You forget to run `poetry format` on the latest iteration.

Sorry about that—I’ll run poetry format and push an update asap.

jcfangc:
When I first pulled the branch locally, I didn’t have `uv.lock` in the working tree, and `poetry format` isn’t a Poetry subcommand in my setup (Poetry 1.1.12 reports: “The command 'format' is not defined”).

Following the repo’s uv/poe workflow, I synced dependencies (needed `uv sync --all-groups` to make `ruff` available), then ran `uv run poe format`, and pushed the formatting update.

Just to clarify / FYI.

noirbizarre:
I think it would be easier and cleaner to just rebase/squash on the current master.
You are not changing dependencies, so your PR should not touch the `uv.lock` file.
To avoid this side effect, run `uv sync --frozen --all-groups` instead of just `uv sync --all-groups` (`uv` default behavior is to check and update the lock).
Also, you might want to update your `uv` version, it is trying to downgrade the lock version so it might be outdated.

jcfangc:
> I think it would be easier and cleaner to just rebase/squash on the current master. You are not changing dependencies, so your PR should not touch the `uv.lock` file. To avoid this side effect, run `uv sync --frozen --all-groups` instead of just `uv sync --all-groups` (`uv` default behavior is to check and update the lock). Also, you might want to update your `uv` version, it is trying to downgrade the lock version so it might be outdated.

Thanks for the feedback.

I’ve rebased/squashed the branch on top of the current master and ensured the PR no longer touches lockfiles:
- reverted/restored `uv.lock` and `poetry.lock` to match `master` (no lockfile changes in the PR diff now)

Going forward I’ll use `uv sync --frozen --all-groups` to avoid accidental lock updates, and I’ll also update my local `uv` to prevent lock format/version downgrades.

All checks are now green.

woile:
The rebase did not succeed, only your commits should appear on "Commits".

On top of that, I wonder if #1715 (in-progress) addresses this issue, can you take a look? It also changes the read order, if there are no complaints about that I was planning on merging it

jcfangc:
> The rebase did not succeed, only your commits should appear on "Commits".
> 
> On top of that, I wonder if #1715 (in-progress) addresses this issue, can you take a look? It also changes the read order, if there are no complaints about that I was planning on merging it

Hi @woile, thanks for flagging #1715 — they are related but not the same.

* **#1715** focuses on two smaller cargo-provider behaviors:

  1. ensuring we use the **same precedence/order** when **reading vs writing** the version between `[package]` and workspace metadata, and
  2. clarifying/fixing the “fixed version for a workspace member” case: there is no `workspace.version = "x.y.z"` in Cargo spec; a member crate must use `version = "x.y.z"` even if it’s part of a workspace.

* **#1733** targets a broader and different failure mode: **workspace virtual manifests** (root `Cargo.toml` with `[workspace]` / `[workspace.package]` but **no `[package]`**). It:

  1. prevents crashes when `[package]` is missing / when `version.workspace` shape appears,
  2. **reads/writes** version from **`[workspace.package].version`** when present (fallback to `[package].version` for non-workspace manifests), and
  3. updates `Cargo.lock` appropriately for virtual manifests by scanning `workspace.members` and updating lock entries for crates with `package.version.workspace = true`.

So #1715 does **not** address the virtual-manifest crash nor the lockfile update logic; #1733 does. The overlap is mostly around “precedence when both `[package]` and `[workspace.package]` exist”, which is a policy choice.

If you’d like to merge #1715 first, I can rebase #1733 on top of it and drop any overlapping hunks/tests. Alternatively, I can incorporate the #1715 precedence decision into #1733 so we end up with a single coherent behavior.

jcfangc:
> The rebase did not succeed, only your commits should appear on "Commits".
> 
> On top of that, I wonder if #1715 (in-progress) addresses this issue, can you take a look? It also changes the read order, if there are no complaints about that I was planning on merging it

I rebuilt the branch on top of current upstream/master by cherry-picking only my 4 commits (no merge commits).
The PR no longer touches any lockfiles (uv.lock / poetry.lock).

Could you please take another look? @woile

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
+    members: Iterable[str], excludes: Iterable[str]
+) -> Iterable[Path]:
+    for pat in members:
+        for p in glob.glob(pat, recursive=True):
+            if any(fnmatch.fnmatch(p, ex) for ex in excludes):
+                continue
+            yield Path(p)
+
+
+def _workspace_inheriting_member_names(
```
woile: what is this supposed to do?

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
+        pkg = ws.get("package")
+        if hasattr(pkg, "get"):
+            return cast("DictLike", pkg)
+    pkg = _table_get(doc, "package")
+    if pkg is None:
+        raise NonExistentKey("expected either [workspace.package] or [package]")
+    return pkg
+
+
+def _is_workspace_inherited_version(v: Any) -> bool:
```
woile: what is this supposed to do?

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
+        for p in glob.glob(pat, recursive=True):
+            if any(fnmatch.fnmatch(p, ex) for ex in excludes):
+                continue
+            yield Path(p)
+
+
+def _workspace_inheriting_member_names(
+    members: Iterable[str], excludes: Iterable[str]
+) -> set[str]:
+    out: set[str] = set()
```
bearomorphism: please use a meaningful variable name for the return value

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
+
+
+def _iter_member_dirs(
+    members: Iterable[str], excludes: Iterable[str]
+) -> Iterable[Path]:
+    for pat in members:
+        for p in glob.glob(pat, recursive=True):
+            if any(fnmatch.fnmatch(p, ex) for ex in excludes):
+                continue
+            yield Path(p)
```
bearomorphism: I'd prefer not to extract this function unless we can reuse the logic. Also the variable names are somewhat confusing. For example, (1) what does `members` mean? (2) why not `for member in members:`?

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
+            name = root_pkg.get("name")
+            if isinstance(name, str):
+                _lock_set_versions(packages, {name}, version)
+            self.lock_file.write_text(dumps(cargo_lock))
+            return
+
+        ws = _table_get(cargo_toml, "workspace") or {}
+        members = cast("list[str]", ws.get("members", []) or [])
+        excludes = cast("list[str]", ws.get("exclude", []) or [])
+        inheriting = _workspace_inheriting_member_names(members, excludes)
```
bearomorphism: You are making the variable names more confusing. The original names contain more information.

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
             self.set_lock_version(version)
 
     def set_lock_version(self, version: str) -> None:
-        cargo_toml_content = parse(self.file.read_text())
-        cargo_lock_content = parse(self.lock_file.read_text())
-        packages = cargo_lock_content["package"]
-
+        cargo_toml = parse(self.file.read_text())
+        cargo_lock = parse(self.lock_file.read_text())
+        packages = cargo_lock["package"]
```
bearomorphism: Please revert

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
     def get(self, document: TOMLDocument) -> str:
-        out = _try_get_workspace(document)["package"]["version"]
-        if TYPE_CHECKING:
-            assert isinstance(out, str)
-        return out
+        t = _root_version_table(document)
+        v = t.get("version")
+        if not isinstance(v, str):
+            raise TypeError("expected root version to be a string")
+        return v
```
bearomorphism, line 37: Please use more meaningful variable names. Do not use `t` or `v`.

commitizen/providers/cargo_provider.py
```diff
@@ -3,103 +3,128 @@
...
+            name = pkgd.get("name")
+            if isinstance(name, str):
+                out.add(name)
+    return out
+
+
+def _lock_set_versions(packages: Any, names: set[str], version: str) -> None:
+    if not names:
+        return
+    for i, p in enumerate(packages):
```
bearomorphism: Do not use single letter for variable names except indices (e.g. `for i in range(...)`)

bearomorphism (changes requested):
Thank you for your contribution! Here are some feedbacks:

commitizen/providers/cargo_provider.py
jcfangc: I kept `_iter_member_dirs` as a small helper because it represents a domain concept (“expand workspace member globs with excludes”) and keeps the call site focused on the higher-level logic (identify version-inheriting members), rather than mixing in glob/exclude plumbing.

That said, I agree the naming was confusing — I’ve renamed the parameters/locals to make it explicit these are glob patterns (e.g. `member_globs` / `exclude_globs`) and added a docstring to clarify intent.

If you still prefer it inlined, I’m happy to inline it, but I’d like to keep it as a readability helper since it’s already reused by the member-scanning logic.

jcfangc:
Hi @bearomorphism @woile @noirbizarre @Lee-W — I’ve addressed the requested changes and force-pushed updates.

* Fixed the review comments in `cargo_provider.py`, hardened parsing/lock updates, removed an unused `type: ignore`, and ran the formatter.
* No `uv.lock` / `poetry.lock` changes; all checks are green.

Could you please re-review when you have a moment? Thanks.

noirbizarre (commented):
Just fix the naming concerns (`t`, `v`...) which makes maintenance harder when you hop from review to review and then it's good to merge (to me)

commitizen/providers/cargo_provider.py
```diff
@@ -3,39 +3,41 @@
...
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from tomlkit.items import AoT
 
 
-class CargoProvider(TomlProvider):
-    """
-    Cargo version management
+DictLike = dict[str, Any]
```
Lee-W, line 19: I'm not a fan of this 🤔 As it's actually a dict

commitizen/providers/cargo_provider.py
```diff
@@ -3,39 +3,41 @@
...
     def get(self, document: TOMLDocument) -> str:
-        out = _try_get_workspace(document)["package"]["version"]
-        if TYPE_CHECKING:
-            assert isinstance(out, str)
-        return out
+        t = _root_version_table(document)
+        v = t.get("version")
+        if not isinstance(v, str):
+            raise TypeError("expected root version to be a string")
+        return v
```
Lee-W, line 37:
```suggestion
        table = _root_version_table(document)
        version = table.get("version")
        if not isinstance(version, str):
            raise TypeError("expected root version to be a string")
        return version
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
-            assert isinstance(workspace, dict)
-        return workspace
+        value = doc[key]
     except NonExistentKey:
-        return document
+        return None
+    return cast("DictLike", value) if hasattr(value, "get") else None
+
+
+def _root_version_table(doc: TOMLDocument) -> DictLike:
```
Lee-W, line 89:
```suggestion
def _extract_root_version_table(doc: TOMLDocument) -> DictLike:
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
-                            members_inheriting.append(package_name)
-                    except NonExistentKey:
-                        pass
-
-            for i, package in enumerate(packages):
-                if package["name"] in members_inheriting:
-                    cargo_lock_content["package"][i]["version"] = version  # type: ignore[index]
-
+        root_pkg = _table_get(cargo_toml_content, "package")
+        if root_pkg is not None:
```
Lee-W, line 56:
```suggestion
        if (root_pkg := _table_get(cargo_toml_content, "package")) is not None:
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
+            continue
+        pkgd = cast("DictLike", pkg)
+        if _is_workspace_inherited_version(pkgd.get("version")):
+            name = pkgd.get("name")
+            if isinstance(name, str):
+                inheriting_member_names.add(name)
+    return inheriting_member_names
+
+
+def _lock_set_versions(packages: Any, package_names: set[str], version: str) -> None:
```
Lee-W, line 168:
```suggestion
def _set_lock_versions(packages: Any, package_names: set[str], version: str) -> None:
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
+            version = "x.y.z"
+
+    2) Regular crate（non-workspace root manifest）:
+            [package]
+            version = "x.y.z"
+
+    The selected table is where `get()` reads from and `set()` writes to.
+    """
+    workspace_table = _table_get(doc, "workspace")
+    if workspace_table is not None:
```
Lee-W, line 105:
```suggestion
    if (workspace_table := _table_get(doc, "workspace")) is not None:
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
+        workspace_package_table = workspace_table.get("package")
+        if hasattr(workspace_package_table, "get"):
+            return cast("DictLike", workspace_package_table)
+
+    package_table = _table_get(doc, "package")
+    if package_table is None:
```
Lee-W, line 111:
```suggestion
    if (package_table := _table_get(doc, "package")) is None:
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
+        raise NonExistentKey("expected either [workspace.package] or [package]")
+    return package_table
+
+
+def _is_workspace_inherited_version(v: Any) -> bool:
+    return hasattr(v, "get") and cast("DictLike", v).get("workspace") is True
```
Lee-W, line 117:
```suggestion
def _is_workspace_inherited_version(version_table: Any) -> bool:
    return hasattr(version_table, "get") and cast("DictLike", version_table).get("workspace") is True
```

commitizen/providers/cargo_provider.py
```diff
@@ -50,56 +52,134 @@ def set_lock_version(self, version: str) -> None:
...
+    rather than glob/filter plumbing.
+    """
+    for member_glob in member_globs:
+        for match in glob.glob(member_glob, recursive=True):
+            if any(fnmatch.fnmatch(match, ex) for ex in exclude_globs):
+                continue
+            yield Path(match)
+
+
+def _workspace_inheriting_member_names(
```
Lee-W, line 139:
```suggestion
def _collect_workspace_version_inheriting_members(
```

Lee-W (commented):
a few nits, but high-level idea looks good

bearomorphism:
conflict. converting this PR to draft

PR #1738, opened 2025-12-27
ci: run unittests only on python and dependency change

### LINKED ISSUE #1697: Skip python package workflow when there is only documentation change
### Description

As title.

### Possible Solution

_No response_

### Additional context

_No response_

### Related issues

_No response_

bearomorphism:
Closes #1697 

# Expected behavior

The unittest workflow should only run when there is workflow change, `commitizen/**` change or `tests/**` change.

# Steps to test

See #1774 

Trigger jobs:

- [This skipped-workflow](https://github.com/commitizen-tools/commitizen/actions/runs/20746580719/job/59565314348) with no python code change.
- [This workflow](https://github.com/commitizen-tools/commitizen/actions/runs/20746719338/job/59565721480?pr=1774) with python code change, the unittest job is indeed triggered.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -8,25 +8,50 @@ on:
   workflow_dispatch:
 
 jobs:
+  detect_changes:
+    runs-on: ubuntu-latest
+    outputs:
+      relevant: ${{ steps.filter.outputs.relevant }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - id: filter
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            relevant:
+              - "commitizen/**"
+              - "tests/**"
+              - ".github/workflows/**"
   python-check:
+    needs: detect_changes
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         platform: [ubuntu-22.04, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
+      - name: No relevant changes, fast-path success
+        if: ${{ needs.detect_changes.outputs.relevant != 'true' }}
+        run: |
+          echo "No relevant file changes; skipping tests and linters."
       - uses: actions/checkout@v6
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
         with:
           fetch-depth: 0
       - name: Set up Python ${{ matrix.python-version }}
         uses: astral-sh/setup-uv@v7
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
         run: |
           uv --version
           uv sync --frozen --group base --group test --group linters
       - name: Run tests and linters
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
         run: |
           git config --global user.email "action@github.com"
           git config --global user.name "GitHub Action"
@@ -34,11 +59,12 @@ jobs:
         shell: bash
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
       - name: Upload test results to Codecov
         uses: codecov/codecov-action@v5
-        if: ${{ !cancelled() }}
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' && !cancelled() }}
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           report_type: test_results


Lee-W (approved):

Lee-W:
oh... this needs extra care on the setting side. I don't have a clear idea how to cater this change yet 🤔

bearomorphism:
It seems that I don't have the permission to read branch protection rules.

bearomorphism:
Waiting for adjusting branch protection rules

noirbizarre:
The problem is that it's not possible to reflect that condition in the branch protection rules.
The solution would be to have a closing job, one that would be required to merge the PR (it would make easier to handle supported version changes too), and have a home-made check skipping the tests (because the path selector applies to the workflow, not jobs).
Or there might be some action dedicated to do that, but it's not easy to make this using Github Actions `path` only.

So not sure if it's worth it (given tests are now also running faster)

bearomorphism:
Agree. Now I prefer not to touch the branch protection rules.

Potential solution by ChatGPT:

```
name: CI

on:
  pull_request:

jobs:
  detect_changes:
    runs-on: ubuntu-latest
    outputs:
      relevant: ${{ steps.filter.outputs.relevant }}
    steps:
      - uses: actions/checkout@v4

      - id: filter
        uses: dorny/paths-filter@v3
        with:
          filters: |
            relevant:
              - "src/**"
              - "tests/**"
              - ".github/workflows/**"

  build_and_test:
    needs: detect_changes
    if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: echo "Run build/test here"

  early_success:
    needs: detect_changes
    if: ${{ needs.detect_changes.outputs.relevant != 'true' }}
    runs-on: ubuntu-latest
    steps:
      - run: echo "No relevant file changes; skipping remaining jobs."

```

Behavior:

- If relevant files changed → build_and_test runs.
- If not → early_success runs, and build_and_test is skipped.
- The overall workflow finishes successfully (skipped jobs do not fail a run).

noirbizarre:
Thanks for the action! 🙏🏼 I'll keep it, this is something I often need.

However this wouldn't solve the branch protection issue.
I would change it this way to have both the branch protection working and the benefit from the closure task:

```yaml
name: CI

on:
  pull_request:

jobs:
  detect_changes:
    runs-on: ubuntu-latest
    outputs:
      relevant: ${{ steps.filter.outputs.relevant }}
    steps:
      - uses: actions/checkout@v4
      - id: filter
        uses: dorny/paths-filter@v3
        with:
          filters: |
            relevant:
              - "src/**"
              - "tests/**"
              - ".github/workflows/**"

  python-check:
    needs: detect_changes
    if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: echo "Run build/test here"

  success:
    needs: [python-check]
	# Force to run when python-check has been skipped
	# Skipped if python-check failed
    if: ${{ needs.detect_changes.outputs.relevant != 'true' && always() || !failure() }}
    runs-on: ubuntu-latest
    steps:
      - run: echo "CI Success"
```
This way we can use `success` as only branch protection condition:
- will run ASAP if no changes require testing
- will run only if test succeeds
- won't have to be changed when we move to Python 3.11-3.15

However, for this to be complete, you will need to split out the linting part.
Enabling pre-commit.ci could do the job but we would have to move the `check-commit` as a pre-commit hook.

Lee-W:
Yep, I think what @noirbizarre suggests is better. We can't configurate the branch protection rule that way

bearomorphism:
I cannot access the branch protection rule, so the latest iteration is based on my assumption that `python-check` should succeed for each platform, each python version.

Please let me know if there is a cleaner approach.

.github/workflows/pythonpackage.yml
```diff
@@ -8,37 +8,63 @@ on:
...
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         platform: [ubuntu-22.04, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
+      - name: No relevant changes, fast-path success
+        if: ${{ needs.detect_changes.outputs.relevant != 'true' }}
+        run: echo "No relevant file changes; skipping tests and linters."
+
```
Lee-W: nit: we should unify blank line usage. if we want it, then add it everywhere.

.github/workflows/pythonpackage.yml
```diff
@@ -8,37 +8,63 @@ on:
...
       - uses: actions/checkout@v6
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
         with:
           fetch-depth: 0
       - name: Set up Python ${{ matrix.python-version }}
         uses: astral-sh/setup-uv@v7
+        if: ${{ needs.detect_changes.outputs.relevant == 'true' }}
```
Lee-W, line 45: Do we really need it? I thought these steps are run sequentially

bearomorphism: I also asked Cursor whether I can remove it, but it said no :(

.github/workflows/pythonpackage.yml
bearomorphism: Let me make it more consistent.

.github/workflows/pythonpackage.yml
Lee-W: ok

PR #1745, opened 2025-12-29
docs(contributing): add poetry and lychee tips

bearomorphism:
Updated `contributing_tldr.md`.

### DIFF
--- a/docs/contributing_tldr.md
+++ b/docs/contributing_tldr.md
@@ -12,6 +12,9 @@ Please check the [pyproject.toml](https://github.com/commitizen-tools/commitizen
 ### Code Changes
 
 ```bash
+# Make sure you have the latest version of poetry installed
+poetry self update
+
 # Ensure you have the correct dependencies, for nix user's see below
 poetry install
 
@@ -34,6 +37,13 @@ pytest -n auto <test_suite>
 ```bash
 # Build the documentation locally and check for broken links
 poetry doc
+```
+
+Also, we use [Lychee](https://lychee.cli.rs/) to check for broken links in the documentation.
+
+```bash
+# Check for broken links in the documentation
+lychee .
 ```
 
 ### Nix Users


docs/contributing_tldr.md
```diff
@@ -12,6 +12,9 @@ Please check the [pyproject.toml](https://github.com/commitizen-tools/commitizen
 ### Code Changes
 
 ```bash
+# Make sure you have the latest version of poetry installed
+poetry self update
```
noirbizarre, line 16: This really depends on how poetry is installed. It will not work with system installation, mise or asdf based installation...
So not sure we should include this instruction. 
However, instead of forcing an update which might not be required, saying that "we require `poetry>=X.Y.Z`, you can verify with `poetry --version`" might be a better option

docs/contributing_tldr.md
```diff
@@ -36,6 +39,13 @@ pytest -n auto <test_suite>
 poetry doc
 ```
 
+Also, we use [Lychee](https://lychee.cli.rs/) to check for broken links in the documentation.
+
+```bash
+# Check for broken links in the documentation
+lychee .
+```
```
noirbizarre, line 47: If this is a TLDR, broken link checking, with a tool not installed with the other dependencies but run on CI, may not be worth adding there. Breaking links while contributing is rare enough to just let it fail on CI if it happends, WDYT ?

docs/contributing_tldr.md
woile: agreed

docs/contributing_tldr.md
woile: agreed, also it's not nice to drop this responsibility on other. I think is better the PR runs, reports and someone can take action if they want

bearomorphism: Agree

PR #1751, opened 2025-12-30
ci: use more workers to speed up tests in pipeline

bearomorphism:
From 3 workers to `auto` workers.


noirbizarre (approved):
Let's wait for #1752 to be merged to `master` build is fixed, then LGTM 🚀 (let's benefit from all those cores we have now).

PR #1756, opened 2026-01-03
docs: update contribution guide related to uv

bearomorphism:
It seems that `uv sync --dev` is not enough. I need to run `uv sync --all-groups` to have correct environment on my machine.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -40,7 +40,7 @@ If you're a first-time contributor, please check out issues labeled [good first 
     ```
 4. Set up the development environment:
     ```bash
-    uv sync --dev
+    uv sync --all-groups
     ```
 5. Set up pre-commit hooks:
     ```bash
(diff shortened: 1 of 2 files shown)


Lee-W:
> It seems that uv sync --dev is not enough. I need to run uv sync --all-groups to have correct environment on my machine.

Could you please elaborate on what you want to achieve?

bearomorphism:
`uv sync --dev` did not install all required packages for development like `ruff` in my environment. I spent some time to figure out what was wrong.

I'm not very familiar with `uv`.

Lee-W:
It's actually part of PEP 735.

https://github.com/commitizen-tools/commitizen/blob/2e4f4b26b5b0a0c0ebc660c1f78aca1e28d912bd/pyproject.toml#L89-L110

What we should use is `uv sync --group dev --group test --group linters` instead.

or we can use https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-group-include syntax. something like `dev {include-group="test"}, {icnlude-group="linters"}`

either ways work better than all-group IMO

bearomorphism:
I didn't know those. Will take a look.

noirbizarre:
Oh sorry, I hadn't read this one yet, so I did it in #1763.
So maybe just add the `--frozen` parameter to avoid dependencies being updated when syncing for contribution.

bearomorphism:
Thanks!

Issue #1702, opened 2025-12-08
Discuss what we should do if there are multiple `cz.*` files


bearomorphism:
## Action Items

- Survey how other tools with multiple configuration files work when they detect multiple configuration files.
  - For example, the tool `prettier` can be configured by `.prettier.rc`, `.prettier.json`, etc. (if I remembered it correctly)
- Decide next steps based on the survey and discussion.

## Reference

I wonder should we throw a warning if there are multiple `cz.*` file. Not sure how other tools like `eslint` or `prettier` handles multiple configuration file edge case.

_Originally posted by @bearomorphism in https://github.com/commitizen-tools/commitizen/pull/1657#discussion_r2573482065_

bearomorphism:
Pasting the response from ChatGPT. Haven't validated yet.

---

They both use a *“find nearest, then merge with clear precedence”* strategy, but with some differences. I’ll break it down and then extract general design lessons you can copy for your own tool.

---

## 1. ESLint: how config resolution works

### 1.1 Where ESLint looks for config

ESLint looks for config starting from the directory of the file being linted and walking upward, in roughly this order:

* `eslint.config.js` / `eslint.config.mjs` / `eslint.config.cjs` (new “flat config” system)
* Or older-style “eslintrc” files:

  * `.eslintrc.js`
  * `.eslintrc.cjs`
  * `.eslintrc.yaml` / `.eslintrc.yml`
  * `.eslintrc.json`
  * `package.json` → `"eslintConfig"` field

It stops when:

* It hits a config with `"root": true` (in old config), or
* It reaches the filesystem root.

### 1.2 Multiple configs: how they’re merged

In the **old eslintrc system** (still widely used):

* ESLint *merges* configs up the directory tree.
* The config closest to the file has highest precedence.
* Parents contribute defaults; children override.

Rough rules:

1. **Scalars and simple keys** (e.g. `parser`, single rule values)
   → “nearest wins”. Child config overrides parent.

2. **Objects like `rules`**
   → Keys are merged, with child keys overriding parent rules:

   ```jsonc
   // parent
   { "rules": { "eqeqeq": "error", "no-console": "warn" } }

   // child
   { "rules": { "no-console": "off" } }

   // effective
   { "rules": { "eqeqeq": "error", "no-console": "off" } }
   ```

3. **Arrays like `extends`**
   → They’re effectively concatenated, but the later/nearest config has precedence for overlapping rules (since the presets are processed in order and later ones override earlier).

[...]

PR #1773, opened 2026-01-05
fix(config): add warning for multiple configuration files and update documentation

### LINKED ISSUE #1771: Add warning if more than one configuration files are found and identify which one we're using

nicoleman0:
## Description

This PR implements a warning system that detects when multiple configuration files exist in the project directory (excluding `pyproject.toml`) and displays a warning message identifying which file is being used.

The implementation follows the consensus from issue #1702 where maintainers agreed that "we should have one and only one configuration file" and decided to add a warning when multiple config files are detected.

## Checklist

### Code Changes

### Documentation Changes

## Expected Behavior

When multiple configuration files are detected (e.g., `.cz.toml` and `.cz.json`), commitizen will display a warning message like:

`Multiple config files detected: .cz.toml, .cz.json. Using config file: '.cz.toml'.`

- The warning only appears when multiple config files exist (excluding `pyproject.toml`)
- The warning lists all detected config files in the order they would be checked
- The warning clearly indicates which file will be used (the first one in the priority order)
- If a config file is explicitly specified via the `--config` flag, no warning is displayed
- If only one config file exists (or only `pyproject.toml` plus one other), no warning is displayed
- The file priority order follows the standard commitizen config file precedence.

---

Fixes #1771

### DIFF
--- a/docs/config/configuration_file.md
+++ b/docs/config/configuration_file.md
@@ -20,6 +20,14 @@ Configuration files are typically located in the root of your project directory.
 
 The first valid configuration file found will be used. If no configuration file is found, Commitizen will use its default settings.
 
+!!! note
+    Commitizen supports explicitly specifying a configuration file using the `--config` option, which is useful when the configuration file is not located in the project root directory.
+    When `--config` is provided, Commitizen will only load configuration from the specified file and will not search for configuration files using the default search order described above. If the specified configuration file does not exist, Commitizen raises the `ConfigFileNotFound` error. If the specified configuration file exists but is empty, Commitizen raises the `ConfigFileIsEmpty` error.
+
+    ```bash
+    cz --config <PATH> <command>
+    ```
+
 !!! tip
     For Python projects, it's recommended to add your Commitizen configuration to `pyproject.toml` to keep all project configuration in one place.
 


commitizen/config/__init__.py
```diff
@@ -34,7 +34,53 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+        existing_files = [
+            filename
+            for filename in defaults.CONFIG_FILES
+            if filename != "pyproject.toml" and (path / filename).exists()
+        ]
+
+        # If more than one config file exists, warn the user
+        if len(existing_files) > 1:
+            # Find which one will be used (first non-empty one in the priority order)
+            used_config = None
```
woile: I don't think we should add the extra steps. 
It should just warn that multiple files were detected.
And alternatively, because the list is already in the order that cz is going to pick it put, you could just get the file that it's going to be used by getting the first element of `existing_files`

nicoleman0: Ok, yes that makes sense - I've simplified the implementation:

- Removed the extra steps that were checking if each config file is empty
- Now just warns about multiple files and uses `existing_files[0]` to show which file will be used

Changes have been pushed. Let me know if you need anything else.

commitizen/config/__init__.py
```diff
@@ -34,7 +34,37 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+def _check_and_warn_multiple_configs(filepath: str | None = None) -> None:
+    """Check if multiple config files exist and warn the user."""
+    if filepath is not None:
+        # If user explicitly specified a config file, no need to warn
+        return
+
+    git_project_root = git.find_git_project_root()
+    cfg_search_paths = [Path(".")]
+    if git_project_root:
+        cfg_search_paths.append(git_project_root)
```
Lee-W:
```suggestion
    cfg_search_paths = [Path(".")]
    if (git_project_root := git.find_git_project_root()):
        cfg_search_paths.append(git_project_root)
```

tests/test_conf.py
```diff
@@ -239,6 +239,90 @@ def test_load_empty_pyproject_toml_from_config_argument(_, tmpdir):
...
+            captured = capsys.readouterr()
+            assert "Multiple config files detected" in captured.err
+            assert ".cz.toml" in captured.err
+            assert ".cz.json" in captured.err
+            assert "Using" in captured.err
+
+            # Verify the correct config is loaded (first in priority order)
+            assert cfg.settings == _settings
+
+    def test_warn_multiple_config_files_with_pyproject(_, tmpdir, capsys):
```
Lee-W, line 262: But technically, it's multiple configs. We probably should handle it better 🤔

commitizen/config/__init__.py
```diff
@@ -34,7 +34,37 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
                 yield out_path
 
 
+def _check_and_warn_multiple_configs(filepath: str | None = None) -> None:
+    """Check if multiple config files exist and warn the user."""
+    if filepath is not None:
```
bearomorphism: The parameter `filepath` is unnecessary, and the function is very similar to `_resolve_config_paths`. I believe it's not necessary to introduce this function.

commitizen/config/__init__.py
```diff
@@ -34,7 +34,37 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+        if len(existing_files) > 1:
+            out.warn(
+                f"Multiple config files detected: {', '.join(existing_files)}. "
+                f"Using {existing_files[0]}."
+            )
+            break
+
+
 def read_cfg(filepath: str | None = None) -> BaseConfig:
+    _check_and_warn_multiple_configs(filepath)
```
bearomorphism:
```suggestion
    if filepath is None:
        _check_and_warn_multiple_configs()
```

commitizen/config/__init__.py
```diff
@@ -34,7 +34,37 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+
     for filename in _resolve_config_paths(filepath):
```
bearomorphism: I think a cleaner approach is to evaluate `_resolve_config_paths` first, exclude `pyproject.toml`, and check if there are multiple config files.

```python
config_candidates = list(_resolve_config_paths(filepath))
config_candidates_exclude_pyproject = ...
if len(config_candidates_exclude_pyproject) > 1:
    out.warn(...)
```

bearomorphism (changes requested):
Great but need some refactoring

bearomorphism:
Could you add the expected behavior in the PR description? Thanks!

commitizen/config/__init__.py
```diff
@@ -34,7 +34,37 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+            filename
+            for filename in defaults.CONFIG_FILES
+            if filename != "pyproject.toml" and (path / filename).exists()
+        ]
+
+        # If more than one config file exists, warn the user
+        if len(existing_files) > 1:
+            out.warn(
+                f"Multiple config files detected: {', '.join(existing_files)}. "
+                f"Using {existing_files[0]}."
```
bearomorphism:
```suggestion
                f"Using config file: '{existing_files[0]}'."
```

commitizen/config/__init__.py
```diff
@@ -34,7 +34,37 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+    if filepath is not None:
+        # If user explicitly specified a config file, no need to warn
+        return
+
+    git_project_root = git.find_git_project_root()
+    cfg_search_paths = [Path(".")]
+    if git_project_root:
+        cfg_search_paths.append(git_project_root)
+
+    for path in cfg_search_paths:
```
bearomorphism: The logic might be incorrect. You should perform the check after resolving all config file candidates, not for each path in `cfg_search_paths`.

commitizen/config/__init__.py
bearomorphism: And maybe consider add the following to the warning: `To avoid confusion, ensure you have only one Commitizen configuration file in your project.`

nicoleman0:
> Could you add the expected behavior in the PR description? Thanks!

I just updated the PR with the expected behavior (sorry about that, completely forgot to add that part), and also refactored my code with the suggestions that you all gave. 

If I missed anything, let me know again.

commitizen/config/__init__.py
```diff
@@ -35,7 +35,21 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
 
 
 def read_cfg(filepath: str | None = None) -> BaseConfig:
-    for filename in _resolve_config_paths(filepath):
+    config_candidates = list(_resolve_config_paths(filepath))
+
+    # Check for multiple config files and warn the user
+    if filepath is None:
```
bearomorphism: I am not sure whether `if filepath is None:` is still needed. We can perform the check for both cases anyway.

commitizen/config/__init__.py
```diff
@@ -35,7 +35,21 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+            path for path in config_candidates if path.name != "pyproject.toml"
+        ]
+        if len(config_candidates_exclude_pyproject) > 1:
+            filenames = [path.name for path in config_candidates_exclude_pyproject]
+            out.warn(
+                f"Multiple config files detected: {', '.join(filenames)}. "
+                f"Using config file: '{filenames[0]}'."
+            )
+
+    for filename in config_candidates:
```
bearomorphism, line 51: This for loop looks strange... I think we should remove the for loop

```python
using_config_file = config_candidates[0]
...
```

but that can be another PR

iirc the underlying logic of `BaseConfig` is a bit confusing

commitizen/config/__init__.py
```diff
@@ -35,7 +35,21 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+
+    # Check for multiple config files and warn the user
+    if filepath is None:
+        config_candidates_exclude_pyproject = [
+            path for path in config_candidates if path.name != "pyproject.toml"
+        ]
+        if len(config_candidates_exclude_pyproject) > 1:
+            filenames = [path.name for path in config_candidates_exclude_pyproject]
+            out.warn(
+                f"Multiple config files detected: {', '.join(filenames)}. "
```
bearomorphism: I am not entirely sure if it's good to just list the filenames but not the full path. There might be files with the same file names but different paths.

commitizen/config/__init__.py
```diff
@@ -35,7 +35,21 @@ def _resolve_config_paths(filepath: str | None = None) -> Generator[Path, None,
...
+                f"Using config file: '{filenames[0]}'."
```
bearomorphism: I am not sure if the used config file here will be the same as the returned config file around the line `if not conf.is_empty_config:`.

In other words, will the return value of this function always be `config_candidates[0]`?

@Lee-W @woile @noirbizarre wdyt

bearomorphism:
> > Could you add the expected behavior in the PR description? Thanks!
> 
> I just updated the PR with the expected behavior (sorry about that, completely forgot to add that part), and also refactored my code with the suggestions that you all gave.
> 
> If I missed anything, let me know again.

@nicoleman0 you're doing it great! But the pipeline is failing because some of the commit messages does not follow the requirements. Could you fix the commit messages? Thanks!

commitizen/config/__init__.py
nicoleman0: ah yes I see how it's redundant, I'll fix that!

nicoleman0:
> > > Could you add the expected behavior in the PR description? Thanks!
> > 
> > 
> > I just updated the PR with the expected behavior (sorry about that, completely forgot to add that part), and also refactored my code with the suggestions that you all gave.
> > If I missed anything, let me know again.
> 
> @nicoleman0 you're doing it great! But the pipeline is failing because some of the commit messages does not follow the requirements. Could you fix the commit messages? Thanks!

@bearomorphism I believe I've fixed the generic commit messages and I also refactored the `read_cfg` function. 

The redundant `if filepath is None:` check has been removed. The code now performs the multiple config file check in both cases, though it will only ever warn when filepath is `None`.

commitizen/config/__init__.py
nicoleman0: Ok, so should the code just use `config_candidates[0]` directly with the first file found, whether it's empty or not?

If so, the behavior would change to:
- When filepath is provided: Use that file, raise `ConfigFileIsEmpty` if empty (same as now)
- When auto-discovering: Use the first file found, return empty `BaseConfig()` if it's empty (currently skips empty files)

This would be simpler and align the warning message with actual behavior. The only downside is if someone has an empty .cz.toml and a valid .cz.json, it would now ignore the .cz.json (whereas currently it would use it).

Shall I go ahead with this simplification in this PR, or would you prefer a separate one?

bearomorphism: You could just add a TODO to remind the future contributors to address this issue. The main goal for this PR is just to add a warning for multiple configuration.

bearomorphism (approved):

tests/test_conf.py
bearomorphism: Agree. Maybe in a future test PR

Lee-W: No, I think it's not a test issue but a functionality issue instead

PR #1780, opened 2026-01-06
fix(git): commit bodies with carriage returns are correctly split by …

MattBelle:
... line

This should enable us to correctly parse squashed commits made on an SCM server residing on a Windows OS.

## Checklist

- [X] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/)

### Code Changes

- [X] Run `uv run poe all` locally to ensure this change passes linter check and tests
  - [X] Document any manual testing steps performed

### Documentation Changes

- [ ] Run `uv run poe doc` locally to ensure the documentation pages renders correctly

> When running `uv run poe doc`, any broken internal documentation links will be reported in the console output like this:
>
> ```text
> ```

## Expected Behavior

Commit bodies that use carriage returns for line breaks will now be successfully split when getting processed. In particular this enables commitizen to process squash commits made on an SCM server that is hosted by a Windows OS.

1. Make a dummy squash commit with carriage returns (`\r\n`):
```
ci: example ci change

fix: example fix change

feat: example feat change

additional context for example feat

test: example test change
```
2. Run `cz changelog --dry-run` and verify that only the `feat` and `fix` entries are included
3. Amend the previously made commit to use standard line feeds (`\n`)
4. Run `cz changelog --dry-run` and verify that only the `feat` and `fix` entries are included


tests/test_git.py
```diff
@@ -395,18 +395,21 @@ def test_get_filenames_in_commit_error(util: UtilFixture):
...
     # Test data with all fields populated
+    body = body_newline.join(
+        ["This is a detailed description", "of the new feature", "with multiple lines"]
+    )
     rev_and_commit = (
         "abc123\n"  # rev
         "def456 ghi789\n"  # parents
         "feat: add new feature\n"  # title
         "John Doe\n"  # author
         "john@example.com\n"  # author_email
```
bearomorphism: qq: is it possible that rev, parents, etc. end with "\r\n"?

If not, could you add comments here about why there will be possibly both CRLF and LF in the `rev_and_commit string?

commitizen/git.py
```diff
@@ -118,8 +118,8 @@ def from_rev_and_commit(cls, rev_and_commit: str) -> GitCommit:
             >>> commit.parents
             ['def456', 'ghi789']
         """
-        rev, parents, title, author, author_email, *body_list = rev_and_commit.split(
-            "\n"
+        rev, parents, title, author, author_email, *body_list = (
+            rev_and_commit.splitlines()
         )
```
bearomorphism, line 123: Should we split `rev, parents, title, author, author_email` with `"\n"` only?

tests/test_git.py
MattBelle: Good catch! While I cannot say definitively, my SCM server only appears to be attaching CRLFs on the body not the metadata. The metadata still appears to be using normal LFs.

commitizen/git.py
MattBelle: As mentioned in [my above comment](https://github.com/commitizen-tools/commitizen/pull/1780#discussion_r2669110270) I think that would technically work. My preference is we stick with my provided implementation (everything uses `splitlines()`)

- it still handles `\n`
- is easier to read
- splitting the output by line seems to be the intended goal of that line

Furthermore, I think there is minimal risk of it causing a bug since the parts that theoretically don't need the extra delimiter logic are all computer generated. That said, I'd be happy to piece out the logic and only apply the `splitlines()` call to the body if that's what you want.

tests/test_git.py
bearomorphism: I prefer `body_newline.join( original lines ov rev_and_commit )`.

The inconsistency between this test and the implementation may cause confusion to future contributors.

bearomorphism: Or you can add test cases when LF and CRLF are mixed in message.

Also, it would be better to add context on why we are covering those test cases. For example:

```python
# This should enable us to correctly parse squashed commits made on an SCM server residing on a Windows OS.
```

MattBelle: I think my latest commit matches your request? I made the test consistent with the implementation.  Let me know if I am wrong and misinterpreted you.

MattBelle:
> LGTM, thanks! @woile @Lee-W @noirbizarre could you also take a look?

Hey @bearomorphism, is this a blocker to merge or just a preference? While I technically have the ability to merge, I don't want to be rude and move forward with that before you are comfortable with it. That said, this would also fix a bug in a release pipeline so I would definitely appreciate some expediency.

PR #1781, opened 2026-01-07
docs: add AGENTS.md

bearomorphism:
Add AGENTS.md to our project. At least this is a start to improve those AI assisted PRs.

## Reference

https://agents.md/

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,54 @@
+# AGENTS instructions
+
+## Purpose
+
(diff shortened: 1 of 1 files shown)


Lee-W (approved):
probably could also mention that we're using conventional commit

bearomorphism:
> probably could also mention that we're using conventional commit

We probably have mentioned that in the contribution guidelines, which alrealy include it

PR #1784, opened 2026-01-12
fix(config): ensure the actually used config file is correct, fix single config file being detected as duplicated, better test coverage

### LINKED ISSUE #1787: Single configuration file detected as duplicate since 4.11.2
### Description

Apparently #1773 introduced a bug.
When running in a git repository with only a single `.cz.toml` file at the root, it is detected as a duplicate:

### Steps to reproduce

In an empty directory run
```
pip install -U commitizen==4.11.2
touch .cz.toml 
git init .
cz version
```

Outputs the wrong error message
```
Multiple config files detected: .cz.toml, .cz.toml. Using config file: '.cz.toml'.
4.11.2
```

### Current behavior

One file is detected twice.

### Desired behavior

File should only be detected once.

### Screenshots

_No response_

### Environment

Commitizen Version: 4.11.2
Python Version: 3.14.2 (main, Jan  7 2026, 12:38:13) [Clang 17.0.0 (clang-1700.6.3.2)]
Operating System: Darwin

bearomorphism:
This is ~~an improvement~~ a fix based on #1773 

- Improved test coverage of multiple config file warning
- Ensured the consistency of actually used config file, which is not always the first element of the candidates

Closes #1787

### DIFF
--- a/docs/config/configuration_file.md
+++ b/docs/config/configuration_file.md
@@ -10,13 +10,15 @@ It is recommended to create a configuration file via our [`cz init`](../commands
 
 Configuration files are typically located in the root of your project directory. Commitizen searches for configuration files in the following order:
 
-1. `pyproject.toml` (in the `[tool.commitizen]` section)
-2. `.cz.toml`
+<!-- DEPENDENCY: commitizen/defaults.py CONFIG_FILES -->
+
+1. `.cz.toml`
+2. `cz.toml`
 3. `.cz.json`
 4. `cz.json`
 5. `.cz.yaml`
 6. `cz.yaml`
-7. `cz.toml`
+7. `pyproject.toml` (in the `[tool.commitizen]` section)
 
 The first valid configuration file found will be used. If no configuration file is found, Commitizen will use its default settings.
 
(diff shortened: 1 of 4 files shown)


tests/test_conf.py
```diff
@@ -239,71 +239,92 @@ def test_load_empty_pyproject_toml_from_config_argument(_, tmpdir):
...
+                ).replace('version = "1.0.0"', 'version = "2.0.0"')
+            subdir.join(config_file).write(subdir_content)
+
+            with subdir.as_cwd():
+                cfg = config.read_cfg()
+                captured = capsys.readouterr()
+
+                assert "Multiple config files detected" in captured.err
+                assert f"Using config file: '{config_file}'" in captured.err
+                assert cfg.path == Path(config_file)
```
bearomorphism, line 319: should also cover no warn case

ref #1787

Lee-W:
> In case your uv project doesn’t have a lockfile or any uv specific config, you can still force poe to use the uv executor like so:

Yep, agree this might not be necessary. No strong opinion on this one. Will leave it to @noirbizarre

bearomorphism:
> > In case your uv project doesn’t have a lockfile or any uv specific config, you can still force poe to use the uv executor like so:
> 
> Yep, agree this might not be necessary. No strong opinion on this one. Will leave it to @noirbizarre

Are you replying to #1795 🤔

docs/config/configuration_file.md
```diff
@@ -28,9 +30,6 @@ The first valid configuration file found will be used. If no configuration file
     cz --config <PATH> <command>
     ```
 
-!!! tip
-    For Python projects, it's recommended to add your Commitizen configuration to `pyproject.toml` to keep all project configuration in one place.
-
```
bearomorphism, line 33: I'm not sure if we should recommend `pyproject.toml` or `.cz.toml`

woile: No recommendation, it's just a flavor. I imagine a person that works with multiple languages, may prefer to have a `.cz.toml` and a pure python person may prefer to have everything in `pyproject.toml`

bearomorphism: Let me revert this part and address this comment in another doc PR

PR #1794, opened 2026-01-13
build: switch configuration file to .cz.toml

bearomorphism:
Just validated that `conf.settings` is the same before and after the change.

## Context

https://github.com/commitizen-tools/commitizen/issues/1702#issuecomment-3711298396


Lee-W:
I actually prefer it to be in the `pyproject.toml` file.

PR #1795, opened 2026-01-13
build: remove unnecessary poe config in pyproject.toml

bearomorphism:


noirbizarre:
Can you provide details? 
Because it's [the official documented way,](https://poethepoet.natn.io/guides/without_poetry.html#usage-with-uv) it works in CI and it works locally.

bearomorphism:
Maybe I need to doubly check again 
both my IDE and ChatGPT said it's invalid

bearomorphism:
My IDE shows this:

image


> Can you provide details? Because it's [the official documented way,](https://poethepoet.natn.io/guides/without_poetry.html#usage-with-uv) it works in CI and it works locally.

That is interesting. In the official documentation you provided, it also says

> In case your uv project doesn’t have a lockfile or any uv specific config, you can still force poe to use the uv executor like so:

so we probably don't need the executor config.

Also, I noticed there is a discrepancy in poethepoet's documentation. The following snippet is newer than the one you provided:

https://github.com/nat-n/poethepoet/blame/6b7691f3949024e01900df1be5ccc60eefb7c8af/docs/tasks/options.rst#L208

Lee-W:
I'm ok with either way if both work. I'll leave it to @noirbizarre

noirbizarre:
Actually, I don't really rely on the executor as I run `poe` directly from the venv (using `mise` so the venv is automatically activated). So I am fine with the removal, it was more for contributors, just to be sure it still works for them.

If we keep the `uv` executor, then I propose we add `no_sync` and `frozen` options to avoid syncing and touching the lock (cf. https://poethepoet.natn.io/global_options.html#uv-executor).

It's up to you @bearomorphism

bearomorphism:
Let me check the documentation. I am currently not familiar with `uv` and the executor stuffs enough.

noirbizarre:
After rereading [the doc](https://poethepoet.natn.io/global_options.html#configure-the-executor), I think that if we remove it, it will be `auto` and then automatically pick `uv` as executor. So maybe keep it to be explicit but add the 2 options to avoid unexpected lock update

bearomorphism:
> My IDE shows this:
> 
> image
> > Can you provide details? Because it's [the official documented way,](https://poethepoet.natn.io/guides/without_poetry.html#usage-with-uv) it works in CI and it works locally.
> 
> That is interesting. In the official documentation you provided, it also says
> 
> > In case your uv project doesn’t have a lockfile or any uv specific config, you can still force poe to use the uv executor like so:
> 
> so we probably don't need the executor config.
> 
> Also, I noticed there is a discrepancy in poethepoet's documentation. The following snippet is newer than the one you provided:
> 
> https://github.com/nat-n/poethepoet/blame/6b7691f3949024e01900df1be5ccc60eefb7c8af/docs/tasks/options.rst#L208

I suspect this is an issue from outdated json schema: https://www.schemastore.org/partial-poe.json

bearomorphism:
The json schema looks correct. Maybe it's my IDE's issue...

bearomorphism:
Just applied @noirbizarre 's suggestion. There is still error in my IDE but the terminal output looks good (it didn't complain)

image

PR #1800, opened 2026-01-14
test: remove unreachable code in pytest.raises block, fix some malformed tests, extract fixtures

bearomorphism:
# Why

`with pytest.raises` block should contain only one statement which we expect it to throw exceptions.

# Changes made

- Removed unreachable code in tests
- Fixed `test_check_command_cli_overrides_config_message_length_limit`. The last `check_cmd` assignment had no effect.
- Extracted `success_mock` and `commit_mock`

### DIFF
--- a/tests/test_cz_customize.py
+++ b/tests/test_cz_customize.py
@@ -376,9 +376,9 @@ def config_with_unicode(request):
 
 
 def test_initialize_cz_customize_failed():
+    config = BaseConfig()
     with pytest.raises(MissingCzCustomizeConfigError) as excinfo:
-        config = BaseConfig()
-        _ = CustomizeCommitsCz(config)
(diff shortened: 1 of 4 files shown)


tests/test_cz_customize.py
```diff
@@ -376,8 +376,8 @@ def config_with_unicode(request):
 
 
 def test_initialize_cz_customize_failed():
+    config = BaseConfig()
     with pytest.raises(MissingCzCustomizeConfigError) as excinfo:
-        config = BaseConfig()
         _ = CustomizeCommitsCz(config)
```
Lee-W:
```suggestion
        CustomizeCommitsCz(config)
```

bearomorphism: thanks

PR #1803, opened 2026-01-15
fix(config): include pyproject.toml in multi config file warning

bearomorphism:
I just remembered that we already have a feature that detects if a configuration file is valid. We can include pyproject.toml in the multi config warning.

Relate #1702 

Also added a test to ensure that no warning when there are both pyproject.toml (without cz settings section) and cz.toml

### DIFF
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -14,7 +14,7 @@ from commitizen.config.toml_config import TomlConfig
 from commitizen.config.yaml_config import YAMLConfig
 from commitizen.exceptions import ConfigFileIsEmpty, InvalidConfigurationError
 
-PYPROJECT = """
+TOML_STR = """
 [tool.commitizen]
 name = "cz_jira"
 version = "1.0.0"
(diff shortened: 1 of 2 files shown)


tests/test_conf.py
```diff
@@ -30,11 +30,17 @@
     "scripts/generate_documentation.sh"
 ]
 post_bump_hooks = ["scripts/slack_notification.sh"]
+"""
 
+PYPROJECT = (
+    TOML_STR
+    + """
```
Lee-W:
```suggestion
PYPROJECT = (
f"""
{TOML_STR}
```

tests/test_conf.py
```diff
@@ -368,6 +365,18 @@ def test_no_warn_with_single_config_file(
             assert "Multiple config files detected" not in captured.err
             assert cfg.path == Path(config_file)
 
+    def test_no_warn_with_no_commitizen_section_in_pyproject_toml_and_cz_toml(
+        _, tmpdir, capsys
```
Lee-W:
```suggestion
        self, tmpdir, capsys
```

Lee-W (approved):
a few nits

PR #1805, opened 2026-01-15
test(version_schemes): replace match with strict string assertions

bearomorphism:
Make the test stricter

### DIFF
--- a/tests/test_version_schemes.py
+++ b/tests/test_version_schemes.py
@@ -61,5 +61,8 @@ def test_warn_if_version_protocol_not_implemented(
     ep.load.return_value = NotVersionProtocol
     mocker.patch.object(metadata, "entry_points", return_value=(ep,))
 
-    with pytest.warns(match="VersionProtocol"):
+    with pytest.warns() as warnings:
         get_version_scheme(config.settings, "any")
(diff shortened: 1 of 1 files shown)


tests/test_version_schemes.py
```diff
@@ -61,5 +61,8 @@ class NotVersionProtocol:
     ep.load.return_value = NotVersionProtocol
     mocker.patch.object(metadata, "entry_points", return_value=(ep,))
 
-    with pytest.warns(match="VersionProtocol"):
+    with pytest.warns() as warnings:
         get_version_scheme(config.settings, "any")
+    assert "Version scheme any does not implement the VersionProtocol" in str(
+        warnings[0].message
+    )
```
Lee-W, line 68: What about move this string to match?


```suggestion
    with pytest.warns(match="Version scheme any does not implement the VersionProtocol") as warnings:
get_version_scheme(config.settings, "any")
```

bearomorphism: but `match` parameter is for a regular expression test. I believe it would be better to explicitly use string assertion.

Lee-W: sounds good 👍

bearomorphism: thanks

PR #1625, opened 2025-09-20
feat: add parameter to changelog generation 

dornech:
to identify incremental generation run in the jinja tmeplate
This allows

PR #1808, opened 2026-01-18
fix(changelog): add incremental parameter to changelog generation

bearomorphism:
Relate #1620, #1625

### DIFF
--- a/commitizen/commands/changelog.py
+++ b/commitizen/commands/changelog.py
@@ -266,6 +266,7 @@ class Changelog:
             self.cz.template_loader,
             self.template,
             **{
+                "incremental": self.incremental,  # extra variable for the template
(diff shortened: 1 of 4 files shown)


commitizen/commands/changelog.py
```diff
@@ -267,6 +267,7 @@ def __call__(self) -> None:
             tree,
             self.cz.template_loader,
             self.template,
+            incremental=self.incremental,  # extra variable for the template
```
Lee-W: I think we should make it part of kwargs instead of making it here

bearomorphism: I don't get it. Isn't it already part of kwargs?

Lee-W: I mane make it part of the dict below

PR #1809, opened 2026-01-18
test: remove temp variable and extract fixtures to shorten tests

bearomorphism:
As title. Parameterized some tests, and extracted some duplicated patterns to fixtures.


tests/commands/test_changelog_command.py
```diff
@@ -30,6 +30,21 @@
     from tests.utils import UtilFixture
 
 
+@pytest.fixture(scope="function")
```
Lee-W: nit: there's a ruff rule suggesting we not add `(scope="function")` as it's the default. no strong opinion. but we should either all add it or all remove

tests/commands/test_changelog_command.py
```diff
@@ -30,6 +30,21 @@
...
+def changelog_jinja_file(tmp_project_root: Path):
```
Lee-W:
```suggestion
def changelog_jinja_file(tmp_project_root: Path) -> Path:
```

tests/commands/test_changelog_command.py
```diff
@@ -30,6 +30,21 @@
...
+    return tmp_project_root / "changelog.jinja"
+
+
+@pytest.fixture(scope="function")
+def changelog_tpl(tmp_project_root: Path, any_changelog_format: ChangelogFormat):
```
Lee-W:
```suggestion
def changelog_tpl(tmp_project_root: Path, any_changelog_format: ChangelogFormat) -> Path:
```

tests/commands/test_changelog_command.py
```diff
@@ -30,6 +30,21 @@
...
+    return tmp_project_root / any_changelog_format.template
+
+
+@pytest.fixture(scope="function")
+def changelog_file(tmp_project_root: Path, any_changelog_format: ChangelogFormat):
```
Lee-W:
```suggestion
def changelog_file(tmp_project_root: Path, any_changelog_format: ChangelogFormat) -> Path:
```

tests/conftest.py
```diff
@@ -88,6 +88,16 @@ def tmp_commitizen_project(tmp_git_project):
     yield tmp_git_project
 
 
+@pytest.fixture(scope="function")
+def tmp_project_root(tmp_commitizen_project):
```
Lee-W:
```suggestion
def tmp_project_root(tmp_commitizen_project) -> Path:
```

tests/conftest.py
```diff
@@ -88,6 +88,16 @@ def tmp_commitizen_project(tmp_git_project):
...
+    return Path(tmp_commitizen_project)
+
+
+@pytest.fixture(scope="function")
+def pyproject(tmp_project_root: Path):
```
Lee-W:
```suggestion
def pyproject(tmp_project_root: Path) -> Path:
```

tests/commands/test_bump_command.py
```diff
@@ -53,62 +53,63 @@
...
-    assert tag_exists is True and "tag:refs/tags/0.2.0\n" in cmd_res.out
+    assert git.tag_exist("0.2.0")
+    assert (
+        "tag:refs/tags/0.2.0"
+        in cmd.run('git for-each-ref refs/tags --format "%(objecttype):%(refname)"').out
+    )
 
-    _is_signed = git.is_signed_tag("0.2.0")
-    assert _is_signed is False
+    assert not git.is_signed_tag("0.2.0")
```
Lee-W:
```suggestion
    assert git.is_signed_tag("0.2.0") is False
```

Let's do this when a bool is encountered so that we won't accidentally get an empty list and not noticed

tests/commands/test_changelog_command.py
bearomorphism: Let me add the ruff rule in a separate future PR

tests/commands/test_bump_command.py
bearomorphism: fixed all `assert not` patterns I found

Issue #1672, opened 2025-11-22
Use `ChainMap` to rewrite the cmd options <- configuration file <- default configuration chain


bearomorphism:
### Description

As title. A `ChainMap` keeps more information and is easier to debug.

### Additional context

https://realpython.com/python-chainmap/#prioritizing-command-line-apps-settings

bearomorphism:
This can be assigned to me, but probably I won't implement until other PRs are merged. There may be some dependencies.

bearomorphism:
I just created a poc #1812.

PR #1812, opened 2026-01-19
fix: try to use ChainMap to chain settings, fix `message_length_limit` inconsistency

bearomorphism:
Relate #1672

### DIFF
--- a/commitizen/config/settings.py
+++ b/commitizen/config/settings.py
@@ -1,0 +1,27 @@
+from collections import ChainMap
+from collections.abc import Mapping
+from typing import Any, cast
+
+from commitizen.defaults import DEFAULT_SETTINGS, Settings
+
+
+class ChainSettings:
+    def __init__(
+        self,
+        config_file_settings: Mapping[str, Any],
+        cli_settings: Mapping[str, Any] | None = None,
+    ) -> None:
+        if cli_settings is None:
+            cli_settings = {}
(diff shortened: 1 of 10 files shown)


commitizen/config/settings.py
```diff
@@ -0,0 +1,27 @@
+from collections import ChainMap
+from collections.abc import Mapping
+from typing import Any, cast
+
+from commitizen.defaults import DEFAULT_SETTINGS, Settings
+
+
+class ChainSettings:
```
woile, line 8: Add some docstrings here:

1. What is this
2. What it does
3. How to use
4. How it compares to the previous solution

That would really help me understand the benefit.

Something I'd like to know as well:
Right now, if I make a plugin, I can create custom settings, but they are not typed. Would this allow plugin to declare their own settings?
[example](https://github.com/kpn/cz-kpn/blob/main/src/cz_kpn/rules.py#L105)

bearomorphism: > Right now, if I make a plugin, I can create custom settings, but they are not typed. Would this allow plugin to declare their own settings?

Thanks, I will put those scenarios into consideration. `ChainMap` approach should not break the type system.

bearomorphism: > How it compares to the previous solution

We can discuss this in #1672. I will share the motivation there.

PR #1816, opened 2026-01-23
feat: add `postrelease` version

kamikazePT:
## Description
- added `postrelease` flag
- added missing docs for `devrelease` and added new docs for `postrelease`
- removed existing `TODO` on lack of support for `postrelease`
- add `prerelease_property` tests as well as the `postrelease_property` ones

## Checklist

### Code Changes

### Documentation Changes

## Expected Behavior
when using the new `postrelease` flag, a `postN` will be added before `devN`

```bash
cz bump --postrelease --yes
```

## Additional Context
over the last couple of years, professionally, our internal workflows have been using `commitizen,` in an older version fork, where we internally added support for a couple of missing features at the time, one was `postrelease`.


bearomorphism:
Nice!!! We will go through it hopefully by mid of February.

bearomorphism:
@Lee-W could you review this PR? I remember you are familiar with these

Lee-W:
This is a bit huge and will need more time to take a deeper look

woile:
Where is `postrelease` being discussed? do we have a ticket?

I'm not very happy with adding support for a post-releases. It adds a burden to the codebase, and it's actually not a good practice. `semver` doesn't have post-releases, and it discourages modifying versions already released. I subscribe to this idea. Specially if the user is already using `commitizen`, which makes it easy to release a new version 🤷🏻‍♂️ 

What do you guys think?

kamikazePT:
> Where is `postrelease` being discussed? do we have a ticket?
> 
> I'm not very happy with adding support for a post-releases. It adds a burden to the codebase, and it's actually not a good practice. `semver` doesn't have post-releases, and it discourages modifying versions already released. I subscribe to this idea. Specially if the user is already using `commitizen`, which makes it easy to release a new version 🤷🏻‍♂️
> 
> What do you guys think?

There was no ticket, it was only mentioned in the docs (https://commitizen-tools.github.io/commitizen/commands/bump/#development-releases) as currently not supported.

`postrelease` is indeed discouraged, even the use cases mentioned by [`pep440`](https://peps.python.org/pep-0440/#post-releases) are quite slim.

[`semver`](https://semver.org/#spec-item-9) in fact, has no post concept, a `post.N` is treated as a `prerelease`, which means that `1.0.0` is always **after** a `1.0.0.post.1` and not before.

We've been using `postrelease` in a standalone fork, as way to explicit tag some services on hotfixes.
`master` can be at `2.0.0` but `production` is still at `1.6.2`, once we detect an issue in prod, we do a quick `1.6.2-post.1` as a patch until we don't promote `2.0.0`. We keep stable versions only at `master`, hence why we avoid a hotfix with a `1.6.Y` patch and we don't 'cherry-pick' a `master` tag to push to `production hotfix`
☝️ is a custom flow and not standard, which might not be a strong enough argument to add this support...

Will wait on a follow-up from you guys 👍

PR #1826, opened 2026-01-31
fix: add pytest ruff rule PT and fix missing deprecation warning

bearomorphism:
There are still some ruff rules pending for enabling

The deleted test cases are duplicated

### DIFF
--- a/tests/test_bump_hooks.py
+++ b/tests/test_bump_hooks.py
@@ -38,5 +38,5 @@ def test_run_error(mocker: MockFixture):
 
 def test_format_env():
     result = hooks._format_env("TEST_", {"foo": "bar", "bar": "baz"})
-    assert "TEST_FOO" in result and result["TEST_FOO"] == "bar"
-    assert "TEST_BAR" in result and result["TEST_BAR"] == "baz"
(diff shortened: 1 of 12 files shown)


tests/test_bump_hooks.py
```diff
@@ -38,5 +38,7 @@ def test_run_error(mocker: MockFixture):
 
 def test_format_env():
     result = hooks._format_env("TEST_", {"foo": "bar", "bar": "baz"})
-    assert "TEST_FOO" in result and result["TEST_FOO"] == "bar"
-    assert "TEST_BAR" in result and result["TEST_BAR"] == "baz"
+    assert "TEST_FOO" in result
+    assert result["TEST_FOO"] == "bar"
+    assert "TEST_BAR" in result
+    assert result["TEST_BAR"] == "baz"
```
Lee-W:
```suggestion
    assert result["TEST_FOO"] == "bar"
    assert result["TEST_BAR"] == "baz"
```

I think this should be sufficient. The test fails with KeyError if key is not present anyway

Lee-W (approved):

PR #1825, opened 2026-01-31
feat(cli): add description when choosing a commit rule

namwoam:
Closes #1822 

## Description

This PR adds descriptions to the commit message rule selection step.

## Checklist

- [ ✅ ] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/contributing)

### Was generative AI tooling used to co-author this PR?

- [X] No

### Code Changes

- [] Add test cases to all the changes you introduce
- [✅ ] Run `uv run poe all` locally to ensure this change passes linter check and tests
- [ ] Manually test the changes:
  - [✅ ] Verify the feature/bug fix works as expected in real-world scenarios
  - [✅ ] Test edge cases and error conditions
  - [ ] Ensure backward compatibility is maintained
  - [ ] Document any manual testing steps performed
- [ ] Update the documentation for the changes

### Documentation Changes

- [ ] Run `uv run poe doc` locally to ensure the documentation pages renders correctly
- [ ] Check and fix any broken links (internal or external)

## Expected Behavior
When selecting the commit rule at `cz init`, it should appear a description when selecting a commit rule.

Before:
Screenshot 2026-01-31 at 11 08 56 AM

After:
Screenshot 2026-01-31 at 12 50 58 PM

## Additional Context

Issue: https://github.com/commitizen-tools/commitizen/issues/1822

PR #1829, opened 2026-01-31
test(init): cover cz without descriptions

bearomorphism:
Follow up of #1825 

Better test coverage for the change and renamed some variables and methods.

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -171,24 +171,31 @@ class Init:
     def _ask_name(self) -> str:
         name: str = questionary.select(
             f"Please choose a cz (commit rule): (default: {DEFAULT_SETTINGS['name']})",
-            choices=self._construct_name_choice_with_description(),
+            choices=self._construct_name_choices_from_registry(),
             default=DEFAULT_SETTINGS["name"],
             style=self.cz.style,
         ).unsafe_ask()
         return name
 
-    def _construct_name_choice_with_description(self) -> list[questionary.Choice]:
+    def _construct_name_choices_from_registry(self) -> list[questionary.Choice]:
+        """
+        Construct questionary choices of cz names from registry.
+        """
         choices = []
         for cz_name, cz_class in registry.items():
             try:
                 cz_obj = cz_class(self.config)
             except MissingCzCustomizeConfigError:
+                # Fallback if description is not available
                 choices.append(questionary.Choice(title=cz_name, value=cz_name))
                 continue
-            first_example = cz_obj.schema().partition("\n")[0]
+
+            # Get the first line of the schema as the description
+            # TODO(bearomorphism): schema is a workaround. Add a description method to the cz class.
+            description = cz_obj.schema().partition("\n")[0]
             choices.append(
                 questionary.Choice(
-                    title=cz_name, value=cz_name, description=first_example
+                    title=cz_name, value=cz_name, description=description
                 )
             )
         return choices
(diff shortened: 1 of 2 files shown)


commitizen/commands/init.py
```diff
@@ -171,24 +171,31 @@ def _ask_config_path(self) -> Path:
...
+    def _construct_name_choices_from_registry(self) -> list[questionary.Choice]:
+        """
+        Construct questionary choices of cz names from registry.
+        """
         choices = []
         for cz_name, cz_class in registry.items():
             try:
+                # TODO(bearomorphism): can we get the description from the cz class without initiating an instance?
                 cz_obj = cz_class(self.config)
             except MissingCzCustomizeConfigError:
```
bearomorphism, line 188: Actually, I'm not sure whether this is compatible with plugins. Should we use general exception?

wdyt @Lee-W

commitizen/commands/init.py
```diff
@@ -171,24 +171,31 @@ def _ask_config_path(self) -> Path:
...
 
-    def _construct_name_choice_with_description(self) -> list[questionary.Choice]:
+    def _construct_name_choices_from_registry(self) -> list[questionary.Choice]:
+        """
+        Construct questionary choices of cz names from registry.
+        """
         choices = []
         for cz_name, cz_class in registry.items():
             try:
+                # TODO(bearomorphism): can we get the description from the cz class without initiating an instance?
```
Lee-W: need to make it a class of static method, as we need to take config as an argument, it's not likely.

commitizen/commands/init.py
Lee-W: why do you think it might not be compatible

commitizen/commands/init.py
```diff
@@ -171,24 +171,31 @@ def _ask_config_path(self) -> Path:
...
                 cz_obj = cz_class(self.config)
             except MissingCzCustomizeConfigError:
+                # workaround for cz_customize
```
Lee-W: why is it a workaround for cz_customize? I thought it's for non-core `cz`s

commitizen/commands/init.py
```diff
@@ -171,24 +171,31 @@ def _ask_config_path(self) -> Path:
...
                 choices.append(questionary.Choice(title=cz_name, value=cz_name))
                 continue
-            first_example = cz_obj.schema().partition("\n")[0]
+
+            # Get the first line of the schema as the description
```
Lee-W, line 193: we'd better add a description in v5. `schema` looks like a workaround for me

commitizen/commands/init.py
bearomorphism: I thought other plugins throw other exceptions when they initialize with empty config, but we actually always initialize that with a `BaseConfig` object.

commitizen/commands/init.py
bearomorphism: ok

commitizen/commands/init.py
bearomorphism: do non-core `cz`s throw `MissingCzCustomizeConfigError`? I am not familiar with it

commitizen/commands/init.py
bearomorphism: I think it does not have to be a `v5` feature. We can make description optional.

commitizen/commands/init.py
bearomorphism: If I understood it correctly, only `CzCustomize` throws the error

commitizen/commands/init.py
bearomorphism: added a TODO

Issue #1831, opened 2026-01-31
Redesigning `cz init` `self.config`


bearomorphism:
### Description

Just noticed this.

I consider it strange that `self.config` is widely referenced by `Init` object. In normal scenarios, there shouldn't be configuration files when user runs `cz init`?

Maybe we should replace those `self.config` references by `DEFAULT_SETTINGS`. wdyt?

The only case `self.config` should be used is to check whether the project is already initialized by `cz`.

Lee-W:
1. Interface consistency https://github.com/commitizen-tools/commitizen/blob/0497ef20b8cbfb2f2bce865753f5861ad09333e4/commitizen/cli.py#L109-L178
2. https://github.com/commitizen-tools/commitizen/blob/0497ef20b8cbfb2f2bce865753f5861ad09333e4/commitizen/commands/init.py#L79-L82 we check whether there's already a config

woile:
A feature I would like to see, which would require a specification first:

How to extend the `init` in your own custom rules? For example, I maintain this convention, and if the user chooses it in the init, then it should ask about the settings provided by the conventions:
https://github.com/kpn/cz-kpn?tab=readme-ov-file#custom-configuration

So the convention, should be able to provide its own "options" if it makes sense

Lee-W:
like what @woile proposed

PR #1832, opened 2026-01-31
refactor(init): replace self.config.settings with DEFAULT_SETTINGS

bearomorphism:
Related to #1831 

In usual use cases, `self.config.settings` is a copy of `DEFAULT_SETTINGS`

This probably makes more sense. wdyt

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -105,7 +105,7 @@ class Init:
             version_provider = self._ask_version_provider()  # select
             tag = self._ask_tag()  # confirm & select
             version_scheme = self._ask_version_scheme()  # select
-            version = get_version_scheme(self.config.settings, version_scheme)(tag)
+            version = get_version_scheme(DEFAULT_SETTINGS, version_scheme)(tag)
             tag_format = self._ask_tag_format(tag)  # confirm & text
             update_changelog_on_bump = self._ask_update_changelog_on_bump()  # confirm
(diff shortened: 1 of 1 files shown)


Lee-W:
conflict. convert it to draft

Manny7717 (approved):
Verified behavior-preserving refactor (head 78956b5):

1. **`get_version_scheme(DEFAULT_SETTINGS, version_scheme)`** — `version_scheme` is always passed explicitly from `_ask_version_scheme()` (a `questionary.select` over KNOWN_SCHEMES), so the `settings` argument is only consulted for the deprecated `version_type` fallback when `name` is None — unreachable here. No behavior change.
2. **`encoding=DEFAULT_SETTINGS["encoding"]`** (write + read of `.pre-commit-config.yaml`) — `Init.__call__` early-returns when `self.config.path` exists (init.py:88-90), so the only reachable path has `config.settings` == a `DEFAULT_SETTINGS.copy()` (base_config.py:20, no config data loaded). Identical values, same behavior. This also removes the latent coupling to whatever a future config-loading path might inject.

- tests/commands/test_init_command.py: 30/30 pass on head.
- CI matrix green (python-check 3.10-3.14 × 3 OS, check-links, codecov).

Agreed this is the right direction for #1831. One non-blocking note for a follow-up: the remaining `self.config` uses in init.py (e.g. `self.config.settings` in `_get_config_data`'s fallback and `self.cz`) could get the same treatment once #1831's discussion settles on the full scope.

PR #1836, opened 2026-02-01
test: rename fixture config to default_config to avoid name conflict

bearomorphism:


tests/commands/conftest.py
bearomorphism, line 1: duplicated fixtures

bearomorphism:
Thanks for explanation, I didn't know the philosophy details behind pytest fixture.

The main motivation of this PR was that we have a module called `config` and a fixture also with that name. It sometimes confuses me. I guess there is a better solution like `import config as cf` in our tests.

Now I agree that we should keep the name unchanged.

bearomorphism:
I will close this PR and make create another one for removing duplicated fixtures.

PR #1839, opened 2026-02-01
refactor(tags): extract version resolution method

bearomorphism:


commitizen/tags.py
```diff
@@ -284,3 +268,16 @@ def from_settings(cls, settings: Settings) -> Self:
             ignored_tag_formats=settings["ignored_tag_formats"],
             merge_prereleases=settings["changelog_merge_prerelease"],
         )
+
+    def _extract_version_parts(self, match: re.Match[str]) -> str:
```
bearomorphism: maybe rename to sth like `_resolve_version_str`

noirbizarre: I would keep the name as is, I don't really like naming redundant with typing. If you want to change the name, just `_extract_version` or `_resolve_version` is enough IMHO (with a little preference for `_extract_version` which is semantically closer to reality than `_resolve_version` which can have another meaning)

bearomorphism: Yes `_extract_version` should be better. I think `_part` is a bit confusing.

PR #1840, opened 2026-02-01
refactor: simplify code with pathlib Path object

bearomorphism:


bearomorphism:
> I was intending to do the same and also switch from using all the legacy pytest path to pathlib too (pathlib everywhere would help a lot)

Sounds good. This can be a good first issue.

PR #1842, opened 2026-02-02
refactor(config): replace is_empty_config with contains_commitizen_section, improve multi config resolution algorithm

bearomorphism:
- Fixed incorrectly used `samefile`
- Removed unnecessary same file check in `_resolve_config_candidates`. The tests already covered the cases "same git repo root and cwd" so it should not be broken after merging this PR.
- Separate "empty configuration" detection and `_parse_settings`

## Why removing `is_empty_config`?

Hmm, I believe a stateless function is easier to read in this case.

### DIFF
--- a/commitizen/config/base_config.py
+++ b/commitizen/config/base_config.py
@@ -17,9 +17,15 @@ if TYPE_CHECKING:
 
 class BaseConfig:
     def __init__(self) -> None:
-        self.is_empty_config = False
         self._settings: Settings = DEFAULT_SETTINGS.copy()
         self._path: Path | None = None
+
+    def contains_commitizen_section(self) -> bool:
+        """Check if the config file contains a commitizen section.
+
+        The implementation is different for each config file type.
+        """
+        raise NotImplementedError()
 
     @property
     def settings(self) -> Settings:
--- a/commitizen/config/toml_config.py
+++ b/commitizen/config/toml_config.py
@@ -26,6 +26,11 @@ class TomlConfig(BaseConfig):
         self.path = path
         self._parse_setting(data)
 
+    def contains_commitizen_section(self) -> bool:
+        with self.path.open("rb") as f:
+            config_doc = parse(f.read())
+        return config_doc.get("tool", {}).get("commitizen") is not None
+
     def init_empty_config_content(self) -> None:
         config_doc = TOMLDocument()
         if os.path.isfile(self.path):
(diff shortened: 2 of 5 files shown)


commitizen/config/base_config.py
```diff
@@ -17,10 +17,16 @@
...
-        self.is_empty_config = False
         self._settings: Settings = DEFAULT_SETTINGS.copy()
         self._path: Path | None = None
 
+    def contains_commitizen_section(self) -> bool:
+        """Check if the config file contains a commitizen section.
+
+        The implementation is different for each config file type.
+        """
+        raise NotImplementedError()
```
bearomorphism, line 28: Not sure whether this is a good enough design. I think the state `is_empty_config` is a bit hard to track. The property is only used in `read_cfg` function call though.

commitizen/config/toml_config.py
```diff
@@ -26,6 +26,15 @@ def __init__(self, *, data: bytes | str, path: Path) -> None:
...
         self._parse_setting(data)
 
+    def contains_commitizen_section(self) -> bool:
+        with self.path.open("rb") as f:
+            config_doc = parse(f.read())
+        tool = config_doc.get("tool")
+        if tool is None:
+            return False
+        commitizen = tool.get("commitizen")
+        return commitizen is not None
```
noirbizarre: I tend to do 1-liners for this kind of check:

```suggestion
        return config_doc.get("tool", {}).get("commitizen") is not None
```

commitizen/config/base_config.py
noirbizarre: It will be easier to track the logic, still better than the existing approach 👍🏼 
Maybe make BaseConfig an abstract base class and `contains_commitizen_section` and abstract method. It will raise an exception sooner if implementation is missing. (If we don't instantiate the base class)

bearomorphism: I remember I have tried to make `BaseConfig` abstract before, but there are a few lines of production code and a lot of lines of test code that instantiate `BaseConfig` if I remembered it correctly. Of course instantiating the base class is a bit strange to me.

I can try to make `BaseConfig` abstract in another PR after the big test PRs get reviewed and merged.

commitizen/config/toml_config.py
bearomorphism: Thanks! This looks a lot cleaner

PR #1849, opened 2026-02-05
feat(commit): add a tag(--body-length-limit) and a function for command commit

### LINKED ISSUE #1597: Add option for body line length
### Description

To allow for more granular control over commit message format, it would be useful if the line length of the commit body could be set. This should be settable from both the config or via a command line argument. 

I was hoping to use commitizen to enforce the 50/72 rule by setting those limits for the subject and body, without the need to customise. 

I found the `--message-length-limit` command line option and am excited to see that there is work being done in [#1574](https://github.com/commitizen-tools/commitizen/pull/1574) to add this to the config. However, this only affects the first line 'subject' of the message.

To be clear, the proposed option should affect the length of each line in the commit body, not the total length of the body.

### Possible Solution

A new config option would be introduced that allows the maximum line length of the commit message body to be set. Suggested, descriptive config option name:

`body-line-length-limit`

The existing `message_length_limit` option could be renamed to more explicitly describe its effect and to align the naming of the two options, becoming:

`subject-length-limit`

The corresponding command line versions would become:

`--body-line-length-limit`

and

`--subject-length-limit`

The short option `--l` would be removed as it would appear ambiguous in this scenario.

### Additional context

_No response_

### Related issues

In progress issue supporting the subject length being set by config [#1574](https://github.com/commitizen-tools/commitizen/pull/1574)

yjaw:
## Description
I added a tag (—body-length-limit) for command commit. This tag utilizes Python’s built-in library, textwrap, to rewrap the body. It also respects the user’s |(\n) signal. This tag will affect the footer as well.
The flow is as follows:
1. Split the body into separate lines using the `\n` character.
2. Start from the third line, where the body begins. Use `textwrap` to rewrap that line.
3. Reconstruct the entire body message from those rewrapped lines.

## Checklist

Generated-by: [Gemini] following [the guidelines](http://commitizen-tools.github.io/commitizen/contributing/pull_request/#ai-assisted-contributions)

### Code Changes

### Documentation Changes

- [ ] Run `uv run poe doc` locally to ensure the documentation pages renders correctly
- [ ] Check and fix any broken links (internal or external)

## Expected Behavior

1. This option should modify the line length for your message if its length exceeds the limit you’ve set. 
2. Adding this tag or setting it to 0 should have no effect. 
3. If the message body is empty, it will be fine. 
4. This option will affect the text in both the body and footer sections. 

1. When using the commit command, you can add the —body-length-limit [int] option to set a limit on the length of the commit message. If you don’t want to set a limit, you can set the option to 0.

## Additional Context

close #1597


tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,230 @@ def test_commit_command_with_config_message_length_limit(
...
+            "footer": "",
+        },
+    )
+
+    commit_mock = mocker.patch(
+        "commitizen.git.commit", return_value=cmd.Command("success", "", b"", b"", 0)
+    )
+
+    # Execute with body_length_limit
+    commands.Commit(config, {"body_length_limit": 72})()
```
bearomorphism: The number 72 should be extracted.

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,230 @@ def test_commit_command_with_config_message_length_limit(
...
+    lines = committed_message.split("\n")
+    assert lines[0] == "feat: add feature"
+    assert lines[1] == ""
+    body_lines = lines[2:]
+    for line in body_lines:
+        if line.strip():
+            assert len(line) <= 72, (
+                f"Line exceeds 72 chars: '{line}' ({len(line)} chars)"
+            )
+
```
bearomorphism: Add a file regression check here and remove `assert line[0] == ...` and `assert line[1] == ...`

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,230 @@ def test_commit_command_with_config_message_length_limit(
...
+            "scope": "",
+            "is_breaking_change": False,
+            "body": "This is a very long line that exceeds 72 characters and should be automatically wrapped by the system to fit within the limit",
+            "footer": "",
+        },
+    )
+
+    commit_mock = mocker.patch(
+        "commitizen.git.commit", return_value=cmd.Command("success", "", b"", b"", 0)
+    )
```
bearomorphism: Please reuse `commit_mock` fixture. Same for other new tests.

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,230 @@ def test_commit_command_with_config_message_length_limit(
...
+    success_mock.assert_called_once()
+
+    # Verify wrapping occurred
+    committed_message = commit_mock.call_args[0][0]
+    lines = committed_message.split("\n")
+    assert lines[0] == "feat: add feature"
+    assert lines[1] == ""
+    body_lines = lines[2:]
+    for line in body_lines:
+        if line.strip():
```
bearomorphism: This `if` can be removed?

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,230 @@ def test_commit_command_with_config_message_length_limit(
...
+
+    # Should have a subject, a blank line
+    assert lines[0] == "feat: add feature"
+    assert lines[1] == ""
+    # Each original line should be wrapped separately, preserving the line breaks
+    body_lines = lines[2:]
+    # All lines should be <= 45 chars
+    for line in body_lines:
+        if line.strip():
+            assert len(line) == 45, (
```
bearomorphism: Why not `<=`?

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,230 @@ def test_commit_command_with_config_message_length_limit(
     success_mock.reset_mock()
     commands.Commit(config, {"message_length_limit": 0})()
     success_mock.assert_called_once()
+
+
+@pytest.mark.usefixtures("staging_is_clean")
+def test_commit_command_with_body_length_limit_wrapping(
```
bearomorphism: I think all the new tests can be merged into one and use parameterize to shorten the tests

Basically we can do like following:

```python

fixtures
parameterize
def test_...(...):
    mocker.patch(
        "questionary.prompt",
        return_value={
            "prefix": "feat",
            "subject": "add feature",
            "scope": "",
            "is_breaking_change": False,
            "body": body, # parameterized, can be a long text with / without line breaks or short strings or an empty string
            "footer": "",
        },
    )

    # commit with parameterized config

    # check each line in line[2:] does not exceed the line length limit if it is not 0

    # file regression check
```

commitizen/commands/commit.py
```diff
@@ -100,6 +105,31 @@ def _validate_subject_length(self, message: str) -> None:
...
             )
 
+    def _rewrap_body(self, message: str) -> str:
+        body_length_limit = self.arguments.get(
+            "body_length_limit", self.config.settings.get("body_length_limit", 0)
+        )
+        # By the contract, body_length_limit is set to 0 for no limit
+        if (
+            body_length_limit is None or body_length_limit <= 0
+        ):  # do nothing for no limit
```
bearomorphism:
```suggestion
        if (
            body_length_limit <= 0
        ):  # do nothing for no limit
```

commitizen/commands/commit.py
```diff
@@ -100,6 +105,31 @@ def _validate_subject_length(self, message: str) -> None:
...
+            return message
+
+        message_parts = message.split("\n", 2)
```
bearomorphism:
```suggestion
        lines = message.split("\n")
```

and I'm entirely not sure whether there will be issues related to LF and CRLF. Need to check further

commitizen/commands/commit.py
```diff
@@ -100,6 +105,31 @@ def _validate_subject_length(self, message: str) -> None:
...
+        # First line is subject, second is blank line, rest is body
+        subject = message_parts[0]
+        blank_line = message_parts[1]
+        body = message_parts[2].strip()
+        body_lines = body.split("\n")
+        wrapped_body_lines = []
+        for line in body_lines:
+            wrapped_body_lines.append(textwrap.fill(line, width=body_length_limit))
+        wrapped_body = "\n".join(wrapped_body_lines)
+        return f"{subject}\n{blank_line}\n{wrapped_body}"
```
bearomorphism:
```suggestion
        # First line is subject, second is blank line, rest is body
        wrapped_body_lines = [textwrap.fill(line, width=body_length_limit) for line in lines[2:]]
        return "\n".join(chain(lines[:2], wrapped_body_lines))
```

bearomorphism (commented):
Thanks!

tests/commands/test_commit_command.py
yjaw: In this test case, every line is exactly 45 characters long, so I check if it truly wraps to 45-character lines. However, there’s a case where it wraps to a line shorter than 45 characters which is not the intended behavior but test still can pass.

* I forgot to modify the comment.

commitizen/commands/commit.py
yjaw: Wow, this code is so clean! I Learned a lot.

commitizen/commands/commit.py
yjaw: Quick question: I understand that the value is guaranteed, but according to defensive programming, isn’t it better to handle the null value?

tests/commands/test_commit_command.py
yjaw: I understand! This will make the test more organized.

tests/commands/test_commit_command.py
yjaw: I believe so. This line was intended to skip empty lines, which are no longer necessary.

commitizen/commands/commit.py
bearomorphism: I am not sure whether `body_length_limit` is guaranteed to be an `int`. If it is `None` under some code paths, then something might be wrong.

commitizen/commands/commit.py
```diff
@@ -100,6 +105,31 @@ def _validate_subject_length(self, message: str) -> None:
...
+            return message
+
+        # First line is subject, second is blank line, rest is body
+        subject = message_parts[0]
+        blank_line = message_parts[1]
+        body = message_parts[2].strip()
+        body_lines = body.split("\n")
+        wrapped_body_lines = []
+        for line in body_lines:
+            wrapped_body_lines.append(textwrap.fill(line, width=body_length_limit))
```
bearomorphism: we should use `wrap` instead of `fill`

Reference: https://docs.python.org/3/library/textwrap.html#textwrap.wrap

tests/commands/test_commit_command.py
bearomorphism: That is strange. Wrapping should not guarantee exactly N chars long.

commitizen/commands/commit.py
```diff
@@ -100,6 +105,31 @@ def _validate_subject_length(self, message: str) -> None:
                 f"Length of commit message exceeds limit ({len(subject)}/{message_length_limit}), subject: '{subject}'"
             )
 
+    def _rewrap_body(self, message: str) -> str:
```
bearomorphism: why `_rewrap`? I think `_wrap_body` is better

commitizen/commands/commit.py
Lee-W: I probably would try `not body_length_limit or body_length_limit < 0`

commitizen/commands/commit.py
```diff
@@ -100,6 +105,31 @@ def _validate_subject_length(self, message: str) -> None:
...
+        body_length_limit = self.arguments.get(
+            "body_length_limit", self.config.settings.get("body_length_limit", 0)
+        )
+        # By the contract, body_length_limit is set to 0 for no limit
+        if (
+            body_length_limit is None or body_length_limit <= 0
+        ):  # do nothing for no limit
+            return message
+
+        lines = message.split("\n")
```
Lee-W, line 119: looks like `lines` can be moved below

commitizen/commands/commit.py
Lee-W: also we should prefer using list comphrehension

commitizen/cli.py
```diff
@@ -164,6 +164,11 @@ def __call__(
                         "type": int,
                         "help": "Set the length limit of the commit message; 0 for no limit.",
                     },
+                    {
+                        "name": ["--body-length-limit"],
+                        "type": int,
+                        "help": "Set the length limit of the commit body. Commit message in body will be rewrapped to this length; 0 for no limit.",
```
Lee-W, line 171: I think we can set a default value here?

yjaw: Yes, I agree. What’s the difference between setting a default value here and using default_setting in defaults.py?

commitizen/commands/commit.py
yjaw: Yes, I found that if I don’t handle None, it can’t commit due to a Null error.

commitizen/commands/commit.py
yjaw: Why is `wrap` used instead of `fill` in this context?

commitizen/cli.py
bearomorphism: Actually I think `DEFAULT_SETTINGS` is not a good design. I am not sure the mechanism of default value in `argparse`, not sure the exact difference here.

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,79 @@ def test_commit_command_with_config_message_length_limit(
     success_mock.reset_mock()
     commands.Commit(config, {"message_length_limit": 0})()
     success_mock.assert_called_once()
+
+
+@pytest.mark.usefixtures("staging_is_clean")
+@pytest.mark.parametrize(
+    ("test_id", "body", "body_length_limit"),
```
bearomorphism: please remove "test_id" and use `pytest.param` instead

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,79 @@ def test_commit_command_with_config_message_length_limit(
...
+    commands.Commit(config, {})()
+
+    success_mock.assert_called_once()
+    committed_message = commit_mock.call_args[0][0]
+
+    # File regression check - uses test_id to create separate files
+    file_regression.check(
+        committed_message,
+        extension=".txt",
+        basename=f"test_commit_command_body_length_limit_{test_id}",
```
bearomorphism: not needed

commitizen/commands/commit.py
```diff
@@ -100,6 +106,24 @@ def _validate_subject_length(self, message: str) -> None:
...
+        if len(lines) < 3:
+            return message
+
+        # First line is subject, second is blank line, rest is body
+        wrapped_body_lines = [
+            textwrap.wrap(line, width=body_length_limit) for line in lines[2:]
+        ]
+        return "\n".join(chain(lines[:2], chain.from_iterable(wrapped_body_lines)))
```
bearomorphism: why `from_iterable` here

tests/commands/test_commit_command.py
```diff
@@ -365,3 +365,79 @@ def test_commit_command_with_config_message_length_limit(
...
+        },
+    )
+
+    config.settings["body_length_limit"] = body_length_limit
+    commands.Commit(config, {})()
+
+    success_mock.assert_called_once()
+    committed_message = commit_mock.call_args[0][0]
+
+    # File regression check - uses test_id to create separate files
```
bearomorphism: remove this comment

commitizen/commands/commit.py
```diff
@@ -100,6 +106,24 @@ def _validate_subject_length(self, message: str) -> None:
                 f"Length of commit message exceeds limit ({len(subject)}/{message_length_limit}), subject: '{subject}'"
             )
 
+    def _wrap_body(self, message: str) -> str:
```
woile, line 107: add a docstring explaining what this `_wrap_body` does, and why it was introduced 🙏🏻

commitizen/cli.py
woile: Normally the default should be in DEFAULT_SETTINGS, which can be updated from the .cz.toml file. A default in argparse may override the setting in the file. You can add a test, to check that this is the case.

commitizen/commands/commit.py
```diff
@@ -100,6 +106,24 @@ def _validate_subject_length(self, message: str) -> None:
...
+        body_length_limit = self.arguments.get(
+            "body_length_limit", self.config.settings.get("body_length_limit", 0)
```
woile: Don't use `get` here, settings already has a default and you know it exists, if it doesn't this should explode, so it would be better to use:

```
self.config.settings["body_length_limit"]
```

This conveys the right meaning: there must be a `body_length_limit` in the settings, which, based on your code, it's there.

commitizen/commands/commit.py
yjaw: Since `wrapped_body_lines` is a list of lists of strings, I flatten it using `chain.from_iterable` before chaining it to a list of strings.
Do you prefer I manually flatten it? `flatten_body = [line for sublist in wrapped_body_lines for line in sublist]`

tests/commands/test_commit_command.py
yjaw: How can I ensure the first line (subject) and second line (blank) are as expected? Is this no need for the test, or is there another method to test it?

bearomorphism: Apologies I wasn't clear enough. I meant the `basename` parameter is not needed here.

commitizen/commands/commit.py
bearomorphism: oh I see. I thought `textwrap.wrap` returns a list of strings.

Then the following is more semantically correct:

```
wrapped_body_lines = chain.from_iterable(
            textwrap.wrap(line, width=body_length_limit) for line in lines[2:]
        )
```

commitizen/commands/commit.py
yjaw: In this case, letting it explode can remind programmers that the behavior is not expected (it should always have a value)?

commitizen/cli.py
yjaw: I agree that multiple default setting is undesirable. Should I remove the default setting for this tag from `cli.py`?

commitizen/commands/commit.py
bearomorphism: because we want to wrap instead of fill?

commitizen/cli.py
bearomorphism: Yes, after a second look, there should not be default values in cli.py because we want cz to use the settings from the configuration file if the value is None in arguments

commitizen/commands/commit.py
woile: Let me rephrase it. 
The code you've written *guarantees* that the value exists, right? Because you've set a default in the settings already.
If that was not the case, there's a bug, and we should surface that bug as soon as possible. Using `.get` would only hide the bug.

bearomorphism: We probably need to check if similar patterns exists in the code base. I probably introduced some in the past...

bearomorphism:
There are conflicts. Turning this PR to draft

commitizen/commands/commit.py
yjaw: I got it! Thanks for explaining. In this context, we should raise an error instead of using `.get` to hide the unexpected behavior.

commitizen/commands/commit.py
yjaw: First time using Generator. It's so cool!

schlotter:
Many thanks for this PR, exactly what I'm looking for 😊

Any help needed to move this over the finish line?

bearomorphism:
Thanks for the reminder. I believe the maintainers barely have bandwidth to review PRs. I'll take a quick look.

commitizen/commands/commit.py
bearomorphism: resolved

bearomorphism (approved):
LGTM.

Lee-W (approved):

woile:
Could you rebase? 🙏🏻

PR #1877, opened 2026-02-23
docs: add git config user.* to commit successfully

### LINKED ISSUE #1876: Broken commit.gif in docs
### Type

- [x] Content inaccurate
- [ ] Content missing
- [ ] Typo

### URL

NA

### Description

See the commit https://github.com/commitizen-tools/commitizen/commit/44ec9b8f4fcc64300169207f28bfd8bc7595987a

The gif is broken after running github action. need to setup git config first

bearomorphism:
Fixes #1876 

Verified the solution by running `vhs docs/images/commit.tape`, and the generated `/tmp/commitizen-demo` folder is with correct git log (with a commit authored by `you@example.com`)

### DIFF
--- a/docs/images/commit.tape
+++ b/docs/images/commit.tape
@@ -53,6 +53,10 @@ Sleep 500ms
 # Initialize git repository
 Type "git init"
 Enter
+Type "git config user.email 'you@example.com'"
+Enter
+Type "git config user.name 'Your Name'"
+Enter
 Sleep 500ms
 
 Type "git checkout -b awesome-feature"


bearomorphism:
@HankyStyle do you want to take a look on this fix?

HankyStyle:
Yes, this PR fixes the issue. The problem wasn't caught earlier because I had global git config set locally, so it always worked on my machine.

Thanks @bearomorphism for the fix! 

I've tested it with clean environment and confirmed it works correctly.

Before:

Screenshot at Feb 23 19-38-34

After:
commit_success_final_success_1771850407987

bearomorphism:
Thanks for double confirming that this PR actually fixes the issue :)

Issue #1897, opened 2026-03-17
cz info does not read from info_path for non root directory


dracorp:
### Description

My cz config:
```
[tool.commitizen]
name = "cz_customize"
[tool.commitizen.customize]
info_path = 'README_cz.md'
```

`cz info` should display the contents of the file at `info_path` from any directory within the repository.

### Steps to reproduce

1. Modify .cz.toml 
2. Run `cz info`

### Current behavior

```
$ cd directory
$ cz info
Traceback (most recent call last):
  File "/opt/homebrew/bin/cz", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Cellar/commitizen/4.13.9/libexec/lib/python3.14/site-packages/commitizen/cli.py", line 689, in main
    args.func(conf, arguments)()  # type: ignore[arg-type]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/commitizen/4.13.9/libexec/lib/python3.14/site-packages/commitizen/commands/info.py", line 13, in __call__
    out.write(self.cz.info())
              ~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/commitizen/4.13.9/libexec/lib/python3.14/site-packages/commitizen/cz/customize/customize.py", line 72, in info
    return Path(info_path).read_text(encoding=self.config.settings["encoding"])
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.3_1/Frameworks/Python.framework/Versions/3.14/lib/python3.14/pathlib/__init__.py", line 787, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors, newline=newline) as f:
         ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.3_1/Frameworks/Python.framework/Versions/3.14/lib/python3.14/pathlib/__init__.py", line 771, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
[...]

PR #1898, opened 2026-03-17
fix(customize): resolve info_path relative to root config

dracorp:
For https://github.com/commitizen-tools/commitizen/issues/1897


bearomorphism:
Could you follow the PR description template and add test cases for your fix? Thanks!

PR #1982, opened 2026-05-09
test(cli): skip invalid-command snapshot on Python 3.12.0-3.12.6

### LINKED ISSUE #1864: Bug: test_invalid_command fails on Python 3.12
### Description

The `test_invalid_command` fails on Python 3.12 due to a mismatch in how argparse formats the "invalid choice" error message. While our existing regression file for Python 3.12 expects **no** quotes around choices, the execution on Python 3.12 seems to produce output that conflicts with the stored reference file, leading to a pytest-regressions failure.

### Steps to reproduce

1. Use a Python 3.12.3 environment.
2. Run the test suite:
  `uv run pytest tests/test_cli.py::test_invalid_command`
3. Observe the `AssertionError` from `file_regression.check`.

### Current behavior

The test fails because the obtained output contains single quotes around the valid choices, whereas the reference file  does not.
Diff output:

Image

### Desired behavior

The regression files for Python 3.12 should be standardized. We need to decide whether Python 3.12 should strictly follow the "no quotes" format or if the reference files need an update to accommodate the specific minor version of Python 3.12 being used.

### Screenshots

Image

### Environment

Commitizen Version: 4.13.6
Python Version: 3.12.3 (main, Jan 22 2026, 20:57:42) [GCC 13.3.0]
Operating System: Linux

bearomorphism:
## Description

Closes #1864.

### Why

`tests/test_cli.py::test_invalid_command` is a snapshot test that captures the stderr output of `cz <invalid>`. The output is generated by argparse, and CPython 3.13 changed the format of the `invalid choice:` error from quoted choices (`'init', 'commit', ...`) to unquoted (`init, commit, ...`) — see CPython gh-129019. The change was backported to 3.12.7. The repo's reference snapshots reflect the no-quote format, so:

| Python | Quoted? | Snapshot match? |
| --- | --- | --- |
| 3.10.x, 3.11.x | yes | yes (snapshots use quotes) |
| 3.12.0 - 3.12.6 | yes | **no — test fails** |
| 3.12.7+, 3.13.x, 3.14.x | no | yes (snapshots use no quotes) |

A user on 3.12.3 hits `AssertionError: FILES DIFFER` running `pytest tests/test_cli.py::test_invalid_command`. The CI matrix runs on a recent 3.12.x, so CI doesn't catch it.

### What changed

| File | Change |
| --- | --- |
| `tests/test_cli.py` | Add `pytest.mark.skipif(sys.version_info[:2] == (3, 12) and sys.version_info < (3, 12, 7), reason=...)` to the `test_invalid_command` parametrised test. The reason text references CPython gh-129019 and this issue. |

### How it works

- The first guard `sys.version_info[:2] == (3, 12)` scopes the skip to Python 3.12.x only — older 3.10/3.11 (which still print quoted choices) and newer 3.13/3.14 (which print unquoted) are unaffected.
- The second guard `sys.version_info < (3, 12, 7)` uses Python's lexicographic tuple comparison: `(3, 12, 6, ...) < (3, 12, 7)` is `True`, `(3, 12, 7, ...) < (3, 12, 7)` is `False`. So 3.12.0 through 3.12.6 skip; 3.12.7+ run.
- The `--invalid-arg` parametrised case shares the same decorator. Its argparse path also formats choices, so the same skip logic applies even though the diff message is slightly different.
- 3.10 / 3.11 snapshots use quotes, so they would *also* break if argparse's no-quote change is back-ported to those branches. However, both are in security-only maintenance (3.10 is EOL, 3.11 since April 2024), and a formatting-only change won't be merged under that policy. We don't need to skip 3.10/3.11.

### Why a `skipif` instead of duplicate snapshots?

A `skipif` is cheaper to maintain (no new fixture file, no risk of the duplicate snapshot drifting). It's also more honest: argparse's behaviour on 3.12.0-3.12.6 is a known upstream issue, fixed in 3.12.7. Users on those patches can upgrade; we don't need commitizen to keep accumulating fixtures for upstream's old behaviour.

### Backward compatibility

- Test-only change. No behaviour change for `cz` users.
- The CI matrix (Ubuntu/macOS/Windows × 3.10/3.11/3.12/3.13/3.14) continues to exercise `test_invalid_command` on every supported Python because the GitHub-hosted runners use a recent patch within each minor (3.12.10+ at the time of writing).

## Checklist

Generated-by: Claude following [the guidelines](http://commitizen-tools.github.io/commitizen/contributing/pull_request/#ai-assisted-contributions)

### Code Changes

- [x] Add test cases to all the changes you introduce (this PR is a test-only change; the regression itself is the existing snapshot, the fix is the skip)
- [x] Run `uv run poe all` locally to ensure this change passes linter check and tests (`poe lint` clean; verified the test runs on a recent 3.12.x and skips on an old one — see "Steps to Test")
- [x] Manually test the changes (see "Steps to Test" below)
- [x] Update the documentation for the changes (no doc change required — this is a test-fixture compatibility fix)

## Expected Behavior

| Python interpreter | `pytest tests/test_cli.py -k test_invalid_command` |
| --- | --- |
| 3.10.x | runs and passes (snapshot uses quotes, argparse uses quotes) |
| 3.11.x | runs and passes (same) |
| 3.12.0 - 3.12.6 | **skipped** with the `gh-129019` rationale |
| 3.12.7+ | runs and passes (snapshot has no quotes, argparse has no quotes) |
| 3.13.x | runs and passes (same) |
| 3.14.x | runs and passes (same) |

```sh
git fetch fork fix/1864-skip-old-3.12-argparse-snapshot
git checkout fork/fix/1864-skip-old-3.12-argparse-snapshot

# 1. On a recent 3.12 patch (any 3.12.7+) — both parametrised cases run and pass.
uv run --python 3.12 pytest tests/test_cli.py::test_invalid_command -v
# expected: 2 passed

# 2. On an old 3.12 patch (where argparse still quotes) — both are skipped with the gh-129019 reason.
uv run --python 3.12.3 pytest tests/test_cli.py::test_invalid_command -v
# expected: 2 skipped (with the rationale visible in -v output)

# 3. On every other supported Python — same as before, no skip.
for py in 3.10 3.11 3.13 3.14 ; do
  uv run --python "$py" pytest tests/test_cli.py::test_invalid_command -v
done
# expected: 2 passed for each

# 4. Full test_cli.py suite (sanity check that nothing else broke).
uv run pytest tests/test_cli.py -q
# expected: 22 passed, 1 skipped (on old 3.12) OR 23 passed (on a recent 3.12 / other versions)
```

## Additional Context

Surfaced while triaging open issues in #1976 (round 3). The original reporter and `@noirbizarre` (the assignee) had already discussed skipping the test on 3.12 — this PR scopes that skip to the affected patch range so the test still runs on every patch where the snapshot is valid. Reviewed by an internal `claude-sonnet-4.6` code-review pass and by GitHub Copilot; both confirmed the version-tuple comparison and that the 3.10/3.11 snapshots are safe (security-only maintenance precludes the same back-port there).

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -34,6 +34,15 @@ def test_no_argv(util: UtilFixture, capsys, file_regression):
         "invalidCommand",
     ],
 )
+@pytest.mark.skipif(
+    sys.version_info[:2] == (3, 12) and sys.version_info < (3, 12, 7),
+    reason=(
+        "argparse stopped quoting choices in 3.13 (CPython gh-129019), "
+        "backported to 3.12.7. The reference snapshot reflects the "
+        "no-quote format, so older 3.12.x patches (3.12.0-3.12.6) print "
+        "quoted choices and fail. See commitizen-tools/commitizen#1864."
+    ),
+)
 @pytest.mark.usefixtures("python_version", "consistent_terminal_output")
 def test_invalid_command(util: UtilFixture, capsys, file_regression, arg):
     with pytest.raises(NoCommandFoundError):


Lee-W (approved):
it seems we're using a different way than the 3.14.5 one. but not a big deal

Issue #1385, opened 2025-04-15
Deprecate `cz_customize`


Lee-W:
### Description

https://github.com/commitizen-tools/commitizen/pull/1381#issuecomment-2801394198

AxTheB:
Hello. To start, I did a MR that adds customizing to conventional_commits. It can either be done to jira commitizen too, or moved to the base class, as you see fit.

Lee-W:
I'm sorry for not getting back to you sooner. Crazy busy since July... (probably till mid Oct...) I have already assigned myself to take a look. Will try to do so when bandwidth allow. Thanks so much for helping out!

---

https://github.com/commitizen-tools/commitizen/issues/1588 has more details on the user requirements. But let's keep the discussion here :)

ryanrapini-bep:
+1 to this as i'm facing this now. For our purposes I want to run a much more cut-down version of cz that simply:

1. Only gives `feat`, `fix`, `test`, `docs`, `chore`, as my options as these are our general use cases
2. skips all other questions except "Write a summary of the code changes" and then "Is this breaking change y/n"

I started on customizing this and got this far:

```

[tool.commitizen]  # https://commitizen-tools.github.io/commitizen/config/
bump_message = "bump: v$current_version → v$new_version"
tag_format = "v$version"
update_changelog_on_bump = true
version_provider = "uv"
name = "cz_customize"

[[tool.commitizen.customize.questions]]
type = "list"
name = "change_type"
choices = [
  {value = "feat", name = "feat: A new feature. Correlates with MINOR in SemVer"}, 
  {value = "fix", name = "fix: A bug fix. Correlates with PATCH in SemVer"}, 
  {value = "test", name = "test: Test Additions."}, 
  {value = "docs", name = "docs: Documentation Additions."}, 
  {value = "chore", name = "chore: Maintenance/cleanup work"}
]
message = "Select the type of change you are committing"

[[tool.commitizen.customize.questions]]
type = "input"
name = "message"
message = "Write a short and imperative summary of the code changes"
```

but it's simply not worth the amount of time I've put into this as I was hoping for something drop-in with easy customization and I gave up trying to figure out how to re-implement the "Breaking change y/n" feature.

At the bare minimum it would be nice to have in the docs the current default workflow re-created in `cz_customize` in .toml format so that users could quickly grab that and tweak it. as it is now I feel as if I am trying to re-invent the whole library to get some minor tweaks for myself. happy to contribute

[...]

PR #2006, opened 2026-05-30
feat(ConventionalCommitsCZ): add support for customizable change type choices and bump map overrides

rockleona:
## Description

According to #1385, I've adding `override` and `extend` attributes under `Settings`, basically, the content was inherited using `CzSettings`.

## Checklist

- [x] Yes (GitHub Copilot)

### Code Changes

### Override
```toml
[tool.commitizen]

[tool.commitizen.override]
bump_pattern = "^((feat|fix|perf)(\\(.+\\))?!?):"
commit_parser = "^((?P<change_type>feat|fix|perf)(?:\\((?P<scope>[^()\\r\\n]*)\\))?(?P<breaking>!)?):\\s(?P<message>.*)$"
changelog_pattern = "^(feat|fix|perf)"
bump_map = { "^feat" = "MINOR", "^fix" = "PATCH", "^perf" = "PATCH" }
change_type_map = { feat = "Features", fix = "Bug Fixes", perf = "Performance" }
change_type_choices = [
  { value = "feat", name = "feat: A new feature", key = "f" },
  { value = "fix", name = "fix: A bug fix", key = "x" }
]
```

### Extend
```toml
[tool.commitizen]

[tool.commitizen.extend]
bump_map = { "^deps" = "PATCH" }
change_type_map = { deps = "Dependencies" }
change_type_choices = [
  { value = "deps", name = "deps: Dependency updates", key = "e" }
]
```

## Additional Context
issue #1385

### DIFF
--- a/commitizen/cz/conventional_commits/conventional_commits.py
+++ b/commitizen/cz/conventional_commits/conventional_commits.py
@@ -1,13 +1,16 @@
 from __future__ import annotations
 
+from collections import OrderedDict
 from pathlib import Path
 from typing import TYPE_CHECKING, TypedDict
 
 from commitizen import defaults
 from commitizen.cz.base import BaseCommitizen
 from commitizen.cz.utils import multiple_line_breaker, required_validator
+from commitizen.question import Choice
 
 if TYPE_CHECKING:
+    from commitizen.config import BaseConfig
     from commitizen.question import CzQuestion
 
 __all__ = ["ConventionalCommitsCz"]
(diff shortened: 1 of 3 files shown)


bearomorphism:
@Lee-W Could you take a look? You have more context on this feature.

woile:
I'll take a look next week

woile (commented):
This is good work! 

However, I don't think override and extend should happen at the `conventional_commits.py` level. **This change should work for any custom rules.**

Instead, I think the override should directly happen here:
https://github.com/commitizen-tools/commitizen/blob/master/commitizen/cli.py#L710

The hierarchy is:

```
settings <- override/extend <- cli
```
Meaning that anything I provide via cli should override the override/extend and the settings. Which is not happening now.

cc: @Lee-W

Lee-W:
@rockleona could you try to resolve the comments on this PR? It's super hard to read with the OpenAI comments....

---

Yep, agreed with @woile — since `bump_pattern / bump_map / bump_map_major_version_zero / change_type_map / change_type_order / changelog_pattern / commit_parser` all live on `BaseCommitizen`, the override/extend logic should move there so it works for any rule set, not just conventional commits.

Manny7717 (commented):
## Verified review (head 1461705)

Locally verified on head (repo venv, Python 3.12):

- **Tests:** `test_cz_conventional_commits.py` 30/30 pass (incl. 4 new); `test_conf.py` 67/67. Full suite 1285 passed / 6 failed / 2 xfailed — the 6 failures (`test_bump_pre_commit_changelog*`) fail byte-identically on `origin/master` (pre-existing env noise from pre-commit/prek hooks). **Zero new failures.**
- **Lint/type:** `ruff check` clean on both changed files + tests; `mypy` clean on the two source files.
- **Behavior probes (all passed):** override replaces maps wholesale; extend merges into defaults while retaining them (OrderedDict append); `questions()[0]` carries extended choices; `bump._find_increment` reads `self.cz.bump_map` so overrides take effect; instance isolation works (`_isolate_mutable_defaults` correctly prevents cross-instance leakage — verified two instances don't share the extended map); `elif` gives override precedence over extend when both are set.
- **Interop with #2081 (strict_config, by Manny7717):** KNOWN_SETTINGS there derives from the `Settings` TypedDict keys, so the new `override`/`extend` keys are auto-included when both merge — no conflict.

## Finding: custom change types cannot pass `cz check` (integration gap)

The feature lets users extend `change_type_choices`, and the interactive prompt happily offers the custom type (verified: `questions()[0]["choices"]` ends with the extended choice). But **`cz check` uses `schema_pattern()`, whose type list is hardcoded** (conventional_commits.py:243-266) and is *not* covered by the new override/extend settings.

Verified end-to-end with the real CLI:

```
[tool.commitizen.extend]
change_type_choices = [{ value = "foo", name = "foo: custom type", key = "o" }]
```

- `cz commit` accepts the `foo` choice and produces `foo: added thing`
- `cz check --message "foo: added thing"` → **`commit validation: failed!`** with pattern `(build|bump|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)...`
- `cz bump` computes **no increment** for `foo:` commits unless `bump_map` is also extended (expected), but note `commit_parser` also needs overriding for changelog classification

So the three consumers (prompt, check, bump/changelog) each use different source-of-truth lists, and the settings mechanism covers only two of them. A user who follows the PR's own "Steps to Test" (`extend.change_type_choices` only) ends up able to *create* a commit that the project's own validator rejects. This is a concrete instance of the design concern woile/Lee-W raised in June (override/extend should live on `BaseCommitizen` / apply to all rule sets): even scoped to conventional commits, the feature is internally inconsistent.

Suggested fixes (any one):
1. Derive `schema_pattern()`'s type alternatives from `change_type_choices` (falling back to the hardcoded list), or
2. Allow `schema_pattern` in override/extend settings, or
3. Document explicitly in the PR that custom types require overriding `commit_parser` + `bump_map` + `schema_pattern`-equivalent, and add a regression test that runs `cz check` on a message with an extended type.

## Minor notes

- **Falsy-value skip:** `_apply_override_settings`/`_apply_extend_settings` use `if x := settings.get(...)`, so empty-dict values are silently ignored — `override: {bump_map: {}}` keeps the default map instead of clearing it. Edge case, but a user attempting to *remove* entries this way gets no feedback.
- **Docs:** the new `override`/`extend` keys are not documented in `docs/config/option.md`/`configuration_file.md` (the PR checklist's "Update the documentation" is unchecked). For a user-facing config feature this will likely be requested before merge.
- **Staleness:** no commits since 2026-05-30 and the June maintainer feedback (woile, Lee-W) is unaddressed — the design-direction question should be resolved before this lands.

PR #2017, opened 2026-06-19
Fix/cargo workspace deps version for publishing

### LINKED ISSUE #2001: CRASH on cargo provider
### Description

Hello,

In cargo_provider.py on line 86, there is `version_workspace = package_content["version"]["workspace"]`.
This causes commitizen to crash (traceback) under the following conditions:
- you have a cargo workspace
- you have a crate that is a worskpace member
- this crate has (in `[package]` key) a `version = 0.1.0`

The error is: `TypeError: string indices must be integers, not 'str'`

This could be fixed by checking that workspace key exists (and check that it is True, which is already done)

### Steps to reproduce

- you have a cargo workspace
- you have a crate that is a worskpace member
- this crate has (in `[package]` key) a `version = 0.1.0`

### Current behavior

TypeError: string indices must be integers, not 'str'

### Desired behavior

no crash

### Screenshots

_No response_

### Environment

Commitizen Version: 4.16.2

vemilano:
## Description
This PR makes 2 things:
1. Fixes #2001. Commitizen was crashing when a workspace member had a fixed version
2. Add support for rust projects with workspace members that want to publish their crates on crates.io

Regarding the second point, to provide more details, crates.io requires that each crate have a version specified in Cargo.toml. For projects that use a workspace, this means that all members using `version.workspace = true` must also have a version specified at the `workspace.dependencies` level (the specified version must be the same as that of `workspace.version`).

When performing a version bump, in addition to updating `workspace.version` (in Cargo.toml) and the versions in Cargo.lock, we must also update the versions specified in Cargo.toml, since they are actually dynamically managed by Cargo via `version.workspace = true`.

Projects such as `crate-ci/cargo-release` or `pksunkara/cargo-workspaces` also use the same process

## Checklist

no

### Code Changes


vemilano:
I checked the fix for #2001 proposed by pull request #2002 and preferred to keep the fix I'm proposing in that pull request. The merged fix introduces a regression, since the `try: expect: NonExistentKey` block was removed, even though the "version" key is optional for a package.

@bearomorphism @Lee-W @woile , do you know if a review of this pull request is planned? Until it's merged, it's not possible to publish to crates.io with commitizen in a Rust project with a workspace.

To recall the objective of this PR, a project may have a Cargo.toml like this:
```toml
[workspace]
members = ["members/*"]
resolver = "2"

[workspace.package]
edition = "2024"
version = "0.1.0"

[workspace.dependencies]
member-hello = { path = "members/hello" }
```

And the Cargo.toml of the workspace member: 
```toml
[package]
name = "member-hello"
version.workspace = true
```

In order to publish on crates.io (`cargo publish --workspace`), every dependency MUST have a specified version, so we need something like:
```toml
[workspace.dependencies]
member-hello = { path = "members/hello", version = "0.1.0" }
```

But this version is dynamically resolved (because the workspace member has `version.workspace = true`). So when we want to cz bump: 
- bump workspace version
- update versions in Cargo.lock
- update the "version" key for workspace dependencies that have a version specified and match a member inheriting (and handle potentially renamed deps)

PR #2042, opened 2026-07-29
docs(tutorial): add Releasing with the SCM Version Provider

woile:
## Description

Adding tutorial for SCM


bearomorphism (approved):
LGTM, but I don't think I'm familiar enough with this area

PR #2065, opened 2026-08-14
fix(init): use prek when installing hooks if pre-commit is missing

### LINKED ISSUE #2018: `cz init` raises FileNotFoundError for `pre-commit` if only `prek` is installed
### Description

During `cz init`, hook installation checks whether `pre-commit` or `prek` is available, but always executes `pre-commit install`. If only `prek` is installed, detection succeeds but execution fails, resulting in a `FileNotFoundError`.

### Steps to reproduce

1. Install commitizen and prek. pre-commit should not be on Path.
2. Run `cz init`
3. Just hit enter on the question `? What types of pre-commit hook you want to install? (Leave blank if you don't want to install) [commit-msg]`
4. Get `FileNotFoundError: [Errno 2] No such file or directory: 'pre-commit'`

### Current behavior

In the init function is a check to ensure that a pre-commit tool is available, but [it checks for pre-commit and prek](https://github.com/commitizen-tools/commitizen/blob/b37c9d7daaffa5f27a2bfbbbb9b89cb6c0985e27/commitizen/project_info.py#L8-L9), while the install command is [invoked with pre-commit](https://github.com/commitizen-tools/commitizen/blob/b37c9d7daaffa5f27a2bfbbbb9b89cb6c0985e27/commitizen/commands/init.py#L137).

### Desired behavior

- Treat `pre-commit` and `prek` as interchangeable hook installers.
- If only one is installed: use it automatically.
- If both are installed: ask the user to choose.

### Screenshots

_No response_

### Environment

Commitizen Version: 4.16.3
Python Version: 3.12.13 (main, Jun 11 2026, 04:55:13) [GCC 14.2.0]
Operating System: Linux

DSeaStar:
## Description

`cz init` already treated `prek` as a valid hook installer (`is_pre_commit_installed()`), but the install step always ran `pre-commit install`. When only `prek` is on PATH, that raises `FileNotFoundError`.

This change:

- Lists available installers (`pre-commit`, `prek`)
- Uses the only one present
- Asks which to use when both are installed
- Invokes `<installer> install --hook-type ...`

Fixes #2018

## Checklist

Generated-by: Cursor Grok following [the guidelines](http://commitizen-tools.github.io/commitizen/contributing/pull_request/#ai-assisted-contributions)

### Code Changes

Manual testing: unit tests cover only-prek, only-pre-commit, both (prompt), and neither (InitFailedError). `ruff check` and `mypy` pass on the touched files. I did not run the full interactive `cz init` UI in a throwaway repo.

### Documentation Changes

- [x] Check and fix any broken links (internal or external)

Docs change is a one-line note in `docs/commands/init.md`. I did not rebuild the full mkdocs site.

## Expected Behavior

- Only `prek` installed: `cz init` runs `prek install`
- Only `pre-commit` installed: same as today (`pre-commit install`)
- Both installed: prompt, then run the chosen installer
- Neither installed: `InitFailedError` mentioning both tools

1. Install commitizen and `prek` only (no `pre-commit` on PATH)
2. Run `cz init` and select a hook type
3. Confirm hooks are installed via `prek` and no `FileNotFoundError`
4. Repeat with only `pre-commit`, then with both (should prompt)

## Additional Context

Owner-confirmed in #2018 (`issue-status: wait-for-implementation`). Desired behavior is from the issue: treat the two tools as interchangeable, auto-select when only one is present, ask when both are.

### DIFF
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -128,13 +128,9 @@ class Init:
             ) as config_file:
                 yaml.safe_dump(config_data, stream=config_file)
 
-            if not project_info.is_pre_commit_installed():
-                raise InitFailedError(
-                    "Failed to install pre-commit hook.\n"
-                    "pre-commit is not installed in current environment."
-                )
+            installer = self._ask_hook_installer()
 
-            cmd_args = ["pre-commit", "install"]
+            cmd_args = [installer, "install"]
             for ty in hook_types:
                 cmd_args.extend(["--hook-type", ty])
             c = cmd.run(cmd_args)
@@ -163,6 +159,29 @@ class Init:
         out.write("\nYou can bump the version running:\n")
         out.info("\tcz bump\n")
         out.success("Configuration complete 🚀")
+
+    def _ask_hook_installer(self) -> str:
+        """Choose ``pre-commit`` or ``prek`` when installing Git hooks.
+
+        Detection already accepts either tool, but install used to
+        hard-code ``pre-commit``. Use the only available installer, or
+        ask when both are on PATH.
+        """
+        installers = project_info.available_hook_installers()
+        if not installers:
+            raise InitFailedError(
+                "Failed to install pre-commit hook.\n"
+                "Neither pre-commit nor prek is installed in the current environment."
+            )
+        if len(installers) == 1:
+            return installers[0]
+
+        installer: str = questionary.select(
+            "Which hook installer do you want to use?",
+            choices=installers,
+            style=self.cz.style,
+        ).unsafe_ask()
+        return installer
 
     def _ask_config_path(self) -> Path:
         filename: str = questionary.select(
--- a/commitizen/commands/init.py
+++ b/commitizen/commands/init.py
@@ -109,13 +109,7 @@ class Init:
             tag_format = self._ask_tag_format(tag)  # confirm & text
             update_changelog_on_bump = self._ask_update_changelog_on_bump()  # confirm
             major_version_zero = self._ask_major_version_zero(version)  # confirm
-            hook_types: list[str] | None = questionary.checkbox(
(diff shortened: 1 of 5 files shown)


commitizen/commands/init.py
```diff
@@ -164,6 +160,29 @@ def __call__(self) -> None:
...
 
+    def _ask_hook_installer(self) -> str:
+        """Choose ``pre-commit`` or ``prek`` when installing Git hooks.
+
+        Detection already accepts either tool, but install used to
+        hard-code ``pre-commit``. Use the only available installer, or
+        ask when both are on PATH.
+        """
+        installers = project_info.available_hook_installers()
+        if not installers:
```
woile, line 184: What if the user doesn't have any installed and doesn't want to use any of these? Like myself.

Maybe if nothing is installed, it should prompt:
> No pre-commit hook detected, skipping question

so users who want this know how to act accordingly (install pre-commit and retry).

What do you think?

woile (commented):
Thanks for the PR, one comment only

commitizen/commands/init.py
DSeaStar: Agreed — if neither pre-commit nor prek is installed we skip the question now and print that message.

DSeaStar:
Addressed in c9ea08e: when neither `pre-commit` nor `prek` is on PATH, `cz init` now skips the hook question entirely and prints `No pre-commit hook detected, skipping question`, so users who want hooks know to install one and re-run.

I also noticed codecov flagged one uncovered line — the `InitFailedError` branch in `_ask_hook_installer()`. I kept it as a guard against the installer disappearing between the question and the install step (e.g. PATH change), and added a test for it in bb343bc8 instead of removing it.

Could you take another look when you get a chance? Thanks!

woile (approved):

PR #2078, opened 2026-08-25
feat(bump): add commit filter pattern

schlotter:
## Description
This change adds support for filtering commits before Commitizen evaluates them for version bumps by introducing a new `bump_commit_filter_pattern` setting.

The new setting defaults to `.*`, so existing projects keep the current behavior. When configured, `cz bump` and `cz version --next USE_GIT_COMMITS` only consider commits whose full message matches the filter pattern before applying the existing bump rules. This is particularly useful for monorepos where only a subset of commits should affect a package's version.

The pull request also updates the related documentation and adds regression tests covering the new filtering behavior and fallback paths.

## Checklist

- [X] I have read the [contributing guidelines](https://commitizen-tools.github.io/commitizen/contributing/contributing)

- [X] Yes (GitHub Copilot CLI)

Generated-by: GitHub Copilot CLI following [the guidelines](http://commitizen-tools.github.io/commitizen/contributing/pull_request/#ai-assisted-contributions)

### Code Changes

- [X] Run `uv run poe all` locally to ensure this change passes linter check and tests
  - [X] Document any manual testing steps performed

### Documentation Changes

- [X] Run `uv run poe doc` locally to ensure the documentation pages renders correctly
- [X] Check and fix any broken links (internal or external) &mdash; There were no broken links.

## Expected Behavior
Projects can opt into filtering commits used for bump calculation without changing the default behavior for existing users. With `bump_commit_filter_pattern` configured, unrelated commits are ignored before `bump_pattern` is applied, so `cz bump` and `cz version --next USE_GIT_COMMITS` derive the next version only from relevant commits.

1. Configure Commitizen with `bump_commit_filter_pattern` for a subset of commit messages, for example `^fix\(library-b\):`.
2. Create a baseline tag and add commits that include both matching and non-matching messages.
3. Run `cz bump --dry-run` or `cz bump --get-next` and confirm that only matching commits affect the calculated increment.
4. Run `cz version --project --next USE_GIT_COMMITS` and confirm it reports the version derived from the filtered commit set.
5. Remove the setting or leave it at the default and confirm the previous behavior is preserved.

## Additional Context
Related discussion: #2075

Manual testing and validation performed during development included targeted pytest coverage for:
- commit filtering in bump calculation
- `USE_GIT_COMMITS` next-version calculation with filters
- fallback behavior when no explicit filter is configured
- configuration and documentation updates for the new setting


woile:
How is this any different from `bump_pattern`?

schlotter:
To illustrate the use case, consider the example commit sequence from #2075. In this monorepo, commits related to applications *AppA* and/or *AppB* are identified through a custom `Applications:` line in the commit message:

```text
-- Commit 1 -----
feat: implement library

Applications: ['AppA', 'AppB']

-- Commit 2 -----
fix: fix nasty bug

Applications: ['AppB']

-- Commit 3 -----
ci: change release pipeline

Applications: ['AppA', 'AppB']

-- Commit 4 -----
...
```

When calculating the version bump for *AppA*, only commits that affect *AppA* should be considered. In this example, `bump_commit_filter_pattern` could be configured as:

```regex
(?s)^.*Applications:.*'AppA'.*\].*
```

This pattern selects only commits that relate to `AppA`.

`bump_pattern` cannot be used for this purpose because it is evaluated **line by line**, rather than against the full commit message.

As implemented today in [`find_increment()` in `commitizen/bump.py`](https://github.com/commitizen-tools/commitizen/blob/4184174324d14448eed0f9814520458f90308d78/commitizen/bump.py#L36-L38):

```py
for commit in commits:
    for message in commit.message.split("\n"):
        result = select_pattern.search(message)
```

(where `select_pattern` is `bump_pattern`)

By contrast, `bump_commit_filter_pattern` would act as a **pre-filter** on the list of commits considered for version bump calculation and would evaluate the **entire commit message**. In my PR, this is implemented in [`filter_commits()` in `commitizen/bump.py`](https://github.com/commitizen-tools/commitizen/pull/2078/changes#diff-ed9bead31489109a8441fbfb00d2ddec00198489b0702bab164825c2d3bd9b4fR73):

```py
return [commit for commit in commits if select_pattern.match(commit.message)]
```

(where `select_pattern` is `bump_commit_filter_pattern`)

I do not think `bump_pattern` itself can be changed to evaluate the full commit message without introducing a backward compatibility break.

Today, Commitizen evaluates each line of a commit message against `bump_pattern`, determines whether that line implies a `PATCH`, `MINOR`, `MAJOR`, or no bump, and then uses the highest bump level found within the commit. Existing configurations may rely on this behavior, including the ability to map different lines of the same commit message to different bump levels via `bump_map`.

For that reason, I see `bump_commit_filter_pattern` as complementary to `bump_pattern` rather than a replacement: it decides *whether a commit should participate in bump calculation at all*, while `bump_pattern` continues to decide *which bump level a selected commit contributes*.

woile:
Yes, but it's the same, now we evaluate twice the commit message. 

Why not a setting like `split_commit_message`?

```
for commit in commits:
    messages = commit.message.split("\n") if split_commit_message else [commit.message]
    for message in messages:
        result = select_pattern.search(message)
```

schlotter:
Mmh, I would still like to support commit messages such as the following:

```text
feat: allow provided config object to extend other configs

BREAKING CHANGE: `extends` key in config file is now used for extending other config files

Applications: ['AppA', 'AppB']
```

When calculating the version bump for *AppA*, I would first want to identify this commit as relevant using a `bump_commit_filter_pattern` such as:

```regex
(?s)^.*Applications:.*'AppA'.*\].*
```

The commit would then participate in the bump calculation. After that, I would still want the existing `bump_pattern` / `bump_map` logic to analyze the commit **line by line** in order to determine the bump level. In this example, both `feat` and `BREAKING CHANGE` would be evaluated, and the resulting bump level would be the highest one found.

This is why I currently do not see how a setting such as `split_commit_message` would solve the problem. These are actually two separate concerns:

1. **Commit filtering**: Decide whether a commit is relevant for a particular application. This requires evaluating the **entire commit message**.
2. **Bump calculation**: Determine the bump level (`PATCH`, `MINOR`, `MAJOR`) for a relevant commit. This currently relies on evaluating the commit **line by line**.

In my use case, both behaviors are needed at the same time.

The proposed `bump_commit_filter_pattern` is intended to address only the first concern. It acts as a pre-filter that decides whether a commit should be considered at all during bump calculation. Once a commit has been selected, the existing `bump_pattern` behavior can remain unchanged.

This would also be analogous to `changelog_pattern`, which is used to select commits for changelog generation based on the full commit message. In #2075 I considered reusing `changelog_pattern` for this purpose, but currently I prefer a dedicated configuration option because changelog generation and bump calculation are conceptually separate features.

Manny7717 (approved):
Verified locally on head 1e56d63. The feature is correctly scoped and the precedence chain is right: cz class attribute (None for custom cz classes, `.*` default for conventional/customize) -> top-level `bump_commit_filter_pattern` config -> `DEFAULT_SETTINGS` fallback, applied consistently in both `bump` and `version` commands.

What I re-ran:
- `filter_commits` edge cases: default `.*` keeps all commits (backward compatible), anchored `re.match` at message start, alternation patterns, and a pattern matching zero commits -> empty list.
- Empty filtered list feeds `find_increment` -> returns None, and the new integration test `test_bump_commit_filter_pattern_excludes_all_commits` confirms the user-facing `NoneIncrementExit` ("The commits found are not eligible to be bumped") rather than a crash.
- `CustomizeCommitsCz` attr wiring: `customize.bump_commit_filter_pattern` overrides via the existing setattr loop; plain class default otherwise. Consistent with `bump_pattern` handling.
- Targeted suites (test_bump_command, test_version_command, test_bump_find_increment, test_conf, test_cz_customize): 345 passed. Full suite: only failure is `test_bump_pre_commit_changelog[True-pre-commit]`, which fails identically on base 4184174 (changelog markdown-escaping env difference) — zero new failures.
- ruff + mypy clean.

Docs (bump.md, configuration_file.md, monorepo_guidance.md) explain the feature and its interaction with `bump_pattern`. No issues found.

Manny7717 (commented):
Reviewed head 1e56d63 against origin/master. The code is verified sound and backward-compatible; one docs claim needs correcting before this should land.

**Code - verified working:**
- End-to-end bump probe in a scratch repo with `bump_commit_filter_pattern = "^(feat|fix)\\(library-b\\)(!)?:"` and commits `feat(library-a)` + `fix(library-b)`: increment correctly resolves to PATCH and tags `0.1.1-library-b` - the library-a feat no longer forces a MINOR. `cz version --next USE_GIT_COMMITS` filtering is covered by the new version-command tests.
- Regression coverage is real: transplanting the PR's tests onto base gives 6 failures (AttributeError `bump_commit_filter_pattern` + behavioral), all pass on head (7/7). Full suite: head 1313 passed/6 failed vs base 1306 passed/13 failed with PR tests transplanted - the 6 head failures are the same pre-existing `prek`/`pre-commit` env failures present on base; failure sets otherwise identical. Default `.*` fallback preserves existing behavior (defaults test passes on both base and head). `ruff check`/`ruff format --check` clean.
- The None/default/configured precedence ladder in `_find_increment`/`_get_next_git_version` behaves correctly for both built-in CZs (class attr is the `.*` default, so the top-level config wins) and `cz_customize` (customize-section value wins via the attr loop). Invalid-pattern handling (bare `re.compile`) matches the pre-existing `find_increment` behavior - consistent, not a regression.

**Finding - docs claim not backed by behavior (monorepo_guidance.md):** the retitled "Bump and changelog per component" section and its new prose ("we want library-b's *bump calculation and changelog* to only include commits that use the library-b scope") now attribute changelog scoping to the step-2 config, which only sets `bump_commit_filter_pattern`. That is not what happens: in my e2e with `update_changelog_on_bump = true`, the generated CHANGELOG.md for the library-b release still listed the `feat(library-a)` commits. The changelog is only scoped by the separate `[tool.commitizen.customize] changelog_pattern` mechanism shown in the second example below - which, on its own, leaves the *bump* unscoped. As written, the section implies one config does both when neither example config does.

  Suggested fix: merge the two examples - one `cz_customize` config showing both `bump_commit_filter_pattern` (top level, scopes the bump) and `customize.changelog_pattern` (scopes the changelog) - and adjust the prose so each mechanism's effect is stated accurately. `docs/config/bump.md` itself is accurate as written.

**Minor:** `test_bump_get_next_warns_when_changelog_flag_is_set` tests a pre-existing warning (bump.py:254 on master, unchanged by this PR) - harmless extra coverage, but it is unrelated to the filter feature; consider dropping it to keep the diff one-logical-change per the contributing guidelines.

schlotter:
Thank you!

In the latest push, I have addressed the docs finding. I have not addressed the minor finding regarding unrelated coverage increase.

woile:
I understand what you want @schlotter. However, this is your own "convention", on top of conventionalcommits.
Which that just makes commitizen more complex.

Instead, to have this kind of flexibility, users are encouraged to make their own "rules", you can check this one: [cz-kpn](https://github.com/kpn/cz-kpn/tree/main).

However, in your case, it wouldn't be enough, because the base class doesn't support filtering commits. 

My suggestion for this PR is to introduce a new function in the [`Base` class](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/cz/base.py#L38)

```python
def filter_commits(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
    """Select the commits to keep during check, bump and changelog generation
    return commits
````

Which could be included, right before [bump.py#L161](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/bump.py#L161) , [changelog.py#L260](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/changelog.py#L260), and after [check.py#L94](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/commands/check.py#L94)

With this, you can come up with your own rule, and even extend the `[tool.commitizen]`  section to include something like:

```toml
[tool.commitizen]
# ...
app = "AppA"
```

@Lee-W @noirbizarre @bearomorphism thoughts? I'm not sure if `check` should be included here or not

schlotter:
OK, thanks. Makes sense. I'm waiting for @Lee-W @noirbizarre @bearomorphism thoughts before proceeding.

Lee-W:
The idea itself seems useful for version bumping and changelog generation, but I'm less sure about using it for checks, unless it's guaranteed to be a superset of the commit checks.

The feature is useful, but it does add some complexity to the project. 🤔 We'll need to think through a few different cases how these configs interact with each other

noirbizarre:
I agree with @Lee-W 

But, given I've faced this limitation many times, I also agree on the root cause:
> `bump_pattern` cannot be used for this purpose because it is evaluated line by line, rather than against the full commit message.

Note that I think there is the same issue with `changelog_pattern` which in my memory is also line by line.

So I think the long-term solution to properly handle that is to make `bump_pattern` multiline-aware.
I have no opinion on how it should be done (breaking change vs opt-in behavior via a flag), but I think it would avoid an extra setting that can be misleading/hard to understand because of the functional overlap.

woile:
> We'll need to think through a few different cases how these configs interact with each other
To make it very explicit, we could have 3 functions, and if eventually needed, we could have `filter_commits_before_check`. Or the user could decide to just implement `filter_commits`

```python
def filter_commits(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
    """Select the commits to keep during check, bump and changelog generation"""
    return commits

def filter_commits_before_bump(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
    """Select the commits to keep during check, bump and changelog generation"""
    return self.filter_commits(commits)

def filter_commits_before_changelog(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
    """Select the commits to keep during check, bump and changelog generation"""
    return self.filter_commits(commits)
```


> So I think the long-term solution to properly handle that is to make bump_pattern multiline-aware.

The problem with multiline is that it becomes unclear how to parse something like this:

```
feat: allow provided config object to extend other configs

BREAKING CHANGE: `extends` key in config file is now used for extending other config files
```

Do you keep the message after `feat:` or after `BREAKING CHANGE`?

noirbizarre:
> > We'll need to think through a few different cases how these configs interact with each other
> To make it very explicit, we could have 3 functions, and if eventually needed, we could have `filter_commits_before_check`. Or the user could decide to just implement `filter_commits`
>
> ```python
> def filter_commits(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
>     """Select the commits to keep during check, bump and changelog generation"""
>     return commits
> 
> def filter_commits_before_bump(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
>     """Select the commits to keep during check, bump and changelog generation"""
>     return self.filter_commits(commits)
> 
> def filter_commits_before_changelog(self, commits: list[git.GitCommit]) -> list[git.GitCommit]:
>     """Select the commits to keep during check, bump and changelog generation"""
>     return self.filter_commits(commits)
> ```
I like this approach, simple but flexible

> > So I think the long-term solution to properly handle that is to make bump_pattern multiline-aware.
> 
> The problem with multiline is that it becomes unclear how to parse something like this:
> 
> ```
> feat: allow provided config object to extend other configs
> 
> BREAKING CHANGE: `extends` key in config file is now used for extending other config files
> ```
> 
> Do you keep the message after `feat:` or after `BREAKING CHANGE`?

In the case of bump, line splitting was OK as we don't use the parsed content, but does not scale for the current use case.

In the case of changelog generation, the multilines splitting is a big constraint as it prevents doing a special treatment on the commit which is breaking, as the `BREAKING CHANGE` is seen as a separate line to process (this is also the case for any git trailer if a line is skipped before, or for a multi-paragraph body with a line skipped, due to https://github.com/commitizen-tools/commitizen/blob/master/commitizen/changelog.py#L153).

Once you know it, you can adapt to it (basically prioritize 1-liner conventional commit messages).

I don't know how to handle this, but the current approach splits and then filter out part of the commit message resulting in in data loss (and this apply to all plugins). 

Maybe the line splitting should be moved into a plugin, and the commit processing should allow splitting from the plugins. Or perhaps there should be something to toggle the line splitting.

woile:
@Lee-W if you are in we can go with my last proposal. I think it would be a good start.

> I don't know how to handle this, but the current approach splits and then filter out part of the commit message resulting in in data loss (and this apply to all plugins).

On the same boat, I was experimenting a bit with regex, and it's possible to parse everything, but it ends in different groups. The problem is that different conventions may need different groups. Which I think it could be done if we were to move [some of this logic](https://github.com/commitizen-tools/commitizen/blob/master/commitizen/bump.py#L38-L60) under the `cz.Base` as well.

bearomorphism:
Sorry for late response. I recall I have a PR aiming to redesign `bump_pattern` related things, and I can probably check that covers your use case.

bearomorphism:
here it is: https://github.com/commitizen-tools/commitizen/pull/1518

schlotter:
Many thanks for the discussion, @Lee-W @noirbizarre @bearomorphism @woile!

I did not dare to build on #1518: I feel this would have been a larger and more substantial change. Instead, I have built on the idea of adding `filter_commits` to `Base`.

You can find my proposal here: #2090. Let me know what you think!

Issue #2079, opened 2026-08-25
body_length_limit collapses blank lines in commit body


schlotter:
### Description

When `body_length_limit` is enabled, blank lines in the commit body are removed.

### Reproduction

With `body_length_limit = 80` (or `cz commit --body-length-limit 80`), create a commit message like:

```text
feat: subject

First paragraph line

Second paragraph line
```

### Expected behavior

Blank lines in the body are preserved, and only long non-empty lines are wrapped.

### Actual behavior

The blank line between the two body paragraphs is removed:

```text
feat: subject

First paragraph line
Second paragraph line
```

PR #2080, opened 2026-08-25
fix(commit): preserve blank lines with body length limit

schlotter:
## Description

Fix `body_length_limit` so it preserves intentional blank lines in commit bodies while still wrapping non-empty lines to the configured width.

## Checklist

Generated-by: GitHub Copilot CLI following [the guidelines](http://commitizen-tools.github.io/commitizen/contributing/pull_request/#ai-assisted-contributions)

### Code Changes

- [ ] <s>Update the documentation for the changes</s> &mdash; not required, just a fix

### Documentation Changes

- [ ] <s>Run `uv run poe doc` locally to ensure the documentation pages renders correctly</s> &mdash; no documentation changes
- [ ] <s>Check and fix any broken links (internal or external)</s> &mdash; no documentation changes

## Expected Behavior

With `body_length_limit` enabled, Commitizen should preserve blank lines in the commit body and only wrap long non-empty lines.

1. Configure `body_length_limit = 80` or run `cz commit --body-length-limit 80`.
2. Create a commit message body with a paragraph break, for example:

   ```text
   First paragraph line

   Second paragraph line
   ```
3. Confirm the generated commit message still contains the blank separator line.
4. Run `uv run pytest tests/commands/test_commit_command.py -k body_length_limit`.
5. Run `uv run ruff check commitizen/commands/commit.py tests/commands/test_commit_command.py`.
6. Run `uv run ruff format --check commitizen/commands/commit.py tests/commands/test_commit_command.py`.

## Additional Context

This fixes a regression where `textwrap.wrap("")` caused empty body lines to be dropped when `body_length_limit` was enabled (#2079).

### DIFF
--- a/commitizen/commands/commit.py
+++ b/commitizen/commands/commit.py
@@ -120,9 +120,12 @@ class Commit:
         if len(lines) < 3:
             return message
 
-        # First line is subject, second is blank line, rest are body lines
+        # Preserve intentional blank (or whitespace-only) lines while wrapping
+        # non-empty body lines. textwrap.wrap() returns [] for such lines, so
+        # they must be special-cased to avoid being silently dropped.
         wrapped_body_lines = chain.from_iterable(
-            textwrap.wrap(line, width=body_length_limit) for line in lines[2:]
+            textwrap.wrap(line, width=body_length_limit) if line.strip() else [line]
+            for line in lines[2:]
         )
         return "\n".join(chain(lines[:2], wrapped_body_lines))
 
--- a/tests/commands/test_commit_command.py
+++ b/tests/commands/test_commit_command.py
@@ -391,6 +391,11 @@ def test_commit_command_with_config_message_length_limit(
             id="preserves_line_breaks",
         ),
         pytest.param(
+            "Line1 is shorter than the limit but has paragraph break\n\nLine2 stays in a separate paragraph",
+            100,
+            id="preserves_blank_lines",
+        ),
+        pytest.param(
             "This is a very long line that exceeds 72 characters and should NOT be wrapped when body_length_limit is set to 0",
             0,
             id="disabled",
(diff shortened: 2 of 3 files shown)


commitizen/commands/commit.py
```diff
@@ -120,9 +120,10 @@ def _wrap_body(self, message: str) -> str:
         if len(lines) < 3:
             return message
 
-        # First line is subject, second is blank line, rest are body lines
+        # Preserve intentional blank lines while wrapping non-empty body lines.
         wrapped_body_lines = chain.from_iterable(
-            textwrap.wrap(line, width=body_length_limit) for line in lines[2:]
+            textwrap.wrap(line, width=body_length_limit) if line else [line]
```
Manny7717: Optional: `if line` is falsy only for the empty string. A whitespace-only line (`'   '`) is truthy, so it goes through `textwrap.wrap('   ', ...)` which returns `[]` — still dropped. If the intent is to preserve any intentional blank-looking line, `if line.strip() else [line]` covers both cases.

Manny7717 (approved):
## Review — Approved

Verified locally on head: **6/6 tests pass** (body-length-limit and message-length-limit params, including the new `preserves_blank_lines` case).

The fix is correct and minimal: `textwrap.wrap` returns `[]` for empty input, which is why blank lines were silently dropped; short-circuiting with `[line]` preserves paragraph breaks exactly as intended.

One optional, non-blocking suggestion (inline): the `if line` check only catches fully-empty strings — a whitespace-only body line still gets dropped because `textwrap.wrap('   ')` is `[]`. `if line.strip() else [line]` would preserve those too, if the intent is to preserve *any* intentional blank-looking line.

commitizen/commands/commit.py
schlotter: Thank you, it makes sense to me to preserve whitespace-only lines. I have applied your suggestion.

Manny7717 (approved):
Verified locally on head 171a234 (worktree against origin/master 4184174):

**Fix is correct.** `textwrap.wrap(line, width=...)` returns `[]` for both empty and whitespace-only lines, so `_wrap_body` silently dropped paragraph separators in commit bodies. The ternary keeps such lines verbatim (`[line]`) while still wrapping non-empty lines — exactly the right special-case. The `body_length_limit <= 0` early-return ("no limit") is untouched, so no behavior change for the disabled case.

**Regression-proven:** copied the PR's two new tests onto the base worktree → both FAIL on 4184174 (`preserves_blank_lines` drops the blank separator; `preserves_whitespace_only_lines` drops the `"   "` line), PASS on head. The direct assertion (rather than file_regression) for the whitespace-only case is a good call given the repo's trailing-whitespace pre-commit hook would strip the fixture.

**No regressions:** full tests/commands/test_commit_command.py 34/34 pass on head; ruff check clean.

One micro-note (non-blocking): the whitespace-only line is preserved byte-for-byte, so a body line of only spaces will survive into the commit message exactly as typed — that's the intent per the PR, just calling out the literal-preservation behavior.

bearomorphism:
LGTM, thanks!

bearomorphism (approved):

